PR #6, opened 2020-05-19
Tower inventories

kedark3:
## Purpose
Adding role that can add inventory and inventory sources. 

## Notes/Open items for discussion
- I deleted the tasks that use Tower-CLI stuff in this PR, any reason we want to keep it?
- OAuth Token - should all the roles support OAuth Token? Currently to acquire the token, user would need to use URI module as there is no Ansible Tower equivalent of https://docs.ansible.com/ansible/latest/modules/ovirt_auth_module.html , hence shall we support token or just username/password authentication? 
- Do we need license in each Role directory?
- Do we use Yaml/Json for data structure? I am using Yaml as it is easier for users to write. Seee `tests/configs`


Wilk42:
I believe ALL roles should support User/Password/OAuth Token, as these are module parameters, by leaving them as default omit it gives the user the choice which method to use. 

In addition, I found some intermittent issues with using user/pass with the new awx modules where it would get a timeout error when updating many items at once, I think its because of how it authenticates, which seemed to be solved by the OAuth token.

Wilk42:
I am wondering what others thoughts are separating inventory and inventory sources, they are intrinsically linked, but a user might wish to do one and not the other.

kedark3:
@Wilk42 for my Repo https://github.com/kedark3/ATCasC I had it separated. But since it was together in this code here, I kept it like that, I'm happy to split it if we all agree.

jlozadad:
to give more information the reason I kept the tower-cli its for compatibility reasons since there's going to be customers still using but, we can put and say "only works 3.5 and up".

jlozadad:
we should stick with YML unless a specific customer requests it. We can have json examples in the readme and such but, not in the variables. Json can be a pain or spend a lot of time troubleshooting.

jlozadad:
we have to look into license best practices and see if they can be changed easily.

Wilk42:
In regards to the licence, I don't think we are changing the licence, more of making it so there is a master licence file instead of 1 per role, as I believe they are all MIT.

Wilk42:
After Our discussion at minimum we need to put in a when option on this module for running each task only when the loop variable is defined. I'd still prefer to split them up, but that is a preference not a blocker.

PR #7, opened 2020-05-19
Proposing initial set of standards

kedark3:
# Purpose

Contribution guidelines - initial draft.

### DIFF
--- a/STANDARDS.md
+++ b/STANDARDS.md
@@ -1,1 +1,13 @@
 
+## Contributor's Guidelines
+
+- All YAML files named with '.yml' extension
+- Use spaces around jinja variables. {{ var }} over {{var}}
+- Variables that are internal to the role should be lowercase
+- Keep roles self contained - Roles should avoid including tasks from other roles when possible
+- Plays should do nothing more than include a list of roles except where pre_tasks and post_tasks are required when possible
+- Separators - Use dashes (e.g. my-role my-playbook) not underscores (my_role)
+- Paths - When defining paths, do not include trailing slashes (e.g. my_path: /foo not my_path: /foo/). When concatenating paths, follow the same convention (e.g. {{ my_path }}/bar not {{ my_path }}bar)
+- Indentation - Use 2 spaces for each indent
+- `vars/` vs `defaults/` - if you have variables that don't need to change or be overridden by user, put those in `vars/` and those that a user would likely override, put those under `defaults/` directory.
+- All playbooks/roles should be focused on compatibility with Ansible Tower
--- a/STANDARDS.md
+++ b/STANDARDS.md
@@ -6,7 +6,7 @@
 - Variables that are internal to the role should be lowercase
 - Keep roles self contained - Roles should avoid including tasks from other roles when possible
 - Plays should do nothing more than include a list of roles except where pre_tasks and post_tasks are required when possible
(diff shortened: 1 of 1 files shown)


STANDARDS.md
```diff
@@ -1 +1,13 @@
 
+## Contributor's Guidelines
+
+- All YAML files named with '.yml' extension
+- Use spaces around jinja variables. {{ var }} over {{var}}
+- Variables that are internal to the role should be lowercase
+- Keep roles self contained - Roles should avoid including tasks from other roles when possible
+- Plays should do nothing more than include a list of roles except where pre_tasks and post_tasks are required when possible
+- Separators - Use dashes (e.g. my-role my-playbook) not underscores (my_role)
```
Wilk42, line 9: In what context, I've just used underscores by default as depending on context. I had thought ansible disliked dashes in some cases, mainly variables, but also with role names.

kedark3: Mainly for file/directory names. Variables I prefer underscores.

STANDARDS.md
```diff
@@ -1 +1,13 @@
 
+## Contributor's Guidelines
+
+- All YAML files named with '.yml' extension
+- Use spaces around jinja variables. {{ var }} over {{var}}
+- Variables that are internal to the role should be lowercase
+- Keep roles self contained - Roles should avoid including tasks from other roles when possible
+- Plays should do nothing more than include a list of roles except where pre_tasks and post_tasks are required when possible
```
Wilk42, line 8: Should be on a case by case basis, but should be kept simple. Such as an example playbook that only imports the vars file if needed based on tags.

kedark3: yup, hence the `except` part. This is more like a suggestion but if it can't be followed, its fine - reviewers will decide if the PR looks good in such cases.

STANDARDS.md
jlozadad: for this collection is probably to continue to use _ since the roles are named like that and other files as well.

kedark3: If we get agreement tomorrow, I will change this.

bherrin3: @kedark3 I believe the right answer is typically dictated by the surround software.

I enforced that as we were tightly coupled with Jenkins which prefers dash case and not deeply integrated with python which prefers underscore.

If this project uses underscores highly, I suppose we can maintain that convention and stick with a specified set of conventions.

bherrin3:
Since this is the standards section, which is pretty broad, can we discuss two high level items?

1. Why is `genie` being used throughout this project? 
2. It seems highly superfluous to be prepending everything with `ansible_tower_genie_`. Can we be a little less verbose and just stick to the Ansible Tower function for the name and/or use simple naming methodologies as dictated by the collection implementation? i.e. `tower_job_template.yml`, `workflow_job_templates.yml`, `workflow_job_nodes.yml`?

Wilk42:
I think the genie and ansible_tower_genie being used is in regards to the roles best practices to use unique names in roles and we hadn't thought of best practices in a collection. Especially since a collection will be within its own namespace such as redhat_cop.tower_automation.rolename
We could likely drop the naming convention with the role names, but not sure about for the variables as they still need to be unique. 

Something to rethought once we think of a new name for the project, which is the 1 open issue at the moment.

PR #8, opened 2020-05-20
Adding Credentials and Credential Types, formating update

Wilk42:
# Purpose

Adding role that can add credentials and credential types
Updating a few older roles in formating.

# Notes/Open items for discussion

- I found an issue when creating credential types as the tower expects them in a variable format, I worked around this by adding a double space between curly braces and removing the double space with a regex filter. If anyone has a better way I'm open to changing this.


roles/ansible_tower_genie_credential_types/defaults/main.yml
```diff
@@ -0,0 +1,27 @@
+---
+tower_server: ""
+tower_oauthtoken: ""
+tower_validate_certs: False
+tower_genie_credential_types_secure_logging: "{{tower_genie_secure_logging | default(False)}}"
+tower_credential_types:
```
kedark3, line 6: Do we want to provide the default for credential_types or for every one of the roles ? :thinking:

roles/ansible_tower_genie_credential_types/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
+---
+- debug:
+    var: tower_credentials
```
kedark3, line 3: Is this something we wanna follow in all our roles? Asking some of these questions to help define guidelines.

roles/ansible_tower_genie_credential_types/tests/collections/requirements.yml
```diff
@@ -0,0 +1,3 @@
+collections:
```
kedark3, line 1: This is something we want to add in guidelines? Something we can make collection wide - that selects correct AWX collection version for entire collection?

roles/ansible_tower_genie_credentials/tasks/main.yml
```diff
@@ -1,2 +1,23 @@
 ---
 
+- name: Add Credentials
+  awx.awx.tower_credential:
+    name:                           "{{ tower_credentials_item.name }}"
```
kedark3, line 5: Are you omitting some of the fields ? I saw many more when I ran: ```ansible-doc awx.awx.tower_credential | grep "^[-=]" | grep -v \: | sed 's:^[-=] \(.*\):    \1\: "{{ source.\1 | default(omit) }}":'```

roles/ansible_tower_genie_credential_types/README.md
```diff
@@ -0,0 +1,210 @@
+# ansible_tower_genie_credential_types
+## Description
+An Ansible Role to create Credential Types in Ansible Tower.
+
+## Requirements 
+ansible-galaxy collection install -r tests/collections/requirements.yml to be installed 
+
+| Requiremed collections |
```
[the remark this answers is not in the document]
Wilk42: I am not seeing the typo, I'll change it but not seeing it.

roles/ansible_tower_genie_credential_types/README.md
```diff
@@ -0,0 +1,210 @@
...
+ansible-galaxy collection install -r tests/collections/requirements.yml to be installed 
+
+| Requiremed collections | 
+|:---:|
+|awx.awx|
+
+## Variables
+|Variable Name|Default Value|Required|Description|Example|
+|:---:|:---:|:---:|:---:|:---:|
+|`tower_server`|""|yes|URL to the Ansible Tower Server.|127.0.0.1|
```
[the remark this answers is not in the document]
Wilk42: Definitely, I may have messed up but something to put into the standards and make it the same.

roles/ansible_tower_genie_credential_types/defaults/main.yml
Wilk42: We could cut these out of the defaults, but keep in documentation.

roles/ansible_tower_genie_credential_types/tests/collections/requirements.yml
Wilk42: This is here mainly because I find it to be a good practice for the tests, once we start Actual testing we can adjust and decide, could also just drop it. 

I am thinking of putting in a documentation folder which has a FULL implementation of all the roles as an example for a user to follow.

roles/ansible_tower_genie_credential_types/tasks/main.yml
Wilk42: No, this is something I need to remove, will update.

roles/ansible_tower_genie_credentials/tasks/main.yml
Wilk42: In this case yes, The Module is built to use many use cases, however they are all being depreciated in favor of of the inputs field that then takes those same inputs in many different use cases. We could add them in, but as they were being depreciated I thought best practice at the moment was to not use them.

kedark3: Ah ok, is there any doc or info that tells this. I'm sure you know something that I don't :(

roles/ansible_tower_genie_credential_types/README.md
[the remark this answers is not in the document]
kedark3: `Requiremed` - this word should probably read `Required`. There is `m` that's not required.

roles/ansible_tower_genie_credentials/tasks/main.yml
Wilk42: I've been going to the module doc itself inside the python [here](https://github.com/ansible/awx/blob/devel/awx_collection/plugins/modules/tower_credential.py) to check on things, 

in credential types new name is listed in the python but not the docs, Haven't investigated but I included it as it should be included in the future at some point but also made a comment note about it.

roles/ansible_tower_genie_credential_types/README.md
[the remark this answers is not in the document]
Wilk42: I was looking at the command line argument and not the line 8 ><

PR #9, opened 2020-05-23
Adding GH actions ansible-lint

kedark3:
## Enable GH Actions

- Yaml-Lint
- Ansible-Lint

### Notes:
- Currently You can see the Actions are failing, that is because our Repo is non-compliant and we can address that in a following PR.
- These are basic checks, we can extend these if needed.

Ref: https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions


Wilk42:
We need to turn off 
line too long (83 > 80 characters)  (line-length)
as Ansible does not follow that convention. Otherwise looks good.

PR #11, opened 2020-05-29
Git lint

Wilk42:
Update to ignore line-length


kedark3:
@Wilk42 we already have a file https://github.com/redhat-cop/tower_configuration/blob/master/.yamllint.yml do you want to just update that?

Wilk42:
Saw that, removing this one

PR #13, opened 2020-05-29
Fix yaml-lint everywhere

Tompage1994:
## What this fixes
The github action is failing because of failing `yaml-lint` tests. This should fix all of them.

This also renames a load of files `*.yaml` > `*.yml` as per the repo standards.

Sorry for the **_HUGE_** diff

### NOTE
The `ansible-lint`ing will be much more involved so I have not made a start on that.

PR #14, opened 2020-05-29
Fix all the ansible-lint

Tompage1994:
## What this fixes
The github action is failing because of failing ansible-lint tests. This should fix all of them.

This builds on top of #13 **which should be merged first!**


Tompage1994:
Marking as draft until #13 is merged

PR #18, opened 2020-05-29
Splitting tower_inventory and inventory_sources role

kedark3:
## Purpose

Splitting the inventory and inventory sources roles, which makes sense and inventory_source itself is a different kind of object and putting it in different role make it clear to anyone looking at the repo and we support both of the objects aka inventory and inventory_sources. I'm open to thoughts otherwise.


Wilk42:
Lint errors, and I'm still not sure on why its causing issues with the conflict, is it not deleting correctly.

PR #19, opened 2020-06-03
Updating tower teams to use collections

kedark3:
## Purpose
Tower Teams role was not conforming with our usage of AWX Collections. Also needed to be renamed.

## Solution

Added:
- Usage of Collections to tower_teams role
- Fixed the name of the role to match repo name
- Added tests


roles/tower_teams/README.md
```diff
@@ -14,50 +14,33 @@ An Ansible Role to create Teams in Ansible Tower.
 |`tower_team_desc`|""|no|Description of the Ansible Tower Team to create.|
```
Tompage1994, line 14: Some of the variables in here are now outdated. `tower_org`,`tower_team_name` and `tower_team_desc` now come under `tower_team` which could then be explained separately. Perhaps see how I have laid out tower_rbac for this formatting, or some other format for how `tower_<object_name>` should be described. Whatever is decided should definitely be common throughout the repo.

kedark3: @Tompage1994 will check into your example. And update this README.

PR #20, opened 2020-06-03
Refactor rbac

kedark3:
## Purpose
We need to have role name consistent with Repo name and make sure the role can run tests against a fresh tower instance. Also to fix tests.

## Solution

- For example, Tower fresh install would usually come with `Demo Job Template` and `Demo Inventory` etc. hence making it use those to ensure tests will always pass. 
- Renamed the role.


roles/tower_rbac/defaults/main.yml
```diff
@@ -13,3 +13,7 @@ tower_rbac:
   - team: "My Team"
```
Tompage1994, line 13: Opening for discussion:
Should we really even have a default like this for this role? Does this make sense as it will fail because the objects may not exist as it is entirely dependent on other objects. It possible makes more sense in other roles but perhaps not here?
In short, would it make more sense to leave the default as `[]` and simply have an example in comments or in the readme?

kedark3: @Tompage1994 I did think about that in my other PRs, and I think we should use `Default` items present on fresh tower, so that test should be able to run on a fresh tower without needing us to create other objects.

kedark3:
@Tompage1994 removed the template for tower cli cfg as requested.

kedark3:
Fixed README @Tompage1994

roles/tower_rbac/defaults/main.yml
Wilk42: This is what the tower team does, For our defaults it makes sense.

PR #22, opened 2020-06-04
Update to Licences, Role names, and galaxy.yml

sean-m-sullivan:
Renamed a ton of files to get rid of tower_genie
* Initially Removed the collections/requirements.yml as this is now in dependences in galaxy.yml
* then found it was required for linting
* moved under .github folder as to not be compiled as part of collection or confuse users. 

Removed Licence files from individual roles
* This is now in the galaxy.yml as MIT which was what all the roles were. 

Updated the galaxy.yml to be in line with the official galaxy 

Moved ansible tower genie surveys to archive
* surveys can now be done in their respective roles, as shown in the playbook file. Should we delete this code?


sean-m-sullivan:
That I was not sure of, and Yes I need to update the playbooks to reflect the changes. I can get on that before we merge.

kedark3:
I was following `tower_*` for naming the roles, and had removed `ansible_` and `genie_` words from names. So `tower_settings`, `tower_projects` an so on. I'd like to have `tower_` prefix if we can keep it.

jlozadad:
I prefer if we keep the gpl v3 license and yes use the tower_ for roles name.

.github/workflows/ansible-lint.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         sudo apt install software-properties-common
         sudo apt-add-repository --yes --update ppa:ansible/ansible
         sudo apt install ansible
-        ansible-galaxy collection install -f -r collections/requirements.yml
+        ansible-galaxy collection install -f -r .github/collections/requirements.yml
```
kedark3, line 37: Why do we need to install collections? Just curious.

willtome:
What is the purpose of keeping `tower_` in the role name? When using it in a playbook it will likely look like `redhat_cop.tower_configuration.tower_project` so it would seem `redhat_cop.tower_configuration.project` would do the trick and save some keystrokes.

PR #30, opened 2020-06-17
Add workflow Role

sean-m-sullivan:
Add workflow Role


sean-m-sullivan:
Work is done, I'll be getting to the linting tomorrow.

sean-m-sullivan:
linting done, also added a thing to be more relaxed about indentation as awx-export is int hat format.

roles/workflow_job_templates/tasks/main.yml
```diff
@@ -0,0 +1,38 @@
...
+  loop_control:
+    loop_var: workflow_loop_var
+  no_log: "{{ workflow_job_templates_secure_logging }}"
+
+# Create links between workflow node
+- name: loop over nodes in schema to add to workflow templates
+  include_tasks: "add_workflows_schema.yml"
+  loop: "{{ tower_workflows }}"
+  loop_control:
+    loop_var: workflow_loop_var
```
Tompage1994, line 38: I think I would personally prefer to see this as `workflow` rather than `workflow_loop_var`. this would then have to be replaced in the `add_workflows_schema.yml` file too.

roles/workflow_job_templates/tests/yaml/tower_workflows.yml
```diff
@@ -0,0 +1,419 @@
...
+          scm_branch:
+          job_type:
+          job_tags:
+          skip_tags:
+          limit:
+          diff_mode:
+          verbosity:
+          all_parents_must_converge: false
+          identifier: d9779889-cfdb-4a8c-8a11-1f54acf84aca
+          workflow_job_template:
```
Tompage1994, line 34: I may be wrong, but it doesnt look like this (`workflow.related.workflow_nodes.X.workflow_job_template`) is actually used in the role

roles/workflow_job_templates/tests/yaml/tower_workflows.yml
```diff
@@ -0,0 +1,419 @@
...
+            organization:
+              name: Default
+              type: organization
+            name: Complicated workflow schema
+            type: workflow_job_template
+          unified_job_template:
```
Tompage1994, line 40: Again, I may be wrong, but I dont think anything other than `unified_job_template.name` is used

roles/workflow_job_templates/tests/yaml/tower_workflows.yml
```diff
@@ -0,0 +1,419 @@
...
+                  organization:
+                    name: Default
+                    type: organization
+                  name: Complicated workflow schema
+                  type: workflow_job_template
+                identifier: ca7ffc3b-8401-4153-83bf-bb3dd918ca87
+                type: workflow_job_template_node
+            failure_nodes: []
+            always_nodes: []
+          natural_key:
```
Tompage1994, line 70: Doesnt look like any of `natural_key` is used.

Am I missing something as to why all these things should be here, or can they simply be omitted from this data structure to simplify significantly?

roles/workflow_job_templates/meta/main.yml
```diff
@@ -0,0 +1,41 @@
...
+  platforms:
+    - name: "EL"
+      versions:
+        - "all"
+
+  galaxy_tags:
+    - "ansibletower"
+    - "tower"
+    - "awx"
+    - "genie"
```
Tompage1994, line 34: perhaps dont need `genie` here

roles/workflow_job_templates/tasks/main.yml
sean-m-sullivan: I thought workflow was a little to generic, so I tried to add on to it, I'm totally up for changing it, just thinking do we have something more descriptive to use?

roles/workflow_job_templates/tests/yaml/tower_workflows.yml
sean-m-sullivan: So this is an exact output from the tower export command. Part of this is that its designated in two places, because its from two places in the API, 
Specfically 
"/api/v2/workflow_job_template_nodes/",
and 
"/api/v2/workflow_job_templates/",
and then
/api/v2/workflow_job_templates/72/workflow_nodes/

I was trying to keep the pure output, but also pick the top level var when needed for things like Which workflow we are working with.

roles/workflow_job_templates/tests/yaml/tower_workflows.yml
sean-m-sullivan: They can be, and maybe thats what we want to do. I opted for the Pure raw tower_export as the data structure. Rather then a distilled version

roles/workflow_job_templates/meta/main.yml
sean-m-sullivan: Oops, fixing.

roles/workflow_job_templates/tests/yaml/tower_workflows.yml
sean-m-sullivan: So key things used are:
identifier
unified_job_template.name
all_parents_must_converge
related.always_nodes
related.success_nodes
related.failure_nodes
state

The rest is fluff pulled from the api with the tower export module.

roles/workflow_job_templates/tests/yaml/tower_workflows.yml
sean-m-sullivan: Removed this from examples but kept pure export in our playbook example

roles/workflow_job_templates/README.md
```diff
@@ -0,0 +1,272 @@
...
+    ask_inventory_on_launch: false
+    ask_scm_branch_on_launch: false
+    ask_limit_on_launch: false
+    webhook_service: ''
+    webhook_credential:
+    organization:
+      name: Default
+    related:
+      schedules: []
+      workflow_nodes:
```
kedark3, line 217: Do you not need to define the nodes separately? Or this is it? Just curious.

PR #34, opened 2020-06-18
Death of the Genie

Tompage1994:
Removal of all mentions of `genie` from the collection.

I think everything should still work off the back of this


kedark3:
:100: Nice work. Thanks.

Issue #24, opened 2020-06-04
Create test infra inside of gitlab runner


mpetrive-rh:
utilize docker-compose to stand up tower/awx infra

sean-m-sullivan:
Complete its now working

PR #35, opened 2020-06-19
Add awx action

Tompage1994:
This PR spins up a full AWX and runs the example yaml playbook against it. 

We could probably refine this a little but I think this is a good start for CI testing

### DIFF
--- a/.github/requirements.txt
+++ b/.github/requirements.txt
@@ -1,0 +1,3 @@
+docker
+docker-compose
+jmespath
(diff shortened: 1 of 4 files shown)


sean-m-sullivan:
I love it, testing now

sean-m-sullivan:
"msg": "Found too many names Demo Credential at endpoint credentials try using an ID instead of a name",
Failed on Template 2

Tompage1994:
I guess that was the first time seeing the pipeline fail because the playbooks were actually wrong 👍

sean-m-sullivan:
Exactly, Thanks for taking care of it, was going to dive into it this morning.

Tompage1994:
closes #24

PR #37, opened 2020-06-23
Workflow fix

sean-m-sullivan:
Fix to allow no survey to be defined. Found the to_json was unneeded.


sean-m-sullivan:
Was thinking of this as well, Going to update the default, the readme, and add a few tests in the playbook, one without everything, one with.

sean-m-sullivan:
Updated so If you were so inclined you could include a workflow and the only thing defined is the name.

roles/workflow_job_templates/tasks/add_workflows_schema.yml
```diff
@@ -6,7 +6,7 @@
     workflow :                      "{{ workflow_loop_var.name }}"        # Workflow job template name to associate with
     state:                          "{{ workflow_loop_node.state | default(tower_state | default('present')) }}"
     all_parents_must_converge :     "{{ workflow_loop_node.all_parents_must_converge | default (omit) }}"
-    organization:                   "{{ workflow_loop_var.organization.name }}" # Workflow job template organization
+    organization:                   "{{ workflow_loop_var.organization.name | default(omit) }}" # Workflow job template organization
```
kedark3, line 9: Can we change this to be `workflow_loop_var.organization` ? Is there a reason to add it with `.name`? If we remove that we will be able to do:
`organization : Default` 

vs 

```organization:
        name: Default
```

Do we envision Organization to have other attributes than name?

roles/workflow_job_templates/tasks/main.yml
```diff
@@ -4,17 +4,17 @@
 - name: Create Workflow
   awx.awx.tower_workflow_job_template:
     name:                               "{{ workflow_loop_var.name }}"
-    description:                        "{{ workflow_loop_var.description }}"
+    description:                        "{{ workflow_loop_var.description | default(omit) }}"
     allow_simultaneous:                 "{{ workflow_loop_var.allow_simultaneous | default('no') }}"
-    organization:                       "{{ workflow_loop_var.organization.name }}"
+    organization:                       "{{ workflow_loop_var.organization.name | default(omit) }}"
```
kedark3, line 9: Same question here as above, for `organization.name`

roles/workflow_job_templates/tasks/add_workflows_schema.yml
sean-m-sullivan: Its because thats the form that comes out of the Tower export 
it actually comes out as 
```
               "organization": {
                    "name": "Testing",
                    "type": "organization"
               },
```

roles/workflow_job_templates/tasks/main.yml
sean-m-sullivan: see above same thing, its the format from tower export.

roles/workflow_job_templates/tasks/add_workflows_schema.yml
Tompage1994: I’m still not sure I’m 100% comfortable with it being aligned to the tower export. I think we need to make a decision whether to align the data structure close to tower export or closer to the awx_collection. I can see both ways being beneficial but I think I’m leaning more towards the simplicity of aligning to the collection. 

I get the use case of having done an export and wanting to then import using the roles, however will it not make more sense that a user would then instead use an import role? 

Additionally we will have to make changes on all other roles to complicate the data structure which will be a big change.

kedark3: I see @sean-m-sullivan and @Tompage1994's point of view. I am not sure what's the right answer here, but I know @john-westcott-iv would be able the right person to answer probably. As he was working on tower feature to sync configs.

sean-m-sullivan:
Updated to use dual models, Will likely fail some lints and other tests, still needs some documentation updates. Will return to this soon. However should be ready for review.

sean-m-sullivan:
Still need to update docs, but everything is passing now.

sean-m-sullivan:
Updated docs, Please review and give comment/approval

.github/workflows/push.yml
```diff
@@ -46,3 +46,6 @@ jobs:
 
       - name: "Perform playbook tests"
         run: ansible-playbook playbooks/example_with_yaml/configure_tower.yml -e tower_hostname=http://localhost -e tower_username=admin -e tower_password=password
+
+      - name: "Perform export model playbook tests"
```
kedark3, line 50: Nice. :100:

roles/workflow_job_templates/README.md
```diff
@@ -59,7 +59,7 @@ workflow_job_templates_secure_logging defaults to the value of tower_genie_secur
 |Variable Name|Default Value|Required|Type|Description|
 |:---:|:---:|:---:|:---:|:---:|
 |`workflow_job_template`|""|yes|str|The workflow job template the node exists in. Used for looking up the node, cannot be modified after creation.|
-|`identifier`|""|yes|str|An identifier for this node that is unique within its workflow. It is copied to workflow job nodes corresponding to this node.|
+|`identifier`|""|yes|str|An identifier for this node that is unique within its workflow. It is copied to workflow job nodes corresponding to this node. It can be any value. Recomended to use Column and row numbers for identifiers.|
```
kedark3, line 62: Are the ids required to be universally unique or just unique within a workflow?

playbooks/example_with_yaml/configure_tower_export_model.yml
```diff
@@ -0,0 +1,18 @@
...
+  # tower_hostname: ansible-tower-web-svc-test-project.example.com
+  # tower_username: admin
+  # tower_password: changeme
+  pre_tasks:
+    - name: Include vars from tower_configs directory
+      include_vars:
+        dir: ./tower_configs_export_model
+        ignore_files: [tower_config.yml.template]
+        extensions: ["yml"]
+  roles:
```
kedark3, line 17: Do you think we should configure tower fully, instead of just the workflows? That will mean we call all roles in here. Or you are planning to add that later?

roles/workflow_job_templates/README.md
sean-m-sullivan: Within a workflow, I've named many Node101 across different workflows.

playbooks/example_with_yaml/configure_tower_export_model.yml
sean-m-sullivan: Tower Export model runs AFTER tower regular, and right now it doesn't export everything. Plan to add them to this list As we add the new data models to roles. Figured instead of Running the old ones, just run the different ones.

sean-m-sullivan:
I'm going to leave this up till the next meeting, in case more questions come up or we want to clarify, We will discuss and merge after.

playbooks/example_with_yaml/configure_tower_export_model.yml
kedark3: Got it. Thanks.

roles/workflow_job_templates/README.md
kedark3: Got it. Maybe worth mentioning this in the README?

sean-m-sullivan: Updated and included link to the said awx documentation about it, that I just now found.

kedark3:
Interesting, so if I omit the identifier, it will still work? I like that flexibility. @sean-m-sullivan thanks for fixing the readme.

PR #38, opened 2020-06-25
Update to fit standards for galaxy publish

Tompage1994:
Tested collection can be published to Galaxy via the action. 
Output here:
- https://github.com/Tompage1994/automate_tower_genie_collections/actions/runs/147659522
- https://galaxy.ansible.com/tompage1994/tower_configuration

Also adds in defaults and tests for

- users
- hosts
- notifications

Also removed the archived role as that would have caused all sorts of problems.


Tompage1994:
No, you can do away with the tests folder entirely, but I thought I would add it for standardisation compared to the rest of the roles

kedark3:
@Tompage1994 I think we are clear to merge :rocket: :ship:

PR #39, opened 2020-06-29
Add credential input sources role

Tompage1994:
My module in awx.awx was merged and 13.0 release has been made meaning it is now available. This role makes use of that.


kedark3:
@Tompage1994 I will get this reviewed today :)

PR #40, opened 2020-07-02
Cleanup lint and files

sean-m-sullivan:
Moved all Workflow files to the workflow folder, and any reference of them fixed. 
Removed all lint warnings by reformating the code to be properly within lint standards
Removed Truthy Lint warings by disabling them, Was complaining about True vs true and False vs false.


.github/workflows/.yamllint.yml
```diff
@@ -8,3 +8,4 @@ rules:
   indentation:
     level: warning
     indent-sequences: consistent
+  truthy: disable
```
kedark3, line 11: I think we should standardize this to something. Internally we have "True/False" - with first letter always in caps. Can we mass replace those in code base. I do not want to disable this option if possible.

kedark3 (changes requested):
Requesting one change, if we can do it. Everything else looks good.

.github/workflows/.yamllint.yml
sean-m-sullivan: I just read up on this based on your comment, I will look into adding allowed values of "True/False" in addition to "true/false" as per https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.truthy

sean-m-sullivan: Updated to allow True/False/true/false changed a yes/no from tower_role/tests/test.yml to pass the test

Tompage1994 (approved):

kedark3 (approved):

PR #51, opened 2020-07-07
Github Actions Fix

Tompage1994:
Github actions looks inside of the workflows dir for any workflows. Right now we have the config files included in that directory (e.g. .ansiblelint.yml) This was causing extra actions to attempt to run (and fail because it wasnt the correct format for an action). 

I can see they dont run on this repo so perhaps some manual config has been made, but it would be best that it just works out of the box, so I have moved the config files to a separate location (`.github/workflow-config`).

Additionally the release action was broken as the playbook was referring to `playbook_dir` from when the playbook was at project root. I have now made a change and done a test [here](https://github.com/Tompage1994/tower_configuration/actions/runs/160493532) which has succeeded in pushing to galaxy [here](https://galaxy.ansible.com/tompage1994/tower_configuration)


sean-m-sullivan (approved):
Looks good

PR #56, opened 2020-07-16
Remove json and simplify playbook folder

sean-m-sullivan:
### What does this PR do?
Removes the out of date json playbook
Converts the export.json file to yaml files for use in testing.

### How should this be tested?
Should be tested with our automated workflow, much of this is just housekeeping.

### Is there a relevant Issue open for this?
No

PR #58, opened 2020-07-16
Update Workflow options

### LINKED ISSUE #55: Update Workflow nodes with new options
Tower workflow nodes have options that are not included in our role, like inventory and scm, we need to add those in for use.

sean-m-sullivan:
### What does this PR do?
Adds missing module options from workflow templates and workflow nodes

Automated tests

resolves #55

Requires PR to update simiplified playbooks.
#56


Tompage1994 (approved):
Looks good.

@sean-m-sullivan All these new options applied against the workflow module, are they new to the awx collection module or were they just missed in the first place? If they were missed does that mean we need to make a check against all the other modules to make sure we hit all options?

sean-m-sullivan:
@Tompage1994 I didn't need to update the Readme, these were all in there when I wrote the module, but somehow missed them. We should check the other modules as well. Wish there was a easy way to do that, but I was using the role as a reference when someone asked about these options and noticed them missing.

sean-m-sullivan:
Rebasing this and then shall be cleaner

playbooks/tower_configs_export_model/tower_workflows.yml
```diff
@@ -12,23 +12,17 @@ tower_workflows:
...
     ask_limit_on_launch: false
-    webhook_service: ''
-    webhook_credential:
     organization:
       name: Default
       type: organization
     related:
       schedules: []
       workflow_nodes:
-        - extra_data: {}
```
kedark3, line 23: We can no longer pass this `extra_data`?

roles/workflow_job_templates/tasks/add_workflows_schema.yml
```diff
@@ -3,6 +3,16 @@
 - name : Create the Workfow Node
   awx.awx.tower_workflow_job_template_node:
     identifier :                    "{{ workflow_loop_node.identifier }}"   # internal identification ID
+    extra_data:                     "{{ workflow_loop_node.extra_data | default (omit) }}"
```
kedark3, line 6: Can we standardize between extra_vars and extra_data? I like former better. Even though modules aren't accepting same variable, we can do maybe:

```
extra_data:                     "{{ workflow_loop_node.extra_vars | default (omit) }}"
```

roles/workflow_job_templates/tasks/add_workflows_schema.yml
```diff
@@ -3,6 +3,16 @@
...
+    inventory:                      "{{ workflow_loop_node.inventory | default (omit) }}"
+    scm_branch:                     "{{ workflow_loop_node.scm_branch | default (omit) }}"
+    job_type:                       "{{ workflow_loop_node.job_type | default (omit) }}"
+    job_tags:                       "{{ workflow_loop_node.job_tags | default (omit) }}"
+    skip_tags:                      "{{ workflow_loop_node.skip_tags | default (omit) }}"
+    limit:                          "{{ workflow_loop_node.limit | default (omit) }}"
+    diff_mode:                      "{{ workflow_loop_node.diff_mode | default (false) }}"
+    verbosity:                      "{{ workflow_loop_node.verbosity | default (omit) }}"
+    workflow_job_template:          "{{ workflow_loop_node.workflow_job_template.name | default( workflow_loop_node.workflow_job_template | default (omit) ) }}"
```
kedark3, line 15: Does double default like this work? Have you tested it? Just curious.

kedark3 (commented):
Couple of comments. If you could please respond, we should be good to go. And I can Approve/Merge it.

roles/workflow_job_templates/tasks/add_workflows_schema.yml
sean-m-sullivan: Yes it was working in the last version, and works with both data models in the testing. It was thinkings like Limit, Diff mode, and other options that were missing, and order got changed in this go around.

roles/workflow_job_templates/tasks/add_workflows_schema.yml
sean-m-sullivan: So I went off the module here [extra data in workflow nodes](https://github.com/ansible/awx/blob/devel/awx_collection/plugins/modules/tower_workflow_job_template_node.py) Where on the node its classfied in Ansible tower as extra data for things that are passed at runtime. 

So in this case a Workflow node passes extra data Along with the Job templates original extra vars. and combines them to one item that is consumed. 

Extra data is only passed when Prompt on box is checked in extra vars, on a job template, and when specified when creating that workflow. Its in addition to the workflow extra vars as well.

aoyawale (approved):

aoyawale (approved):

aoyawale:
approved

playbooks/tower_configs_export_model/tower_workflows.yml
sean-m-sullivan: It raises hell when its included and the template hasn't been set to prompt for extra vars.

roles/workflow_job_templates/tasks/add_workflows_schema.yml
sean-m-sullivan: Got a screenshot of the underlying documentation, so including things here, its extra vars but its called extra data in the api. 
extra_data1
extra_data_api

kedark3: I think the difference in my opinion is `extra_vars` is for WF and `extra_data` is node level. Is that right summary? Maybe that's something you can add into README ? :)

sean-m-sullivan: yes. Will add something here soon.

sean-m-sullivan: Updated Readme

PR #59, opened 2020-07-17
Fixing job template ask credentials on Launch

kedark3:
### What does this PR do?
Ask credential was launch was looking at `tower_templates_item.ask_inventory_on_launch` which is wrong, this PR fixes that.

Re-running smoke tests would suffice.


sean-m-sullivan (approved):
Good Catch!

PR #64, opened 2020-07-25
Tower settings update to use settings option along with some other refactors

### LINKED ISSUE #62: Tower Settings role needs rewritten to use faster process
Tower Settings should add the settings option and use it to send all the settings in one task without a loop. This will speed up the process of the role. 
```
- name: Set all the LDAP Auth Bind Params
  tower_settings:
    settings:
      AUTH_LDAP_BIND_PASSWORD: "password"
      AUTH_LDAP_USER_ATTR_MAP:
        email: "mail"
        first_name: "givenName"
        last_name: "surname"
```

sean-m-sullivan:
### What does this PR do?
Updates tower settings to both use the settings options and the name/value option. Name/value is slower and depends on loop. Role has been setup to do either using a when to detect when a list is used.

Automated tests updated to test both scenarios.

#62 

Include file was needed, as the loop was evaluated before the when, this triggering errors, Tried a few different options but this was the one that works.


playbooks/tower_configs/tower_settings_individuale.yml
```diff
@@ -0,0 +1,46 @@
+---
+tower_settings_individuale:
```
kedark3, line 2: So both this and new format will work for the role?

roles/inventory_sources/README.md
```diff
@@ -1,16 +1,140 @@
...
+        method: POST
+        user: "{{ tower_username }}"
+        password: "{{ tower_passname }}"
+        force_basic_auth: yes
+        status_code: 201
+        validate_certs: no
+      register: user_token
+      no_log: True
+
+    - name: Set Tower oath Token
```
kedark3: this has got some typo "oath" vs "oauth"

roles/tower_settings/tasks/main.yml
```diff
@@ -3,14 +3,16 @@
...
     validate_certs:      "{{ validate_certs | default('false') }}"
-  loop: "{{ tower_settings }}"
-  loop_control:
-    loop_var: tower_setting_item
+  when: tower_settings.settings is defined
+  no_log: "{{ tower_configuration_settings_secure_logging }}"
+
+- name: use loop when tower_settings is a list
+  include_tasks: "settings_list.yml"
+  when: tower_settings[0] is defined
```
kedark3, line 17: I see this preserves backward compat. Nice. Is there a better way to test this though? instead of tower_settings[0]? I get the logic but I would assume there must be better way to check this in jinja2 format?

roles/workflow_job_templates/tasks/add_workflows_schema.yml
```diff
@@ -3,6 +3,16 @@
 - name : Create the Workfow Node
   awx.awx.tower_workflow_job_template_node:
     identifier :                    "{{ workflow_loop_node.identifier }}"   # internal identification ID
+    extra_data:                     "{{ workflow_loop_node.extra_data | default (omit) }}"
```
kedark3, line 6: I think I understand this but still for `extra_data` can you explain that more on this PR here or in Readme please? How it differs from `extra_vars`? In your words. Thanks.

kedark3 (changes requested):
Thanks Sean for the PR. Good work in here. Please see my comments.

playbooks/tower_configs/tower_settings_individuale.yml
sean-m-sullivan: Yes. and we test for both, the 2nd test using the way we are doing it before, uses this file.

roles/tower_settings/tasks/main.yml
sean-m-sullivan: I tried doing it like 15 different ways, this is the one that works. The issue being that the Loop variable check happens before the when, So even doing it in a block, it would fail because the loop var doesn't exist or something like it. This allows the when to take precedence.

roles/workflow_job_templates/tasks/add_workflows_schema.yml
sean-m-sullivan: This I think is back to the workflow PR. Can likely update that one instead of here. 

But basically there is no difference, but they've called it something else in the api endpoint and the module. 
extra_data_api

roles/inventory_sources/README.md
sean-m-sullivan: Fixed

roles/workflow_job_templates/tasks/add_workflows_schema.yml
kedark3: Got it. Updating in the workflow PR is fine. Thanks.

kedark3 (approved):

PR #72, opened 2020-08-19
Remove role defaults to enable role conditions

### LINKED ISSUE #68: Role tower_credential_input_sources should not have a default set.
The tower_credential_input_sources role has a default var `tower_credential_input_sources` set as an example. This causes the role to always be run when setting a conditional on the role as in the example playbooks.

This snippet will never be skipped.
```yaml
roles:  
   - {role: credential_input_sources, when: tower_credential_input_sources is defined, tags: credential_input_sources}
```

### LINKED ISSUE #69: Role user should not have default set
The users role has a default var `tower_user_accounts` set. This causes the role to always be run when setting a conditional on the role as in the example playbooks.

This snippet will never be skipped.
```yaml
roles:  
    - {role: users, when: tower_user_accounts is defined, tags: users}
```

### LINKED ISSUE #70: Role credentials has default value set
The credentials role has a default var `tower_credentials` set. This causes the role to always be run when setting a conditional on the role as in the example playbooks.

This snippet will never be skipped.
```yaml
roles:  
   - {role: credentials, when: tower_credentials is defined, tags: credentials}
```

### LINKED ISSUE #71: Role projects has a default var set
The projects role has a default var `tower_projects` set. This causes the role to always be run when setting a conditional on the role as in the example playbooks.

This snippet will never be skipped.
```yaml
roles:  
   - {role: projects, when: tower_projects is defined, tags: projects}
```

Tompage1994:
### What does this PR do?
Currently (As mentioned in #68) many roles have defaults set. This means we cannot do logic to include the roles as by adding `- {role: organizations, when: tower_organizations is defined, tags: organizations}`, `tower_organizations` will always be set as defaulted in the role.

It doesn't actually make sense to have this as a default anyway (in my opinion) so this PR removes defaults.

All of the roles (except 2, there will be another PR for that) have examples of how to run the role, and the sensible first point of call for finding out how to run is probably the documentation (the README) rather than the defaults.

The automation *should* still work, and to test this specifically you can remove the pre-tasks section of `playbooks/configure_tower.yml` and observe that all roles are skipped.

resolves #68 
resolves #69 
resolves #70 
resolves #71 

N/A


aoyawale (approved):

sean-m-sullivan (approved):
Looks good and now awx didn't time out.

PR #86, opened 2020-09-10
Workflow bugfix

sean-m-sullivan:
### What does this PR do?
Fix playbook to not skip workflows.


Tompage1994 (approved):
Makes much more sense than your late PR last night... 👍

PR #91, opened 2020-09-22
Update to add oauth to roles

sean-m-sullivan:
### What does this PR do?
Update to add oauth to roles


changelogs/fragments/auth_update.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - updated various roles to include oauth token and tower config file.
```
Tompage1994: Linting fails here. In fairness we could *probably* safely ignore linting in this fragments directory but also probably best to just follow the standards everywhere. So if you wouldnt mind adding a new line that would be great and then this should be good to go.

Tompage1994 (changes requested):
LGTM, just the one minor linting fix then good to go

Tompage1994 (approved):

PR #93, opened 2020-09-24
Add ability to use variables in notifications

sean-m-sullivan:
### What does this PR do?
Add ability to use variables in notifications

Added tests

#Other notes
Assumes any instance of two space __ are used for adding variables and removes them. Does not effect single space.


Tompage1994 (approved):
Awesome. This looks good to me

PR #94, opened 2020-10-02
fix notification task name

coolbry95:
### What does this PR do?
I have changed the name of the notifications task so that it refers to the proper name. Without this change a fresh install of tower_configuration will fail to run because awx.awx.tower_notification does not exist but awx.awx.tower_notification_template does.

Not completely sure if the automated tests will get this as I do not know what the automated tests do for this project.

This is the command I was using when I ran into this issue.
```
ansible-playbook configure_tower.yml --tags hosts
```

resolves #<number>

### DIFF
--- a/roles/notifications/tasks/main.yml
+++ b/roles/notifications/tasks/main.yml
@@ -1,6 +1,6 @@
 ---
 - name: Add ansible tower notifications
-  awx.awx.tower_notification:
+  awx.awx.tower_notification_template:
     name:                           "{{ tower_notifications_item.name }}"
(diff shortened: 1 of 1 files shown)


Tompage1994:
@coolbry95 thanks for this. I can see that awx.awx v15.0 was released 2 days ago and that changed `tower_notification` to `tower_notification_template`.

The tests shouldn't need to be changed as they test an action is made against an AWX instance and pull the latest `awx.awx` collection

Thanks for the contribution!

Tompage1994 (approved):

PR #98, opened 2020-10-07
Fix auth notifications

sean-m-sullivan:
### What does this PR do?
Fix notifications role to do auth correctly instead of default value.


Tompage1994 (approved):
lgtm

PR #107, opened 2020-10-21
Fix the defaults for survey, survey_spec to default omit

kedark3:
### What does this PR do?
WF Role was failing if the survery is not enabled or survey_spec is not passed.

Basic testing calling WFs role should be good. As long as survey and survey_spec is not passed.


sean-m-sullivan:
I've seen issues with this in the past, on other ones. It should default to false, not omit on survey enabled, will want to test this as well because of other odd behaviours with omit.

sean-m-sullivan:
tested with 
```
    survey_enabled:                     "{{ workflow_loop_var.survey_enabled | default('false') }}"
    survey:                             "{{ workflow_loop_var.related.survey_spec | default( workflow_loop_var.survey_spec | default(omit)) }}"
```
and worked without issue,

Tompage1994 (dismissed):
Like the addition of the new test case. This is something we should probably ensure we add when we fix/change functionality like this.

Only request is could you add a changelog fragment for this change as a bug fix. Then will be good to go

kedark3:
@Tompage1994 sure let me check it out. I haven't been able to keeping up enough with repo to know that unfortunately.

sean-m-sullivan:
@kedark3 Copy this file, change the filename, and change the text to describe what you did, 
https://github.com/redhat-cop/tower_configuration/blob/devel/changelogs/fragments/auth_update.yml

sean-m-sullivan (approved):
Looks good now.

PR #108, opened 2020-10-22
Update changelog and release

sean-m-sullivan:
### What does this PR do?
Update changelog  to be verbose to fix future errors when running in release
Updated fragment to not contain a dict that broke changelog


Tompage1994 (approved):
Lgtm

PR #117, opened 2020-11-17
Make sure defaults/main.yml is complete in all roles

ericzolf:
### What does this PR do?
This makes sure that one can copy the file in one's inventory and just adapt it to one's needs.
Also the when for each role's task isn't necessary as the default for the variable list is
an empty list, meaning an empty loop, i.e. one skip message less.

TBD, it's still work in progress

no

n/a


Tompage1994:
There is a reason why we didn't default the values of `tower_credentials` or the like. This is because when we include the role we include it by:
```
- {role: credentials, when: tower_credentials is defined, tags: credentials}
```
meaning that if there was a default in the role, then the flow will proceed into the role.

I agree we should ensure the role default are all correct but I think we should not set a default for the `tower_*****` values.

sean-m-sullivan:
I like this PR, I love the documentation. Thank you for this. 

Not sure on the removal of whens, I think we put it in so you could do a partial update still calling all roles, but that could just be defined at a play book level. 
@Tompage1994 what do you think?

Tompage1994:
> I like this PR, I love the documentation. Thank you for this. 
> 
> 
> 
> Not sure on the removal of whens, I think we put it in so you could do a partial update still calling all roles, but that could just be defined at a play book level. 
> 
> @Tompage1994 what do you think?

Definitely agree I like it in general. Just need to put thought into consequences of adding the defaults in which is something we had previously removed.

Tompage1994:
In regards to the removal of whens, it makes no difference it being there as it will always be defined because of the fact that the value is. defaulted to `[]`. So happy to remove that

ericzolf:
> In regards to the removal of whens, it makes no difference it being there as it will always be defined because of the fact that the value is. defaulted to `[]`. So happy to remove that

Exactly! Less code and it avoids useless skips. What do you want more :-)

The PR is _almost_ finished, I gave up for now on the workflow templates, I first need to understand them before modifying this role (it'll be another PR), and I need to make my code pass the pipeline. But this is something for next week. Enjoy your week-end!

ericzolf:
I guess that one succeeding integration test is sufficient as the other one is a timeout... Asking now for final review.

sean-m-sullivan (approved):
Looks good, sorry it took so long

Issue #124, opened 2020-12-04
Various Fixes needed to match awx


sean-m-sullivan:
Here is a list of roles that need fixed and what needs added.

1. Credentials
-  update_secrets
2. Inventory Source
- enabled_var
- enabled_value
- host_filter
3. license
- change from licence to manifest
4. users
- update_secrets
5. workflows - job nodes
- identifier needs format fixed with extra space.

The following role needs to be created:
- instance_group

sean-m-sullivan:
I've checked the other roles/modules to make sure all options match, these changes should bring us up to date.

ericzolf:
I wanted to take a chance at the license one, but the published version of awx.awx:15.0.1 doesn't yet support the manifest format.

tonyreveal:
Unless you've fixed it, which I don't see that you have, you also need to include a fix for updating an Organization with a Galaxy credential for  [Issue 123](https://github.com/redhat-cop/tower_configuration/issues/123)

sean-m-sullivan:
16.0 is up on galaxy, it should be the one we are pulling when doing builds.

Issue #121, opened 2020-11-25
workflow templates role doesn't have a defaults/main.yml describing all variables


ericzolf:
I didn't fully document all possibilities in PR #117 because I never used the role, it's rather complex and I wasn't sure how to document it properly. Someone with more experience might take this issue, else I'll do it once I need the role :smirk:

Issue #115, opened 2020-11-08
Support ansible.tower as well as awx.awx in roles


ericzolf:
I wasn't aware but you can [define a search list for collections in roles](https://docs.ansible.com/ansible/latest/user_guide/collections_using.html#using-collections-in-roles). That means that we should be able to remove the explicit `awx.awx` in all roles by adding something like the following to `meta/main.yml`:

```
collections:
- ansible.tower
- awx.awx
```
(assuming we want `ansible.tower` to be the default, and entries are ignored if they're not installed). That would make reuse in an enterprise context much easier.

What do you think?

sean-m-sullivan:
I think its a good idea, but I'm wary of it using the internal modules that were a part of awx.awx that were packaged with 2.9, and can't remember how they did it for 2.10. So testing is needed. I believe we want to move off of calling them explicitly, and/or set minimum versions for each with each release.

tonykay:
I would be cautious, and it is against the Ansible Recommendations. For example there are `roles` called `users`, `credentials` etc. As soon as you start using this collection in a much larger project where a user may do something like the below you have no idea where `users` is coming from:
```
collections:
    - foo.foo.
    - foo.bar
    - community.git
    - ansible.tower
    - awx.awx
```

One solution is to add context to `role` names `tower_configuration_users`

ericzolf:
Perhaps one step back on what we're trying to reach: collections hosted by the CoP  should (IMHO always) be usable internally (IT. demos, agnosticd, etc), but also be usable by Consultants at customer so that they can deliver their work more efficiently and provide more value in the same time for our customers (and make it easy for them to provide back improvements). If internally, we don't care too much about support, we do at customer site, hence need to use ansible.tower and not awx.awx (our collection remains unsupported, no question).

If we all agree on the objectives, we can continue to discuss :smirk: I think that the renaming suggested by @tonykay is good, we should just chose a prefix not colliding (potentially at least) with the awx.awx/ansible.tower content, so `tower_` doesn't sound a good plan to me but I can't come with a better idea right now... something like `towerconf_users` perhaps?

sean-m-sullivan:
Agree, we can rename them, and see about doing testing around this. I'd like to see it be agnostic, but could maybe even build the collection upload to do one or the other.

sean-m-sullivan:
I just did testing on this, verifying that:
A. No default modules are included such as tower_label from the ansible install
[...]

Issue #87, opened 2020-09-16
defaults/vars descriptions in role documentation don't explicitly state a list of dicts with these vars must be passed


mpetrive-rh:
For example https://github.com/redhat-cop/tower_configuration/tree/devel/roles/credentials

consider clarifying the data structure format to ensure the end user realizes they need to pass a list of dictionaries

PR #129, opened 2020-12-25
FQDN prep, additional workflow, other issues fixed

sean-m-sullivan:
### What does this PR do?
Fixes several issues, brings everything up to date with awx.awx. 
Updated workflow role to have better documentation in the defaults
Updated parts for meta/main that allows for removal of fqdn to allow for Either awx.awx or ansible.tower to be used for the roles, however there is a failure for ansible lint to qualify modules that aren't referenced by fqdn inside of roles. 

The big change is the addition of some new runners and a new workflow using them.
two runners have been added one for awx, one for tower. 
These runners are hosted on a private Openshift install and are able to deploy awx/tower to openshift and do testing, hopefully decreasing the amount of failures that we have in the workflows, and ensuring compatibility with BOTH awx/tower now. 

The testing_playbooks_awx.yml workflow is this change for now, From what I have read it will fail initially because new PRs/workflows cannot access the project secrets for security reasons, while that workflow works 100% in my branch, believe it should and we will diagnose with this test balloon. 

This should use the workflow actions already in place and test accordingly. 

#115 Partially - this has the prework for this, but ansible-lint fails due to an issue mentioned in that issue. 
#121 / #87 All defaults are described for the workflow role and checked other roles another once over. 
#124 Added missing options, updated docs and other parts in roles.


Tompage1994 (approved):
Looks good. Could you set up the secrets `openshift_username` etc. so the pipelines work and then rerun?

Then happy to merge

sean-m-sullivan:
@Tompage1994 The secrets are already setup and ready to run, but its a security thing that it doesn't run until commited at first to disallow extraction or other malicious acts from a random PR.

Tompage1994:
Looking at the run it looks like it tried to run but failed because hasn't got openshift_username. Looking at the secrets I think it needs changing to that rather than openshift_user?

sean-m-sullivan:
Here is the first run, and it succeeded as I thought it would, once merged, and should for any future PR.
https://github.com/redhat-cop/tower_configuration/runs/1669361783?check_suite_focus=true

PR #133, opened 2021-01-15
Survey options updated

sean-m-sullivan:
### What does this PR do?
Updated survey options, also ran into problem with the openshift cluster, and changing username for now.


.github/tower_install_inventories/openshift_awx
```diff
@@ -6,7 +6,7 @@ dockerhub_base=ansible
 
 openshift_host=https://api.crc.testing:6443
 openshift_project=awx-github
-openshift_user=developer
+openshift_user=kubeadmin
```
Tompage1994: Could you add this to the changelog. I'm not 100% sure I agree we should change to kubeadmin as that probably isn't good practice. I guess there isnt necessarily a sensible and usable default (maybe we could pick it up from the kubeconfig or something?) Though if you update the changelog for now we can think about that in the future.

roles/job_templates/README.md
```diff
@@ -65,6 +65,7 @@ tower_configuration_job_templates_secure_logging defaults to the value of tower_
 |`ask_credential_on_launch`|""|no|bool|Prompt user for credential on launch.|
 |`survey_enabled`|""|no|bool|Enable a survey on the job template.|
 |`survey_spec`|""|no|dict|JSON/YAML dict formatted survey definition.|
+|`survey_spec`|""|no|dict|JSON/YAML dict formatted survey definition. Alais of survey_spec|
```
Tompage1994: I assume this should be
`|`survey`|""|no|dict|JSON/YAML dict formatted survey definition. Alias of survey_spec|`

Tompage1994 (changes requested):
Looks good. Just a couple of small things

.github/tower_install_inventories/openshift_awx
sean-m-sullivan: Even better I found how to grant the user the right privileges so we can use developer and not make this change.

roles/job_templates/README.md
```diff
@@ -65,7 +65,7 @@ tower_configuration_job_templates_secure_logging defaults to the value of tower_
 |`ask_credential_on_launch`|""|no|bool|Prompt user for credential on launch.|
 |`survey_enabled`|""|no|bool|Enable a survey on the job template.|
 |`survey_spec`|""|no|dict|JSON/YAML dict formatted survey definition.|
-|`survey_spec`|""|no|dict|JSON/YAML dict formatted survey definition. Alais of survey_spec|
+|`survey_spec`|""|no|dict|JSON/YAML dict formatted survey definition. Alias of survey_spec|
```
Tompage1994: I think this should be `survey` rather than `survey_spec` shouldn't it?

Tompage1994 (changes requested):
Just the one README correction then I'm happy

roles/job_templates/README.md
sean-m-sullivan: One of them should, fixing.

Tompage1994 (approved):
LGTM

PR #134, opened 2021-01-15
Fix credentials defaults in hosts and users roles

### LINKED ISSUE #132: Using environment variables/config file for authentication does not work
The default values for e.g. `tower_username` and `tower_oauthtoken` as defined in the roles in this collection are empty strings, which causes an empty string to be passed to the various `awx.awx` modules instead of null. This results in requiring that e.g. `tower_oauthtoken` is explicitly defined somewhere in Ansible rather than inheriting its value from the appropriate env.

ericzolf:
### What does this PR do?

- make the default for the username to be omitted instead of admin
- fix a few spurious defaults (we should use the default of the module not overwrite it without reason)
- make the default for the password to be omitted
- correct a few booleans expressed as strings
- add a default password for all users, it can be set empty and be ignored as well

I will trust the automated tests :smile: 

resolves #132 (I assume, it's not yet absolutely clear)

n/a


ericzolf:
@sean-m-sullivan the pipeline failed on the OpenShift login, could it be the secret issue you referred to yesterday? This said, I don't want to exclude the possibility that the issue is on my side.

sean-m-sullivan:
The issue is I had an issue yesterday, and had to rebuild the openshift cluster, so that test will fail for the two active prs, till we merge them

PR #135, opened 2021-01-22
Remove requirement for awx.awx and allow users to choose either awx or ansible.tower

sean-m-sullivan:
### What does this PR do?
Updates tower_configuration to be in line with standards with ansible-lint v5.0.0 which is in pre-release
Uses feature of ansible-lint v5 to be able to check collections, roles, and modules that was not in v4. 
Removes awx.awx prefix from all roles and moves declaration to playbook

Automated tests have been updated to use this change. 

#115 support ansible.tower as well as awx.awx

TBD after this is to add a build for standing up a full ansible tower, and testing ansible.tower against it in addition to awx.

This uses a personal branch of ansible-lint that is a clone of devel ansible-lint, when v5 is actually released or a new beta release is made, we should then switch it back to the main repo.


sean-m-sullivan:
Closing this until I get the awx enviroment up and tested

PR #139, opened 2021-01-28
Update main.yml

axil76:
add parameters to import container group

### DIFF
--- a/playbooks/tower_configs/tower_instance_groups.yml
+++ b/playbooks/tower_configs/tower_instance_groups.yml
@@ -1,3 +1,20 @@
 ---
 tower_instance_groups:
   - name: test_instance_group
+  
+  - name: Runner Ansible 2.8
+    credential: 61
+    is_containerized: true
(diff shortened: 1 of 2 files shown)


Tompage1994 (changes requested):
Thanks for the contribution.

I've taken a look and it doesn't look as though the two options you've added to the `instance_group` module are available on `awx.awx.instance_group`. This makes me think this will not work.

Unfortunately our CI is down at the moment and I don't have a Tower instance to try this against.

There are also a few trailing whitespace issues which means linting is not passing.

sean-m-sullivan:
Going to do a quick PR to fix the integration test, hopefully it fixes the openshift one too.

sean-m-sullivan:
Ok, made another PR with the fix for the integration tests, but looking at the backend, this should be something done on the module level first, and we then can integrate it into the role. 

However looking at 
https://github.com/ansible/awx/blob/devel/awx_collection/plugins/modules/tower_instance_group.py
and the tower api backend, I can note that we should not be doing summary fields, as those are automatically generated based on inputs.

In addition, doing testing, it looks like the following is true
Running the following creates a standard instance group
```
    - name: Update an Instance Group
      tower_instance_group:
        name: "{{ result.id }}"
        policy_instance_percentage: 34
        policy_instance_minimum: 24
        state: present
      register: result
```

Running the code next, will set "is_containerized" to true, by virtue of the openshift credential being specified. 
```
    - name: Create a container group
      tower_instance_group:
        name: "{{ group_name2 }}"
        credential: "{{ cred_result.id }}"
      register: result
```

Based on this, the module/role likely needs to have documentation updated, however I don't think the parameters for invoking it need any additions.

sean-m-sullivan:
Closing this as we have not heard back. These roles use the awx.awx/ansible.tower modules, and thus cannot have fields those modules don't support, The Kubernetes credential can be created wit the credential role, and the instance group with pod spec override can currently be created with the instance group role. The is_containerized and summary fields are not needed in this respect. 

Thanks for your contribution, and we welcome any changes proposed in the future.

PR #140, opened 2021-02-08
Update to remove awx.awx from module names

### LINKED ISSUE #115: Support ansible.tower as well as awx.awx in roles

sean-m-sullivan:
### What does this PR do?
This PR updates the yaml and ansible lint to their latest versions, and by extension allows us to remove awx.awx from module names and still pass ansible lint. These are then tested against a deployment of awx, and have been tested on tower.

resolves #115


changelogs/fragments/ansible_lint.yml
```diff
@@ -0,0 +1,5 @@
+---
+major_changes:
+  - Updated collection to use and comply with ansible-lint v5
+breaking_changes:
+  - Updated to allow use of either awx.awx or ansible.tower
```
Tompage1994, line 5: Is it actually a breaking change? Will old playbooks fail? If so it may be worth having a major change log in the README so anybody who comes back to the collection is aware of what needs to be changed in their playbooks

Tompage1994 (approved):
Just one question to check but otherwise happy 👍

changelogs/fragments/ansible_lint.yml
sean-m-sullivan: Without the opening playbook listing awx.awx, its not going to find the right modules specifcally for applications and tower notification template. 

Did a deep dive and for those that were included in regular ansible 2.9, they map it to the new awx.awx modules, and those aren't mapped as they didn't come out till later.

Tompage1994: ok, probably worth calling out in the README that the collection needs to be specified in the playbook?

sean-m-sullivan: Updated the base readme under using the collection, Should we update it elsewhere?

sean-m-sullivan (commented):
Reviewed all the files again, fixed a few errors, I think its ready.

sean-m-sullivan:
After rereading yaml lint rules, and seeing as warnings were not showing up easily, I've updated the lint rules to fit our criteria, and also adjusted the colon rule and several spaces before colon areas we were missing.

README.md
```diff
@@ -27,6 +27,20 @@ collections:
 ```
 
 ## Using this collection
+The awx.awx or ansible.tower collection must be invoked in the playbook in order for ansible to pick up the correct modules to use.
+
+Otherwise it will look for for the modules only in your base installation. If there are errors complaining about "counldn't resolve module/action' this is the most likely cause.
```
Tompage1994: `Otherwise it will look for for the modules only in your base installation. If there are errors complaining about "couldn't resolve module/action" this is the most likely cause.`

Tompage1994 (changes requested):
Just a single minor typo in the readme. Probably best to make sure the main readme is as good as possible. After fixing that Im completely happy for this to be merged

README.md
```diff
@@ -29,7 +29,7 @@ collections:
 ## Using this collection
 The awx.awx or ansible.tower collection must be invoked in the playbook in order for ansible to pick up the correct modules to use.
 
-Otherwise it will look for for the modules only in your base installation. If there are errors complaining about "counldn't resolve module/action' this is the most likely cause.
+Otherwise it will look for the modules only in your base installation. If there are errors complaining about "counldn't resolve module/action' this is the most likely cause.
```
Tompage1994: That wasn't even the typo I spotted 😂

Could you replace `counldn't` with `couldn't` and then the final single quote should be a double quote

Tompage1994 (changes requested):

README.md
sean-m-sullivan: Whoops on it.

README.md
sean-m-sullivan: Thought you were just highlighting missed what you pointed out later.

Tompage1994 (approved):
Looks good now

PR #142, opened 2021-02-09
this is to enforce the required fields for each module

djdanielsson:
### What does this PR do?
sets "mandatory" for the required fields

I think automated tests should work fine for this.

no

none

### DIFF
--- a/roles/inventory_sources/tasks/main.yml
+++ b/roles/inventory_sources/tasks/main.yml
@@ -1,12 +1,12 @@
 ---
 - name: Add an inventory source
   awx.awx.tower_inventory_source:
-    name:                                 "{{ source.name }}"  # REQUIRED
+    name:                                 "{{ source.name | mandatory }}"
     new_name:                             "{{ source.new_name | default(omit) }}"
     description:                          "{{ source.description | default(omit) }}"
-    inventory:                            "{{ source.inventory.name | default( source.inventory ) }}" # REQUIRED
+    inventory:                            "{{ source.inventory.name | default( source.inventory ) | mandatory }}"
     organization:                         "{{ source.inventory.organization.name | default( source.organization | default(omit)) }}"
-    source:                               "{{ source.source }}"  # REQUIRED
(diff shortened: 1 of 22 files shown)


ericzolf:
I never had to use `mandatory`, for me it's only necessary if you want to make a variable mandatory which isn't mandatory according to the module, which isn't the case here. The module would choke for us if the variable isn't defined.
It doesn't really hurt but it's more code to maintain. What would be the added value?

djdanielsson:
clarity when looking at the role which parts are required plus it would also remind the user that its a required field when it errors on them.

roles/inventory_sources/tasks/main.yml
```diff
@@ -1,12 +1,12 @@
...
   awx.awx.tower_inventory_source:
-    name:                                 "{{ source.name }}"  # REQUIRED
+    name:                                 "{{ source.name | mandatory }}"
     new_name:                             "{{ source.new_name | default(omit) }}"
     description:                          "{{ source.description | default(omit) }}"
-    inventory:                            "{{ source.inventory.name | default( source.inventory ) }}" # REQUIRED
+    inventory:                            "{{ source.inventory.name | default( source.inventory ) | mandatory }}"
     organization:                         "{{ source.inventory.organization.name | default( source.organization | default(omit)) }}"
-    source:                               "{{ source.source }}"  # REQUIRED
+    source:                               "{{ source.source }}"  # REQUIRED ? according to defaults this is not required
```
sean-m-sullivan, line 9: After looking at the gui, and doing a test of the API, This is a required field, updating the module to fix it over at awx, good catch!

sean-m-sullivan (approved):
Looks good!, I've reviewed the changes and all of these are required fields in the modules, so have no issue making them mandatory.

roles/inventory_sources/tasks/main.yml
sean-m-sullivan: PR here: https://github.com/ansible/awx/pull/9284

PR #146, opened 2021-02-12
Linguist fix

djdanielsson:
### What does this PR do?
forcing linguist to read yml/yaml files as a language so it stops calling the repo html

Automated

no

none

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+*.yml linguist-detectable
+*.yaml linguist-detectable
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,2 @@
 *.yml linguist-detectable
 *.yaml linguist-detectable
-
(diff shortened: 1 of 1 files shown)


djdanielsson:
I don't understand how it failed the lint, the extra line is there you can see that in the 2nd commit...

.gitattributes
```diff
@@ -0,0 +1,3 @@
+*.yml linguist-detectable
+*.yaml linguist-detectable
+
```
Tompage1994: This is one too many lines at the end. The github UI only seems to show one of them but I've checked out your code and can see you have blank line on lines 3 and 4. If you delete line 4 then it should all pass linting.

Tompage1994 (changes requested):
Thanks for the contribution. Just the one comment on how to fix the linting error

.gitattributes
sean-m-sullivan: Removed it, we will see how the testing goes :)

sean-m-sullivan (approved):
Looks good, That fixed it. Didn't even know about this.

Issue #141, opened 2021-02-09
Treat empty strings as unset?


DWSR:
Came up in the discussion around #132, should empty strings (i.e. `""`) be omitted (or use the default value) when passed to modules in some or all cases?

As a consumer of this collection, there's a couple of spots where I think an intentional empty string is a "valid" value (e.g. descriptions), but most of the time I suspect an empty string will be unintentional, especially when used for credentials or connection info.

sean-m-sullivan:
I definitly can see this for the tower host/username/password. And as you said description makes sense, Extra vars, surveys, and other fields that could stay the same if not overwritten with "" should use the default, omit/true like the username/password/host. 

The more I look at it, I agree with you on the better way of doing things, just was working on another PR, and other things, so just dont have the bandwidth atm to do it. However we should probably fix it before the next version, as it should be a larger version change due to the removal of awx.awx. prefixes

sean-m-sullivan:
Another person came up with an issue where tower export has strings as "" or null, and believe this can fix it, will do some testing, but in that case things like Inventory, extra vars, and a host of other vars that have omit should have this, Not just user/pass

Tompage1994:
@sean-m-sullivan this can presumably be closed now?

sean-m-sullivan:
Yes.

PR #147, opened 2021-02-13
Omit change

### LINKED ISSUE #132: Using environment variables/config file for authentication does not work

sean-m-sullivan:
### What does this PR do?
Changed the omit behaviour to not send empty strings in some cases. 
This allows for empty items to be passed as omit rather then empty strings or false in some cases. 
In addition all boolean fields have been converted into either true or false , rather then omit in some cases which could cause problems when the user entered false. 
Also fixed some grammar problems so that all true/false omits are single quoted.
Also fixed it so that all verify_ssl fields are True by default rather then a handful being false by default.
I also fixed some tests that were failing with the new defaults, Such as some projects weren't waiting to be updated, and an empty dict in individual settings.

#141 and closed #132 , the later has a breakdown of how omit functions with the true flag.


roles/ad_hoc_command_cancel/tasks/main.yml
```diff
@@ -3,16 +3,16 @@
 - name: Cancel Tower ad hoc command
   tower_ad_hoc_command_cancel:
     command_id:             "{{ __ad_hoc_command_cancel_item.id }}"
-    fail_if_not_running:    "{{ __ad_hoc_command_cancel_item.fail_if_not_running | default(omit) }}"
-    interval:               "{{ __ad_hoc_command_cancel_item.interval | default(omit) }}"
-    timeout:                "{{ __ad_hoc_command_cancel_item.timeout | default(omit) }}"
+    fail_if_not_running:    "{{ __ad_hoc_command_cancel_item.fail_if_not_running | default('false') }}"
+    interval:               "{{ __ad_hoc_command_cancel_item.interval | default(1) }}"
```
Tompage1994, line 7: Just a question on whether/how we decide to default values. 

Should this really be defaulted to 1 rather than omitted? What if the awx collection changes its default for some reason?

roles/project_update/tasks/main.yml
```diff
@@ -3,17 +3,17 @@
 - name: Run Tower project update
   tower_project_update:
     name:               "{{ __project_update_update_item.name }}"
-    organization:       "{{ __project_update_update_item.organization | default('') }}"
-    wait:               "{{ __project_update_update_item.wait | default(omit) }}"
-    interval:           "{{ __project_update_update_item.interval | default(omit) }}"
-    timeout:            "{{ __project_update_update_item.timeout | default(omit) }}"
+    organization:       "{{ __project_update_update_item.organization | default(omit, true) }}"
+    wait:               "{{ __project_update_update_item.wait | default('true') }}"
```
Tompage1994, line 7: The other updates we default to false. Is this intended or should they all follow the same?

Tompage1994 (changes requested):
Mostly looks good. (Though not the easiest to review). I've added a couple of comments. The main question I have is around where we are defining a default other than 'omit' and whether this is the right thing to be doing

roles/ad_hoc_command_cancel/tasks/main.yml
sean-m-sullivan: So on all of these, I kept the module default on most of them, in this case the module defaults to 1, so it matches. The omit on True/False is problematic, if you put false and default omit, it omits it. So a user would not be able to turn a feature off once turned on.

roles/project_update/tasks/main.yml
sean-m-sullivan: The omit on True/False is problematic, if you put false and default omit, it omits it. So a user would not be able to turn a feature off once turned on. There is a good breakdown in #132

Tompage1994: What I meant was that on the other 'wait' options on modules we have `default('false')` rather than `default('true')` eg. on `ad_hoc_command`, `inventory_source_update`, `job_launch` etc.

roles/ad_hoc_command_cancel/tasks/main.yml
Tompage1994: I get that. The issue I can see though is if `awx.awx` changes its default, we will not also do so. So that may make sense on boolean values, but values such as `interval` they may change the default to `5` for whatever reason and ours is hardcoded to 1. 

Its a minor issue but one we should think about.

sean-m-sullivan: At least for these, The interval is only for the module running, it doesn't actually post that to tower, going to look over these, agree with you on things like forks and others though. Anything posted to tower should default to omit so not change things unless specified.

roles/project_update/tasks/main.yml
sean-m-sullivan: For project update, [its default true](https://github.com/ansible/awx/blob/67f1f9ac6904b800165ef1e4705f551fd39b60e0/awx_collection/plugins/modules/tower_project_update.py#L35), and isn't posted to tower, and with inventory update, it [defaults to false](https://github.com/ansible/awx/blob/67f1f9ac6904b800165ef1e4705f551fd39b60e0/awx_collection/plugins/modules/tower_inventory_source_update.py#L41)

sean-m-sullivan: For project update, [its default true](https://github.com/ansible/awx/blob/67f1f9ac6904b800165ef1e4705f551fd39b60e0/awx_collection/plugins/modules/tower_project_update.py#L35), and isn't posted to tower, and with inventory update, it [defaults to false](https://github.com/ansible/awx/blob/67f1f9ac6904b800165ef1e4705f551fd39b60e0/awx_collection/plugins/modules/tower_inventory_source_update.py#L41)

roles/ad_hoc_command_cancel/tasks/main.yml
sean-m-sullivan: I just checked through and changed forks on job templates, but did not find any other ones that would be like this, when we are setting these its not in tower, its for when the module is running and matches the module default.

Tompage1994 (approved):

roles/project_update/tasks/main.yml
Tompage1994: Ah makes sense.

PR #151, opened 2021-02-16
Do not use 'kind' when importing credential.

### LINKED ISSUE #150: Using deprecated field "kind" when importing credentials breaks the import if kind is cloud or external.
The [awx.awx.tower_credential module has deprecated the parameter `kind`](https://docs.ansible.com/ansible/latest/collections/awx/awx/tower_credential_module.html#parameter-kind) in favor or `credential_type`.

In [roles/credentials/tasks/main.yml:9](https://github.com/redhat-cop/tower_configuration/blob/c6b68596b18ff1579b2af028e56e927770c365a8/roles/credentials/tasks/main.yml#L9), this parameter is still provided to tower_credential, along the `credential_type` parameter.

An exported credential with awxkit 17.0.0 looks as follows:
```
tower_credentials:
- credential_type:
    kind: cloud
    name: '[DEMO] Demo Credential Type'
    type: credential_type
  description: ''
  inputs:
    demo_username: admin
  name: '[DEMO] Demo Credential'
  natural_key:
    credential_type:
      kind: cloud
      name: '[DEMO] Demo Credential Type'
      type: credential_type
    name: '[DEMO] Demo Credential'
    organization:
      name: Default
      type: organization
    type: credential
  organization:
    name: Default
    type: organization
```
Importing such credential will break since `kind` will have the value `cloud`, which is not authorized in [tower_credential.py from awx.awx](https://github.com/ansible/awx/blob/devel/awx_collection/plugins/modules/tower_credential.py#L281).

By simply removing [roles/credentials/tasks/main.yml:9](https://github.com/redhat-cop/tower_configuration/blob/c6b68596b18ff1579b2af028e56e927770c365a8/roles/credentials/tasks/main.yml#L9) and passing only `credential_type` (line 8 of the same file), the import for credentials of `cloud` or `external` types is correctly working.

Since `kind` is deprecared in awx.awx.tower_credential, perhaps it should be removed in the role.

ghurel-rh:
### What does this PR do?
This PR allows to import credentials of custom credential types, exported with awxkit (17.0.0), having the `kind` field value to `cloud` or `external`.

Create a custom credential type and a credential of that new custom type. Export the credential using awxkit and try to import it using this collection.

resolves #150 

### DIFF
--- a/changelogs/fragments/credential_kind_deprecated.yml
+++ b/changelogs/fragments/credential_kind_deprecated.yml
@@ -1,2 +1,4 @@
+---
 breaking_changes:
-  - Removed kind from to credentials role. This will be depreciated in a few months. Kind arguments are replaced by the credential_type and inputs fields. 
+  - Removed kind from to credentials role. This will be depreciated in a few months. Kind arguments are replaced by the credential_type and inputs fields.
+...
--- a/changelogs/fragments/credential_kind_deprecated.yml
+++ b/changelogs/fragments/credential_kind_deprecated.yml
@@ -1,0 +1,2 @@
+breaking_changes:
+  - Removed kind from to credentials role. This will be depreciated in a few months. Kind arguments are replaced by the credential_type and inputs fields. 
--- a/playbooks/tower_configs/tower_credentials.yml
+++ b/playbooks/tower_configs/tower_credentials.yml
@@ -1,12 +1,12 @@
 ---
 # Commented out content is serving as example for you to add new content.
 tower_credentials:
-  - kind: ssh
+  - credential_type: Machine
     name: Demo Credential
     inputs:
       username: username
     organization: Default
-  - kind: rhv
+  - credential_type: Red Hat Virtualization
     name: admin@internal-RHVM-01
     description: infra-rhvm-01 creds for inventory sources.
     inputs:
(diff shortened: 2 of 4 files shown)


ghurel-rh:
Ok, simply deleting the `kind` field is actually breaking the import of the credentials defined in [tower_configs/tower_credentials.yml](https://github.com/redhat-cop/tower_configuration/blob/devel/playbooks/tower_configs/tower_credentials.yml) since they do not have any `credential_type` field defined.

Not sure about how to handle the issue here, I see two options:
1. refactor the credentials defined in [tower_configs/tower_credentials.yml](https://github.com/redhat-cop/tower_configuration/blob/devel/playbooks/tower_configs/tower_credentials.yml) to not use `kind` anymore;
2. in [roles/credentials/tasks/main.yml#L9](https://github.com/redhat-cop/tower_configuration/blob/c6b68596b18ff1579b2af028e56e927770c365a8/roles/credentials/tasks/main.yml#L9) use a condition to define `kind` only when `credential_type.name` and `credential_type` are not defined:
```
kind:  "{{ ((tower_credentials_item.credential_type.kind | default( tower_credentials_item.kind)) if tower_credentials_item.credential_type.name is not defined and tower_credentials_item.credential_type is not defined ) | default(omit, true) }}"
```

I tested the option 2 locally and it seems to work, but I'll wait for your advice before pushing this fix on the PR.

sean-m-sullivan:
Change the tests, Tower 4.0.0 is coming out this summer, and the module is removing those options around then. 
```
 [DEPRECATION WARNING]: The kind parameter has been deprecated, please use 
credential_type instead. This feature will be removed in version 
ansible.tower:4.0.0. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
```
We haven't released yet, was going to sometime this week, and I believe the next release is going to break things for some people anyway, so whats one more, especially when its a depreciated feature.

ghurel-rh:
I've pushed the updated credentials and it works correctly now.

sean-m-sullivan:
Sorry, one last step, can you create a file in tower_configuration/changelogs/fragments/somefilename.yml
with the following:
```
breaking_changes:
  - Removed kind from to credentials role. This will be depreciated in a few months. Kind arguments are replaced by the credential_type and inputs fields. 
```

sean-m-sullivan (approved):
Looks great, and good catch on the kind, We should remove the other kind arguments, but that can be another PR for anyone.

sean-m-sullivan:
Great, thanks!

ghurel-rh:
Just added the changelog fragment file.
Thanks for the help and the reactivity, cheers!

Issue #154, opened 2021-02-17
Lint Pipeline Fails


Tompage1994:
Just spotted that the Lint pipeline has started failing due to a change in a dependency. This will need resolving

```
> Run sudo apt install software-properties-common

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

Reading package lists...
Building dependency tree...
Reading state information...
software-properties-common is already the newest version (0.98.9.3).
0 upgraded, 0 newly installed, 0 to remove and 2 not upgraded.
Hit:1 http://azure.archive.ubuntu.com/ubuntu focal InRelease
Get:2 http://azure.archive.ubuntu.com/ubuntu focal-updates InRelease [114 kB]
Get:3 http://azure.archive.ubuntu.com/ubuntu focal-backports InRelease [101 kB]
Hit:4 https://storage.googleapis.com/bazel-apt stable InRelease
Get:5 http://dl.google.com/linux/chrome/deb stable InRelease [1811 B]
Ign:6 https://repo.mongodb.org/apt/ubuntu focal/mongodb-org/4.4 InRelease
Get:8 https://repo.mongodb.org/apt/ubuntu focal/mongodb-org/4.4 Release [5384 B]
Hit:9 https://packages.microsoft.com/repos/azure-cli focal InRelease
Hit:11 https://download.mono-project.com/repo/ubuntu stable-focal InRelease
Get:12 https://packages.microsoft.com/ubuntu/20.04/prod focal InRelease [10.5 kB]
Hit:10 https://cli-assets.heroku.com/apt ./ InRelease
Get:13 https://repo.mongodb.org/apt/ubuntu focal/mongodb-org/4.4 Release.gpg [801 B]
Get:14 https://packages.cloud.google.com/apt cloud-sdk InRelease [6739 B]
Get:15 http://security.ubuntu.com/ubuntu focal-security InRelease [109 kB]
Ign:16 http://ppa.launchpad.net/ansible/ansible/ubuntu focal InRelease
Hit:7 https://packages.cloud.google.com/apt kubernetes-xenial InRelease
Hit:17 https://apt.postgresql.org/pub/repos/apt focal-pgdg InRelease
Hit:19 http://ppa.launchpad.net/apt-fast/stable/ubuntu focal InRelease
Ign:20 https://dl.bintray.com/sbt/debian  InRelease
[...]

PR #156, opened 2021-02-24
ignore changelog files

Tompage1994:
### What does this PR do?
The files in changelog are auto generated and don't contain `---` or `...` at the start and end so will be picked up by yamllint. This adds the changelog dir to be ignored by yamllint

Yamllint will pass the CI tests

N/A

Discovered on retesting for #154 and found these fails.

### DIFF
--- a/.github/workflow-config/.yamllint.yml
+++ b/.github/workflow-config/.yamllint.yml
@@ -1,6 +1,9 @@
 ---
 extends: default
 
+ignore: |
+  changelogs
+
 rules:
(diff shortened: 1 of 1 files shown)


Tompage1994:
The CI check will fail because the used config is from devel rather than this check. To check this config works, run 

```
yamllint -c .github/workflow-config/.yamllint.yml .
``` 

from the base of the repo

sean-m-sullivan (approved):
LGTM

PR #157, opened 2021-03-09
Remove dependency

crenwick93:
### What does this PR do?
AWX.AWX no longer required as a dependency, if you're pulling from private automation hub you may not have access to awx.awx and therefore it fails. 

Manually tested

N/A


Tompage1994 (approved):
Looks good. I've seen the reason for this change with @crenwick93 and can see its required when using AH. I think this dependency was kicking around from older versions when we only supported `awx.awx`

PR #162, opened 2021-03-24
Workflow schema

sean-m-sullivan:
### What does this PR do?
Adds option for workflow schemas to be used instead of workflow nodes lope. 
uses [This Awx.awx PR](https://github.com/ansible/awx/pull/9309)

Additinonal tests should be added to test this, I propose we use the tower_export data as it is already in this format, but as an additional test so we keep testing the now 3 ways to deploy schemas.


roles/workflow_job_templates/tasks/main.yml
```diff
@@ -21,6 +21,8 @@
     webhook_credential:                 "{{ workflow_loop_var.webhook_credential.name | default(workflow_loop_var.webhook_credential | default(omit, true)) }}"
     survey_enabled:                     "{{ workflow_loop_var.survey_enabled | default('false') }}"
     survey:                             "{{ workflow_loop_var.related.survey_spec | default( workflow_loop_var.survey_spec | default( workflow_loop_var.survey | default(omit, true))) }}"
+    schema:                             "{{ workflow_loop_var.related.workflow_nodes | default(omit, true) }}"
```
sean-m-sullivan: I am conflicted on this part of me wants to keep workflow_nodes not under related, but then it will die if anyone has entries under previous, related.workflow nodes is what export outputs, should we rename the old? 

or just update tests to use the new schema system, and trust the user to know the difference.

sean-m-sullivan:
Believe this should be ready for review, have made some changes, updated docs, and tested it. Talked with Marc about doing the variable name change, and think its the best of going about this to preserve backward compatibility.

Tompage1994 (approved):
lgtm

PR #167, opened 2021-03-31
Added condition so that add_workflows_schema.yml would not execute wh…

### LINKED ISSUE #166: workflow schema fails when state is absent
Modify your JSON or YAML definition of a workflow to set the workflow state to absent.  Then execute tower_configuration.workflow_job_template.  The play will fail because the task will still execute add_workflows_schema.yml even though the workflow was removed from Tower.

tonyreveal:
…en state is absent

### What does this PR do?
Added condition to the workflow_job_templates role so that add_workflows_schema.yml only executed when workflow_loop_var.state was present. 

resolves #166

### DIFF
--- a/roles/workflow_job_templates/tasks/main.yml
+++ b/roles/workflow_job_templates/tasks/main.yml
@@ -43,4 +43,5 @@
   loop: "{{ tower_workflows }}"
   loop_control:
     loop_var: workflow_loop_var
+  when: workflow_loop_var.state == "present"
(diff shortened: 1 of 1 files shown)


sean-m-sullivan:
Was looking through this, and thought for a minute it would require a list of whens for the WIP PR, but that should be an AND with this. so this is good

sean-m-sullivan (approved):
LGTM

PR #169, opened 2021-03-31
Workflow task name

### LINKED ISSUE #168: Consider renaming the workflow_job_templates task name from Create Workflow
View https://github.com/redhat-cop/tower_configuration/blob/devel/roles/workflow_job_templates/tasks/main.yml

The task name is "Create workflow" however this isn't always appropriate since the user can set the state to absent.  Maybe change the task name to something like "Manage Workflows" instead.

tonyreveal:
### What does this PR do?
Updated task name in workflow_job_templates since "Create" may not always be appropriate.

resolves #168

### DIFF
--- a/roles/workflow_job_templates/tasks/main.yml
+++ b/roles/workflow_job_templates/tasks/main.yml
@@ -1,6 +1,6 @@
 ---
 # Create Workflow
-- name: Create Workflow
+- name: Manage Workflows
   tower_workflow_job_template:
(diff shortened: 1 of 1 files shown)


sean-m-sullivan:
Can you remove the loop logic part on this PR? We can then do the approval separately, otherwise, no problem with this PR. Didn't think of this before.

sean-m-sullivan (approved):
LGTM with the other merged

PR #170, opened 2021-04-20
Fix job template survey table

ericzolf:
### What does this PR do?

The job template survey table was broken, it is now fixed.

I'm just missing the meaning of the `new_question` field. It's not documented in the API reference...

n/a

n/a

n/a


sean-m-sullivan (approved):
LGTM, good catch that we hadn't updated that.

PR #174, opened 2021-04-25
Update Tower configuration to use new minikube based install for testing awx

sean-m-sullivan:
### What does this PR do?
Update Tower configuration to use new minikube based install for testing awx
Finally figured out what was wrong, and have even submitted a bug for the operator with decreasing postgres defaults. Have this working locally and on my actions, it takes a bit to run, but its working!

The tests should hopefully work now


playbooks/tower_configs/tower_credentials.yml
```diff
@@ -1,11 +1,6 @@
 ---
 # Commented out content is serving as example for you to add new content.
 tower_credentials:
-  - credential_type: Machine
```
sean-m-sullivan, line 4: Not sure why, but the modules were complaining that two existed when trying to do Job templates, The rest of the demo objects didn't create duplicates, but removed this one till its figured out.

Tompage1994 (changes requested):
Generally looks good though looks like the checks are failing for a genuine cause. I guess that should be fixed by a separate PR?

sean-m-sullivan:
I am looking into it, and think I may have found a short term fix, and the problem itself should be solved overall by the Workflow Schema PR, once that is published soon, Its already in awx.awx devel.

sean-m-sullivan:
@Tompage1994 You may be right on rather then trying to fix all of our issues with this PR, and leaving it to a separate one. From what I am seeing its reliably getting us to awx now, including the many times I've run it leading to this. I'll revisit tomorrow, and we can discuss then if you want.

sean-m-sullivan:
Seems auth fails every so often on the settings, and the adhoc was unable to be canceled due to it being finished before I could cancel, so reduced the # of adhocs. But passing checks now. @Tompage1994

playbooks/tower_configs/tower_credentials.yml
Tompage1994: I suspect its probably because when AWX is spun up it has the Demo Credential (and presumably the other 'demo' objects) there.

Tompage1994 (changes requested):
Looks good to me.

Only change I would like to see is all the kubernetes definitions under a kube directory (e.g. `.github/workflow-config/kube/<file>`)

playbooks/tower_configs/tower_credentials.yml
sean-m-sullivan: It shouldn't but somehow its slightly different somehow, but I couldn't see it.

Tompage1994 (approved):

PR #178, opened 2021-04-30
add option to use jinja templating to inventories, hosts, and groups

sean-m-sullivan:
### What does this PR do?
add option to use jinja templating to inventories, hosts, and groups
A user came to me to figure out how to do this, and found the credential type fix worked.

It has been tested and works like our credential type change.


Tompage1994 (changes requested):
LGTM

Could you add an example in the CI/CD playbooks so we can run a test for this. (I know thats not necessarily working atm but will at least give us confidence going forwards.)

Tompage1994 (changes requested):
Looks like there is an issue with this change:

```
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while templating '{'name': 'RHVM-01', 'organization': 'Satellite', 'description': 'created by Ansible Playbook - for RHVM-01', 'variables': {'site_subnet': \"{  % if site == 'SITE-2' %}10.200.{% else %}10.100.{% endif%}\", 'remote_subnet': \"{  % if site == 'SITE-2' %}10.100.{% else %}10.200.{% endif%}\"}}'. Error was a <class 'ansible.errors.AnsibleError'>, original message: template error while templating string: Encountered unknown tag 'else'.. String: {  % if site == 'SITE-2' %}10.200.{% else %}10.100.{% endif%}"}
```

Tompage1994 (approved):

PR #179, opened 2021-05-07
Tower authentication Defaults

Tompage1994:
### What does this PR do?
At present when running from Tower with an Ansible Tower credential (which sets the environment vars for TOWER_HOSTNAME etc) this will fail unless we explicitly set the variables required. We should be able to not set them and let the env vars be picked up as per the defaults of the modules.

Attempt to run the roles without setting tower_hostname but with an env var of TOWER_HOSTNAME

N/A

N/A


sean-m-sullivan (approved):
LGTM, going to merge it.

PR #184, opened 2021-05-24
Removed update_secrets parameter from users role as it is no longer valid

suppaduppax:
### What does this PR do?
awx.tower_user  no longer contains the update_secrets parameter so the users role fails when run. This pull request removes that parameter from the task in the users role

### DIFF
--- a/roles/users/tasks/main.yml
+++ b/roles/users/tasks/main.yml
@@ -9,7 +9,7 @@
     first_name:          "{{ tower_user_accounts_item.firstname | default(tower_user_accounts_item.first_name | default(omit, true)) }}"
     last_name:           "{{ tower_user_accounts_item.lastname | default(tower_user_accounts_item.last_name | default(omit, true)) }}"
     is_superuser:        "{{ tower_user_accounts_item.is_superuser | default(tower_user_accounts_item.superuser | default('false')) }}"
-    update_secrets:      "{{ tower_user_accounts_item.update_secrets | default(omit) }}"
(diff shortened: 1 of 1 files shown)


Tompage1994 (changes requested):
It looks as though `update_secrets` is an option which has been added recently rather than removed.

Having a look at latest awx.awx I can see the option is there https://docs.ansible.com/ansible/latest/collections/awx/awx/tower_user_module.html and the same in the codebase for `devel`. 

I think there is a valid issue here, however, and that is to default the option to omit rather than `false` so that you wouldn't get an error if you are using an older version of the collection.

suppaduppax:
Made the requested change

suppaduppax:
On a second look, should default be set to 'true' ? The role readme says the default is 'true', but the default value is currently 'false' in the main branch

Tompage1994:
The default will be to omit. `default(omit, true)` doesn't do what you would think. The true is to ensure any false value causes the value to be omitted. In this case probably best to just use `default(omit)`. This will then mean if no value is given the awx collection (or the tower collection's) default value will be used. In this case the value is true. 

From awx collection:
```
update_secrets:
      description:
        - C(true) will always change password if user specifies password, even if API gives $encrypted$ for password.
        - C(false) will only set the password if other values change too.
      type: bool
      default: true
```

So I think the value we need here is `default(omit)`

Thanks!

suppaduppax:
Change made

Tompage1994 (approved):
Looks good!

PR #186, opened 2021-05-27
this should be true to prevent passwords leaking

djdanielsson:
### What does this PR do?
changes default from false to true, this will prevent password leaks from happening on accident if the user does not realize they need to set the tower_configuration_credentials_secure_logging to true.

Automated

no

none


sean-m-sullivan (approved):
LGTM, had not thought of this before, and false is fine for other modules, but agree for credentials.

PR #189, opened 2021-06-14
Fix boolean omits

### LINKED ISSUE #188: Boolean variables can't be defaulted with true as 2nd parameter
As described in https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#providing-default-values

> If you want to use the default value when variables evaluate to _false_ or an empty string you have to set the second parameter to true:

Meaning that e.g. `tower_validate_certs: false` will trigger the default in `{{ tower_validate_certs | default(omit, true) }}`, even though a value has been defined for the variable. There might be other places where this happens, but the issue became apparent based on this discussion.

Tompage1994:
### What does this PR do?
For all boolean values I have changed the default to `omit` rather than `omit, true` which means you can't set a value of `false`. I have additionally changed any values that we set a default for which was boolean to also omit to take the module's default.

Should pass CI

resolves #188 

N/A


sean-m-sullivan:
This goes back to breaking ENV vars being used
https://github.com/redhat-cop/tower_configuration/issues/132
I suggest we put in each roles defaults to looking up the ENV vars so we are in line with the awx collection
```
- name: Basic usage
  debug:
    msg: "'{{ lookup('env', 'HOME') }}' is the HOME environment variable."
```

Tompage1994:
> This goes back to breaking ENV vars being used
> #132
> I suggest we put in each roles defaults to looking up the ENV vars so we are in line with the awx collection
> 
> ```
> - name: Basic usage
>   debug:
>     msg: "'{{ lookup('env', 'HOME') }}' is the HOME environment variable."
> ```

I must say I feel uncomfortable defaulting options which are defaulted in the module. I'm also pretty certain that if the variable isn't set it will use the env var (unless you set `tower_validate_certs: ''` which doesn't make much sense.

I think its far worse that we can't set `tower_validate_certs: false` than catering for a strange setting of variables.

sean-m-sullivan:
I mean we are reverting to what the code was before when it was broke for env variables. #132 
Agree with the defaulting, but we need an override to make it work with both ways. Was debating on things, but I think double default comes into play. 
```
"{{ tower_validate_certs | default(lookup('env', 'TOWER_VERIFY_SSL') | default(omit)) }}"
```
Haven't had time to test that, but should default if neither is set, go to var if set, or env if not set.

ericzolf:
You need a `true` for the 2nd default, because lookup env returns always a value, be it the empty string.

This said, I looked through history, and I'm more with Tom, we tried to be nice with people but actually, the variables should not be defined and then we don't need all this hocus pocus.

sean-m-sullivan:
I'm doing a query to see about if they were thinking of dropping the legacy auth variables, but if you two are on board, I'm ok with dropping support, If anything, they can still use ENV for auth like they have, but have to use the global var for tower_ssl.

sean-m-sullivan:
@Tompage1994 if you can add a 
breaking_changes
to the changelog for breaking ssl when using env vars, we can merge this.

Tompage1994:
@sean-m-sullivan got rid of your commit. Should be good to go now

sean-m-sullivan (approved):
LGTM

Issue #181, opened 2021-05-11
Playbook example order


sean-m-sullivan:
Update playbook example order to 
settings, orgs teams, labels, cred types, creds, notification, project, inventories, inventory sources, job templates, workflows, roles, schedules,

right now a few are out of order and can cause issues on a blank tower, or updating certain things.

PR #191, opened 2021-06-21
Fix linting, workflows, and other various issues

### LINKED ISSUE #120: ansible-lint fails locally and possibly in the pipeline
I've been trying to run ansible-lint locally before I push my commits but it proved more difficult than I thought. The only way to fix the issue I could find and it's not really a solution was:

1. temporarily remove `ansible.cfg` so that it uses the standard collection directory (`ansible-galaxy collection install awx.awx` did also the trick and installed under `collections`)
2. I could get the check to work only by setting the `exclude_paths` list to _absolute_ paths in `.github/workflow-config/.ansiblelint.yml`:

```diff
-# exclude_paths:
-#  - roles/master_role_example/
+exclude_paths:
+- /home/myuser/[...development path...]/redhat_cop/tower_configuration/roles/master_role_example/
```
at the end calling then `ansible-lint -v --show-relpath -c .github/workflow-config/.ansiblelint.yml` worked.

I can live with the first point but perhaps we should document a bit somewhere how to test locally, and for the 2nd point, we could update the `.ansiblelint.yml` file to use the absolute path as in the pipeline, or is not an issue in the pipeline?

sean-m-sullivan:
### What does this PR do?
This fixes
#120 Linting failure in github action
#181 Playbook order for blank tower
And fixes bug which caused the workflow_job_template role not to run.

The automated tests should be updated and runnow.

### DIFF
--- a/roles/workflow_job_templates/tasks/add_workflows_schema.yml
+++ b/roles/workflow_job_templates/tasks/add_workflows_schema.yml
@@ -49,7 +49,7 @@
     validate_certs:                 "{{ tower_validate_certs | default(omit, true) }}"
   loop: "{{ workflow_loop_var.simplified_workflow_nodes }}"
   loop_control:
-    loop_var: workflow_loop_node"{{ tower_validate_certs | default(omit) }}"
+    loop_var: workflow_loop_node
   no_log: "{{ workflow_job_templates_secure_logging }}"
   # Execute only the nodes that define links to the following
   when: ((workflow_loop_node.always_nodes is defined and workflow_loop_node.always_nodes | length > 0 ) or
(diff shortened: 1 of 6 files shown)


roles/workflow_job_templates/tasks/add_workflows_schema.yml
```diff
@@ -49,7 +49,7 @@
     validate_certs:                 "{{ tower_validate_certs | default(omit, true) }}"
   loop: "{{ workflow_loop_var.simplified_workflow_nodes }}"
   loop_control:
-    loop_var: workflow_loop_node"{{ tower_validate_certs | default(omit) }}"
```
Tompage1994, line 52: This one's a bit worrying. Not sure how it got there... I can also see the validate_certs set to default(omit, true) above too so I'll raise a separate PR for that.

Tompage1994 (approved):
LGTM

PR #193, opened 2021-06-21
Update Readme and validate certs on workflow

### LINKED ISSUE #185: Using AWX Collection in playbook grabs older version of awx
Having the awx.awx collection as a parameter in the playbook uses an older version of that collection

```
- hosts: awx
  collections:
    - awx.awx
  tasks:
    - include_role:
        name: redhat_cop.tower_configuration.users
```

This was tested on a fresh Ubuntu Server vm when running the task:
```
    - include_role:
        name: redhat_cop.tower_configuration.users
```

which gives the error:
`Unsupported parameters for (tower_user) module: update_secrets`

When awx.awx collection is installed using: `ansible-galaxy collection install awx.awx`, the error disappears. 

This may be an Ansible issue but I am adding it as an issue just in case.

sean-m-sullivan:
### What does this PR do?
Updating Readme to be more explicit on requirements. 
Updating workflow role for the omit to match the rest of collection

Fixes #185

### DIFF
--- a/.github/workflow-config/.pre-commit-config.yml
+++ b/.github/workflow-config/.pre-commit-config.yml
@@ -1,19 +1,19 @@
 ---
 repos:
   - repo: 'https://github.com/pre-commit/pre-commit-hooks'
-    rev: v3.2.0
+    rev: v4.0.1
     hooks:
       - id: end-of-file-fixer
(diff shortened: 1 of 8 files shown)


Tompage1994 (approved):
LGTM

PR #196, opened 2021-06-23
Controller changeover

sean-m-sullivan:
### What does this PR do?
This is the update that will be the basis for controller_configuration 2.0
All roles written to be compatible with awx.awx 19.2.1
All role tests rewritten to follow the config/import and remove all json imports.
Update to missing options including instance groups, copy from, and a few others, each role checked for missing module options
All roles internal loops follow the __role_only_variable naming practice
Depreciated options removed from notifications_templates
Various other changes.

Automated tests


Tompage1994 (commented):
At a glance this looks good. I can't see anything wrong with it.

My only question is whether it would be a good idea to try to allow some level of backwards compatibility by adding the symlinks from the old module names and also doing some defaulting with the variable names? I don't mind if we don't want to go down that path though.

sean-m-sullivan:
@Tompage1994 I thought about the symlinks, but I don't think it works the same in terms of modules vs roles. We have a few options we COULD do: 
1. include backward compat where we invoke the old module names with additional roles
    a. use a var, or something to denote old ones, where it does a conversion between old role name that then invokes new role name? THis then runs the question on roles with same name as old but use new modules which is the majority. 
   b. have tower_roles that invoke old module name that is basically our Legacy branch

Right now the new awx.awx 19.2.2 and this works on tower 3.8, and I've tested that, however requires the variable name changeover from tower_ to controller_ , which we might be able to do something about, but everything with default isn't that great, but could be done on role invocation.

I'm not against including better backward compat, but not sure on how we choose when it comes to conflicts.

Tompage1994:
> @Tompage1994 I thought about the symlinks, but I don't think it works the same in terms of modules vs roles. We have a few options we COULD do:
> 
> 1. include backward compat where we invoke the old module names with additional roles
>    a. use a var, or something to denote old ones, where it does a conversion between old role name that then invokes new role name? THis then runs the question on roles with same name as old but use new modules which is the majority.
>    b. have tower_roles that invoke old module name that is basically our Legacy branch
> 
> Right now the new awx.awx 19.2.2 and this works on tower 3.8, and I've tested that, however requires the variable name changeover from tower_ to controller_ , which we might be able to do something about, but everything with default isn't that great, but could be done on role invocation.
> 
> I'm not against including better backward compat, but not sure on how we choose when it comes to conflicts.

Fair enough. Sounds like a complete pain so let's just say there is no backwards compatibility. (We can build some in later if we need to). I guess we will still have the `tower_configuration` collection on galaxy which points to v1.x anyway and the new controller_configuration would have its first release at 2.0.0 so the fact you literally need to specify a different collection is probably good enough for me. I think we should create a doc page for migrating and have that linked right at the top of the readme to make it as painless as possible though. Happy for that to be a separate PR though. I'm approving this now. Let me know if you want me to merge (or happy for you to do so)

Tompage1994 (approved):

PR #198, opened 2021-07-06
[cherry-pick] Adapt Workflows docs to new schema and fix schedules

ericzolf:
### What does this PR do?
- adapt workflow_job_templates' readme examples to new schema
- schedules defaults was missing an empty list
(cherry pick)

n/a

n/a

See PR #197 - note that it's a cherry pick towards the legacy branch (not towards devel)


sean-m-sullivan (approved):
LGTM thanks for updating, technically those fields can be left blank if there is only one, but better to have it in the examples.

PR #212, opened 2021-07-23
Parametrize playbook to run with CICD tools directly

### LINKED ISSUE #211: Make playbooks usable with CI/CD by default
The playbooks supplied with the collection have quite a bit of hardcode in them. With more parametrization or by using env vars they can be used directly with CI/CD tools such as GitLab CI.

With ansible 2.11 it's even [possible](https://docs.ansible.com/ansible/latest/user_guide/collections_using.html#using-a-playbook-from-a-collection) to run them straight from ansible-playbook command. With older versions - with full path to `~/.ansible/collections/...`

Either way this allows to use the collection as the single dependency for CI tools with no need for writing extra playbooks and storing them somewhere.

nesanton:
### What does this PR do?
Enriches the supplied playbook with parameters (env lookups) to allow running it more dynamically, e.g. from CI/CD tools

Fixes #211

### DIFF
--- a/playbooks/configure_controller.yml
+++ b/playbooks/configure_controller.yml
@@ -3,7 +3,7 @@
   hosts: localhost
   connection: local
   vars:
-    controller_validate_certs: "{{ lookup('env', 'CONTROLLER_VERIFY_SSL') | default('true') }}"
+    controller_validate_certs: "{{ lookup('env', 'CONTROLLER_VALIDATE_CERTS') | default('true') }}"
     controller_hostname: "{{ lookup('env', 'CONTROLLER_HOSTNAME') | default(omit) }}"
     controller_username: "{{ lookup('env', 'CONTROLLER_USERNAME') | default(omit) }}"
     controller_password: "{{ lookup('env', 'CONTROLLER_PASSWORD') | default(omit) }}"
--- a/playbooks/configure_controller.yml
+++ b/playbooks/configure_controller.yml
@@ -3,49 +3,44 @@
 #
 # Set the following variables on host_vars/localhost or group_vars/all
 #
-# - awx_validate_certs:
+# - controller_validate_certs:
 #     description: |
 #                  Wether to trust self-signed or invalid certificates
-#                  Falls back to env AWX_VALIDATE_CERTS. If nothing
-#                  provided defaults to true.
+#                  Falls back to env CONTROLLER_VERIFY_SSL and then to awx-cli config.
 #
-# - awx_hostname:
+# - controller_hostname: |
 #     description: |
 #                  Hostname of AWX or Red Hat Ansible Automation Platform Controller.
-#                  Falls back to env AWX_HOSTNAME. Omitted otherwise.
+#                  Falls back to env CONTROLLER_HOST and then to awx-cli config.
 #
-# - awx_username:
+# - controller_username:
 #     description: |
 #                  Username for AWX or Red Hat Ansible Automation Platform Controller.
-#                  Falls back to env AWX_USERNAME. Omitted otherwise.
+#                  Falls back to env CONTROLLER_USERNAME and then to awx-cli config.
 #
-# - awx_password:
+# - controller_password:
 #     description: |
 #                  Password for AWX or Red Hat Ansible Automation Platform Controller.
-#                  Falls back to env AWX_PASSWORD. Omitted otherwise.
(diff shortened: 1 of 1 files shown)


Tompage1994:
So the playbooks that are currently in this collection are not really intended for anything more than use in *our* CI for this project. We did previously discuss moving those to somewhere else and then providing playbooks which can be consumed. 

I believe this change here will break our CI as there is a step to copy across the current playbook to a different directory and point the workflow to use that. We can then remove any clutter from this playbook to more freely define a generic playbook for use in CI.

sean-m-sullivan:
@Tompage1994 I think the idea that we have working playbooks for users in the playbooks directory is where we should move, I believe latest ansible you can refer to them. That being said your right our Testing needs to be kept, so we should move it into a test folder. I think its a question of what do we want that to look like.

playbooks/configure_controller.yml
```diff
@@ -3,34 +3,37 @@
   hosts: localhost
   connection: local
   vars:
-    controller_validate_certs: false
+    controller_validate_certs: "{{ lookup('env', 'CONTROLLER_VALIDATE_CERTS') | default('true') }}"
```
sean-m-sullivan: So I've been thinking about this, and put in a PR to move our testing playbooks to the .github directory, but then it comes down to how we want to populate the playbooks.
Should it be a var file that gets loaded, that gives the user the option to lookup env, or however they want to load vars, I do like the idea of parameterized path, but that can be an extra var loaded in upon exection from lookup as well.

nesanton: Thanks guys for looking at this. I did suspect that changing this exact playbook would not work for you. Yet, funny enough, for most use cases I think it would be quite a perfect playbook. With these changes it works for me. I think it deserves to be in the collection for user consumption. Even with ansible older than 2.11 users can still call it from `~/.ansible/blah/`

As for how to parameterize it, falling back to `lookup_env` is very handy when doing CICD (at least with gitlab, but I'm sure it's similar with github actions etc.). CICD vars + installed collection and you're good to go.

I'd happily jump on a call and show what I mean if it's useful.

playbooks/configure_controller.yml
```diff
@@ -1,36 +1,78 @@
 ---
+### Variables
+#
+# Set the following variables on host_vars/localhost or group_vars/all
+#
+# - awx_validate_certs:
```
nesanton: Would this be a good idea to describe vars this way?
I had to deviate from controller_* to awx_* to avoid templating recursion. Is there a better way?

nesanton: Ah, wait, the controller_hostname, _validate_certs and credentials fall back to env vars and are omitted by default via collection. No need to do it again. Reverting to controller_* naming

playbooks/configure_controller.yml
```diff
@@ -1,36 +1,73 @@
 ---
+### Variables
+#
+# Set the following variables on host_vars/localhost or group_vars/all
+#
+# - controller_validate_certs:
+#     description: |
+#                  Wether to trust self-signed or invalid certificates
+#                  Falls back to env CONTROLLER_VERIFY_SSL and then to awx-cli config.
```
nesanton, line 9: `_validate_certs` <> `_verify_ssl` as well as `_host` <> `_hostname` are misleading. Perhaps they need to be homogenized either here or in the awx.awx collection

playbooks/configure_controller.yml
```diff
@@ -1,36 +1,72 @@
...
+#
+
+
 - name: Playbook to configure ansible Controller post installation
   hosts: localhost
   connection: local
-  vars:
-    controller_validate_certs: false
   collections:
     - awx.awx
```
nesanton, line 45: Unfortunately you can't template in here.  I think the only way to adjust it for ansible.controller would be to sed it:

`sed -i 's/- awx.awx/- ansible.controller/' playbooks/configure_controller.yml`

sean-m-sullivan: We can make a lot of changes now that we aren't using this to run CI/CD, and just rely on the user, in the end we should make changes to make it work for everyone, or have them supply their own vars.

sean-m-sullivan (approved):
I've been thinking a lot about this, and we made changes to not use this file for CI/CD tests for the collection, and move those into the .github folder, This should be for playbooks for users to use as you suggest. Sorry for taking so long to get back around to this, but approving it, and we can make further changes. I am thinking even a readme in the folder in addition to your top comments.

PR #217, opened 2021-08-11
Rename scm_credential to credential

nesanton:
In the awx.awx collection it's called "credential".
Calling it "scm_credential" here causes confusion and unexpected results
when users rightfully try to use "credential", and it gets silently ignored.


nesanton:
how wrong I was ;)

PR #219, opened 2021-08-23
updated meta files for rebranding from tower to controller

bradynathan:
### What does this PR do?
Allows for use with the supported content on cloud.redhat.com

Remove the legacy ansible.tower collect and the awx.awx collect.  Then install the ansible.controller collection and verify execution.

none

none

### DIFF
--- a/roles/ad_hoc_command/meta/main.yml
+++ b/roles/ad_hoc_command/meta/main.yml
@@ -37,6 +37,8 @@ galaxy_info:
     - "commands"
 
 collections:
+  - ansible.controller
+  - ansible.tower
   - awx.awx
 
 dependencies: []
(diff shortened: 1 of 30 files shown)


Tompage1994 (approved):
Thanks for the contribution this looks good.

@sean-m-sullivan do you reckon we should just pull out the `ansible.tower` collection in here entirely given this is for targeting 4.x onwards and the expectation will presumably be to use ansible.controller or awx.awx with it (obviously only for the v2 of this repo)

sean-m-sullivan:
So if these roles were standalone, the meta file would make a difference, as it would dictate what ansible galaxy would pull when installing a stand alone role, but because its a collection, it doesn't care, With the tests passing, and this being the Controller configuration. ansible.tower needs to be ansible.controller, for this update, Otherwise I see no problem with making these changes. Going to add a specific note to that effect.

sean-m-sullivan:
2nd thought after reading through it, going to make a seperate issue to remove ansible.tower from the meta since it was in before and left

sean-m-sullivan (approved):

PR #226, opened 2021-08-31
Ansiblelint ignore changelogs dir

Tompage1994:
### What does this PR do?
Fixes an issue with ansible lint where it was trying to parse the changelogs dir

CI should pass

N/A

N/A


sean-m-sullivan (approved):
Sure, understand about ansiblelint, but for a second thought it was yamllint as well, that we should still do.

Issue #228, opened 2021-09-03
use case of update role, example: project_update,inventory_source_update


dhikrahashim:
When we are going to remove any projects or inventory, if we enabled role like inventory_source_update and project_update and if it uses config from controller_projects and controller_inventory_sources

We are going to get  "couldn't find mentioned project"

"ansible_loop_var": "__project_update_update_item", "changed": false, "msg": "Unable to find project"

Because I mentioned the project state to absent and at the first role projects the play is removed the project and when project update role comes we are getting error

```
pre_tasks:
    - name: Include vars from configs directory
      include_vars:
        dir: ./configs
        ignore_files: [controller_config.yml.template]
        extensions: ["yml"]
  roles:
    - {role: projects, when: controller_projects is defined, tags: projects}
    - {role: inventories, when: controller_inventories is defined, tags: inventories}
    - {role: project_update, when: controller_projects is defined, tags: projects}
    - {role: inventory_sources, when: controller_inventory_sources is defined, tags: inventory_sources}
    - {role: inventory_source_update, when: controller_inventory_sources is defined, tags: inventory_sources}
    - {role: job_templates, when: controller_templates is defined, tags: job_templates}
```

@Tompage1994 @sean-m-sullivan @kedark3

sean-m-sullivan:
Sounds like we need a check of state before invoking the role, thanks for pointing it out

sean-m-sullivan:
@dhikrahashim  added PR to fix the issue, please review :)

PR #229, opened 2021-09-11
Update fix

sean-m-sullivan:
### What does this PR do?
- added when to skip project update when item is absent
- added when to skip inventory source update when item is absent
- added default to organization on project as it is not required for the module, but it is highly recommended.
- added tests for the above skips

Automated tests have been updated to reflect this change and test for them.

#228


Tompage1994 (approved):
Looks good :)

PR #231, opened 2021-09-12
playbok update

sean-m-sullivan:
### What does this PR do?
Update playbooks for users to use and move the testing playbooks to .github

Automated tests are set


sean-m-sullivan:
I'm not fixing the meta file as its fixed in the update fix for legacy, so we don't have a merge conflict.

Tompage1994 (changes requested):
Looks like you've got a linting error (probably in your changelog only).

Otherwise if this is the same change as made in v2 then this is all good

sean-m-sullivan:
Followed the instructions and it merged on me, This is the same change as in v2, just brings the playbook functionality to the main and move the config files/testing location.

PR #234, opened 2021-09-29
Example update

sean-m-sullivan:
### What does this PR do?
By Popular demand. Moving the testing files to the examples folder. People didn't like that we moved them out of the playbooks dir and hid them, but those are now functional playbooks for use, so moving them back to a top level dir, crossing fingers we never move them again.


Tompage1994 (approved):

Tompage1994 (approved):
LGTM

PR #240, opened 2021-10-29
the correct var is controller_templates not templates

djdanielsson:
### What does this PR do?
the correct var is controller_templates not templates

Automated

none

none


sean-m-sullivan (approved):
Good Catch, LGTM!

PR #241, opened 2021-11-03
Auto remove

### LINKED ISSUE #239: feature, remove old content
Some people may want the ability to remove old content that is no longer in their code, I have an draft of how we could do this if this feature idea is accepted. How it would work is pulling the current list of x (project, job_templates etc) and compare that list vs the list in code and make a delete call before updating/creating any new content when a specific variable is set to true that they want to remove old items.

djdanielsson:
### What does this PR do?
gives the option to remove content that is no longer in code that is on controller

each section needs to be tested to prove it will delete just the content not in code without errors

resolves #239

I have had issues with my controller currently to test any of the code, I hope to get it in a working state soon so I can test it myself

### DIFF
--- a/roles/credential_types/tasks/main.yml
+++ b/roles/credential_types/tasks/main.yml
@@ -27,7 +27,7 @@
       ansible.builtin.set_fact:
         controller_credential_types: "{{ (controller_credential_types | default([])) + [{'name': item, 'state': 'absent'}] }}"
       loop: "{{ credential_types_difference }}"
-  when: credential_types_auto_remove_old
+  when: credential_types_auto_remove_old or auto_remove_all_old
 
 - name: Add Credential Types
   credential_type:
--- a/roles/credential_types/tasks/main.yml
+++ b/roles/credential_types/tasks/main.yml
@@ -1,4 +1,49 @@
 ---
+- name: Block for removing old content
+  ansible.builtin.block:
+    - name: Pull Current Credential Types
+      ansible.builtin.uri:
+        url: "{{ controller_hostname | mandatory }}/api/v2/credential_types/"
+        method: GET
+        return_content: true
+        body_format: json
+        status_code: [200, 202]
+        force_basic_auth: true
+        user: "{{ controller_username | default(omit, true) }}"
+        password: "{{ controller_password | default(omit, true) }}"
+        validate_certs: "{{ controller_validate_certs | default(omit) }}"
+      register: r_current_credential_types
+
+    - name: Print pulled output
+      ansible.builtin.debug:
+        var: r_current_credential_types
+        verbosity: 2
+
+    - name: Create lists
+      ansible.builtin.set_fact:
+        cred_type_in_controller: "{{ cred_type_in_controller | default([]) + [item.name] }}"
+        cred_type_in_code: "{{ _cred_type_in_code }}"
+      loop: "{{ r_current_credential_types.json.results }}"
+      vars:
+        _cred_type_in_code: "{{ controller_templates | json_query(_query2) }}"
+        _query2: "[].name"
+
+    - name: Print lists
+      ansible.builtin.debug:
+        msg: "cred_type_in_controller: {{ cred_type_in_controller }}, cred_type_in_code: {{ cred_type_in_code }}"
+        verbosity: 2
+
+    - name: Create list of items to remove
+      ansible.builtin.set_fact:
+        credential_types_to_delete: "{{ cred_type_in_code | difference(cred_type_in_controller) }}"
+
+    - name: Delete old Credential Types
+      ansible.controller.job_template:
+        name: "{{ item }}"
+        status: "absent"
+      loop: "{{ credential_types_to_delete }}"
+  when: credential_types_auto_remove_old
+
 - name: Add Credential Types
   credential_type:
     name:                           "{{ __controller_credential_type_item.name | mandatory }}"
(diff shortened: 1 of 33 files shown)


sean-m-sullivan:
I like it, and looks good, I'm sure there will be some tweaks after we play with it, but I think its a great addition.

roles/credential_types/tasks/main.yml
```diff
@@ -1,4 +1,34 @@
 ---
+- name: Block for removing old content
+  block:
+    - name: Pull Current credential_types
+      uri:
```
Tompage1994, line 5: I do dislike the uri module being used here.

I think we should be able to neatly pull it out using the following code snippet:

```yaml
- name: Load the credential_types
  set_fact:
    _credential_types_in_controller: "{{ lookup('awx.awx.controller_api', 'credential_types') }}"
```

roles/inventories/tasks/main.yml
```diff
@@ -1,4 +1,34 @@
...
+        password: "{{ controller_password | default(omit, true) }}"
+        validate_certs: "{{ controller_validate_certs | default(omit) }}"
+      register: r_current_inventories
+
+    - name: Get differences
+      set_fact:
+        inventories_difference: "{{ _inventories_in_controller | difference(_inventories_in_code) }}"
+      vars:
+        _inventories_in_controller: "{{ r_current_inventories.json.results | json_query(_query2) }}"
+        _query2: "[].name"
```
Tompage1994, line 22: We may need to be careful here as this would break if we have multiple orgs and items with the same name. We need to do some level of matching on org name as well (for relevant endpoints)

roles/credential_types/tasks/main.yml
```diff
@@ -1,4 +1,34 @@
...
+        user: "{{ controller_username | default(omit, true) }}"
+        password: "{{ controller_password | default(omit, true) }}"
+        validate_certs: "{{ controller_validate_certs | default(omit) }}"
+      register: r_current_credential_types
+
+    - name: Get differences
+      set_fact:
+        credential_types_difference: "{{ _credential_types_in_controller | difference(_credential_types_in_code) }}"
+      vars:
+        _credential_types_in_controller: "{{ r_current_credential_types.json.results | json_query(_query2) }}"
```
Tompage1994, line 21: We may want to avoid `json_query` as that isn't in any RH Certified collections. If just doing a query for `[].name` you can replace with `map(attribute='name')`

Tompage1994 (commented):
This is looking good to me too.

I wonder whether it would be better to move that block out into a separate tasks file so the main.yml is easier to read. (I know that I find it easiest to find the correct options required by going to the code so having this fluff around it will make that harder)

roles/inventories/tasks/main.yml
djdanielsson: would it? i thought it should only compare the names.... ill have to test that to confirm

roles/credential_types/tasks/main.yml
djdanielsson: good point... as a side note we might want to try and get that changed because it is super powerful

roles/credential_types/tasks/main.yml
djdanielsson: oh interesting, i haven't ever used that lookup before

roles/credential_types/tasks/main.yml
Tompage1994: I believe it is something to do with the dependency on jmespath and that's why it isn't being moved into AH

roles/inventories/tasks/main.yml
Tompage1994: Yeah, you can have most objects with the same name but in different orgs. Because of that the matching wouldn't always work here (you may get false positives/negatives)

roles/credential_types/tasks/main.yml
sean-m-sullivan: Going to leave this code snippet here, its a bit more encompassing, and may have more use here
```
"{{ lookup('awx.awx.tower_api', 'workflow_job_templates', query_params={ 'name': workflow_template.name } ,host=tower_hostname, oauth_token=tower_oauthtoken, verify_ssl=false }}"
```
but agree on using the org on lookup as well.

roles/inventories/tasks/main.yml
sean-m-sullivan: Always assume that Name + org is needed, its a cardinal rule of awx club.

roles/credential_types/tasks/main.yml
sean-m-sullivan: Exactly, We are trying to be a close to supportable, and while i LOVE json_query, and use it at the client, we have to be careful about that. 

That being said, If that becomes unusable to pair them down, and doing a compare, I would not be against making a module to do it.

djdanielsson: just to note in my testing I found it looks like you use json_query in the workflow_job_template_node so it would already be a requirement, but also to be fair i don't need to use it so I will swap it over to attribute

roles/inventories/tasks/main.yml
djdanielsson: I spent some time on it and without json_query it's super hard to do when adding in checking the org, so I think it needs to be written as a module/plugin

djdanielsson:
see PR #242 with the better answer

sean-m-sullivan:
#242 would still need the state absent, and the block to do that Before you run the regular tasks, so it doesn't completely solve it, Was just trying to solve the difference problem.

PR #242, opened 2021-11-07
Auto Remove Diff Plugin

sean-m-sullivan:
### What does this PR do?
Adds a lookup plugin to aid in the process of auto remove. 
lookup plugin finds diff between two lists one from the API and another user provided. 

Automated tests have been updated to account for python.

#239


djdanielsson:
looks good to me, haven't had time to test it yet.

Tompage1994:
Not sure if I agree with how this plugin is structured. Should we not just take in a list of the objects and the endpoint and return the diff? Seems annoying to be required to first do your own lookup when we could easily do that as part of this?

I think I'd like to see something like this quickly thought out example be possible.

e.g.
```
- set_fact:
    org_diff: "{{ lookup('redhat_cop.controller_configuration.controller_object_diff', 'organizations', controller_organizations }}"

- include_role:
    name: redhat_cop.controller_configuration.organizations:
  vars:
    controller_organizations: "{{ org_diff }}"
    controller_state: absent

- include_role:
    name: redhat_cop.controller_configuration.organizations:
```

I'd also like to see some documentation. (possibly as an example in the plugin) of a more end to end solution like above.

sean-m-sullivan:
So at least my idea was that this would be combined with @djdanielsson 's PR
https://github.com/redhat-cop/tower_configuration/pull/241
But after doing some testing on Merging with awx.awx, I think I have a way to incorporate it easily in with others. as all you need to do is invoke 
```
- include_role:
    name: redhat_cop.controller_configuration.organizations:
  vars:
    controller_organizations: "{{ org_diff }}"
    controller_state: absent
```
Like you said.

djdanielsson:
another idea would be to insert the org_diff objects into the already created controller_organizations list with the state absent so the org role only needs to run once

Tompage1994:
Yeah I was thinking that but couldn't think of a good syntax to add `state: absent` to each of the objects. (Perhaps a map but I dont know if there is anything that will do that?) and then append the two lists together

sean-m-sullivan:
@Tompage1994 Easiest way is an option IN this plugin, I'll tinker with that :). then can work a way to merge the two lists.

djdanielsson:
yes it should be pretty easy to do in the plugin itself and have it return the full list of objects both to add/update and remove.

Tompage1994:
Maybe also an option to include the initial list too so you can literally use the following:

```
roles:
  - name: organizations
    vars:
      controller_organizations: "{{ lookup('redhat_cop.controller_configuration.controller_object_diff', 'organizations', controller_organizations, set_absent=true, with_present=true }}"
```

I think something like that would be pretty cool

djdanielsson:
> ```
> roles:
>   - name: organizations
>     vars:
>       controller_organizations: "{{ lookup('redhat_cop.controller_configuration.controller_object_diff', 'organizations', controller_organizations, set_absent=true, with_present=true }}"
> ```
What would it do if both options were false? because it seems like it wouldn't do anything, so I think we should just ask with_present true or false and if false it would just return the list of items to remove.

Tompage1994:
> > ```
> > roles:
> >   - name: organizations
> >     vars:
> >       controller_organizations: "{{ lookup('redhat_cop.controller_configuration.controller_object_diff', 'organizations', controller_organizations, set_absent=true, with_present=true }}"
> > ```
> 
> What would it do if both options were false? because it seems like it wouldn't do anything, so I think we should just ask with_present true or false and if false it would just return the list of items to remove.

My thoughts are that `set_absent` would add `state: absent` to the data structure returned and `with_present` would also return the list passed into the plugin. Though not sure it would make sense if `set_absent` was false and `with_present` was true.

sean-m-sullivan:
My thoughts are that 'set_absent' would add state: absent to the data structure returned and 'with_present' would only use those already set to present to compare. 

I think we agreed on the 'set_absent', but the 'with_present' what are we looking for exactly??? 
1. That we return all items with the absent set only on those diffed? I'm on the fence about this one, but could see where we merge the two lists somehow to get add absent to the ones that we Identify. 
2. Only use those with state absent or no state set?
3. Returning one of the lists? they are already set as vars, not sure what the return would do.

djdanielsson:
it would be 1. the idea is if you could return everything that you want to add and remove into one list then you could just call the role to do that once which would be more efficient. I think it should be pretty easy for the plugin to do once you have the diff list and added the state absent to each one of those objects then you could just combine that list with the given list that it compared with to have a full list of everything you want to add and remove.

sean-m-sullivan:
@djdanielsson If you could review this, I think I've updated it to work like you said, and agree that it makes more sense once I returned to it.

I've updated the lookup plugin, and if run in default, and prior to invoking roles, it should remove any unwanted objects. 
I've added tests for all the objects I think that should be used with, if we want more, we can look at that as well. Right now the had code is to look at the summary fields in the api for the org name. 

The testing went through a few iterations, but I like the solution presented. Its looping through a file, and using a file to describe each item. Changes were made to at least have two items of every object in the default org.

examples/tasks/differential.yml
```diff
@@ -0,0 +1,14 @@
...
+
+- name: "Find the difference of {{ differential_item.name }} between what is on the Controller versus curated list."
+  set_fact:
+    set_absent_diff: "{{ lookup('controller_object_diff', api_list=controller_api_results, compare_list=differential_item.differential_test_items, with_present=differential_item.with_present ) }}"
+
+- name: "Assert that the expected {{ differential_item.name }} results match."
+  assert:
+    that:
+      - set_absent_diff ==  differential_item.expected_test_result
+...
```
Tompage1994, line 14: Could you add a test case for where you have set_absent and with_present set as true and compare to 

```
differential_item.expected_test_result + differential_item.differential_test_items
```

plugins/lookup/controller_object_diff.py
```diff
@@ -0,0 +1,112 @@
...
+  set_fact:
+    project_diff: "{{ lookup('redhat_cop.controller_configuration.controller_object_diff', api_list=controller_projects_api,
+    compare_list=controller_projects) }}"
+
+"""
+
+RETURN = """
+_raw:
+  description:
+    - Name and Organzization of objects that are NOT in the list
```
Tompage1994: I suppose this is not necessarily true if you set `with_present` so you may want to add a caveat for that

plugins/lookup/controller_object_diff.py
```diff
@@ -0,0 +1,112 @@
...
+    type: boolean
+    default: True
+  with_present:
+    description:
+      - Include items in the original compare list in the output, and set state: present
+    type: boolean
+    default: True
+"""
+
+EXAMPLES = """
```
Tompage1994, line 41: Could you add an example with the set_absent and with_present set. Maybe also add using it with another role to show a fuller usage

Tompage1994 (changes requested):
A couple of small things but otherwise looking good

plugins/lookup/controller_object_diff.py
sean-m-sullivan: Getting both the return and examples updated, as things have changed quite a bit since I wrote that.

djdanielsson:
looks pretty good to me

examples/tasks/differential.yml
sean-m-sullivan: The 2nd test item Projects is specifically set to have both set_absent and with_present to be true, and the expected test has the relevant items already added. 
Normal default tests are with_present as false, set_absent to be True.

examples/configs/differential_items.yml
```diff
@@ -0,0 +1,137 @@
...
+      - name: controller_application-app3
+        organization: Default
+        authorization_grant_type: password
+        client_type: confidential
+        skip_authorization: true
+    expected_test_result:
+      - name: controller_application-app2
+        organization: Default
+        state: absent
+  - name: projects
```
sean-m-sullivan, line 15: @Tompage1994 This is the example with set_absent and with_present to be true

plugins/lookup/controller_object_diff.py
sean-m-sullivan: @Tompage1994 A better more comprehensive example with both options has been set.

sean-m-sullivan:
@Tompage1994 should be ready for review

Tompage1994 (approved):
Looks good!

PR #243, opened 2021-11-10
Default cloud for kind and remove 'raw'

Tompage1994:
### What does this PR do?
Fixes the documentation for credential types and also adds 'cloud' as the default kind as that is what happens in the UI

CI will run

N/A

N/A


sean-m-sullivan:
Going to need to test this, as last I checked the raw was needed for the input to come out correct.

djdanielsson:
yes I recently tried without raw and it doesn't work

Tompage1994:
With `"{% raw %}{  { myvar }}{% endraw %}"` the result in Tower is `"{  { myvar }}"` whereas I've found without the raw it works perfectly as required.

i.e. just setting `"{  { myvar }}"` in the variables will result in `"{{ myvar }}" in Tower.

sean-m-sullivan:
I think I see what your saying, the double space stays in, but the raw isn't required, I think this stems back to when I was creating it, that I was playing around with using raw, and likely never tried without, still I want to do a test on my end, JUST to make sure, as it has the potential to break things.

sean-m-sullivan (approved):
Looks good, was able to test, and it works as advertised, would never have 2nd guessed the raw part

PR #244, opened 2021-11-18
Adding instance_groups variable definition to the default for The job_templates role

zed-shamim:
### What does this PR do?
The variable "instance_groups" definition was missing from the default/main.yml for the job_template role for the rhel_cop controller_configuration collection. 

There is no test needed as its a comment for the variable that is present inside the tasks for the job_template role but missing from the default section. 

No open issue


sean-m-sullivan (approved):
Nice Catch! LGTM

PR #246, opened 2021-12-10
remove json_query dependency from collection

Tompage1994:
### What does this PR do?
Gets rid of dependency for json_query which is only available in community.general which is difficult for some people to get access to.

CI pipeline should run

N/A

N/A

### DIFF
--- a/changelogs/fragments/remove_json_query.yml
+++ b/changelogs/fragments/remove_json_query.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - Removes json_query which is not in a RH Certified collection so does not receive support and replaced with native ansible filters
+...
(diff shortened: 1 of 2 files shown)


sean-m-sullivan:
it looks like a test errors, I also want to run this inside of client and do testing as well as its a big, but welcome change. Will do some testing this weekend and on Monday.

Tompage1994:
@sean-m-sullivan Thanks for pushing your fix to this. As you may have been able to tell this was a dump and run last thing on a Friday after quickly typing out what I and tested in a customer environment and never tested it locally (I was hoping CI/CD would catch as it did)

sean-m-sullivan (approved):
Looks good, haven't run into any problems with testing. Good to remove this dependency!

PR #247, opened 2021-12-17
Add error message for diff plugin

Tompage1994:
### What does this PR do?
Slight addendum to #242 (which I spotted straight after merging) that if you don't give the organization on `compare_list` and you don't give `summary_field` on `api_list` then it fails (correctly) but doesn't give a helpful error message. This does a check on the field then errors stating that the field is missing from the input.

CI should still pass
Manually running the below will fail, but with a useful error:

```
  - debug:
        msg: "{{ lookup('redhat_cop.controller_configuration.controller_object_diff', api_list=current_proj, compare_list=exp_proj) }}"
      vars:
        current_proj:
          - name: proj1
        exp_proj:
          - name: proj1
            organization: org1
```

N/A

Addendum to #242

### DIFF
--- a/plugins/lookup/controller_object_diff.py
+++ b/plugins/lookup/controller_object_diff.py
@@ -100,6 +100,16 @@ class LookupModule(LookupBase):
         keys_to_keep = ["name", "organization"]
         api_keys_to_keep = ["name", "summary_fields"]
 
+        for item in compare_list:
+            for key in keys_to_keep:
+                if key not in item.keys():
+                    self.handle_error(msg="Key: '{0}' missing from item in compare_list".format(key))
+
+        for item in api_list:
+            for key in api_keys_to_keep:
+                if key not in item.keys():
+                    self.handle_error(msg="Key: '{0}' missing from item in api_list. Does this object come from the api?".format(key))
+
         # Reduce list to name and organization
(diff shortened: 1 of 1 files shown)


plugins/lookup/controller_object_diff.py
```diff
@@ -100,6 +100,16 @@ def run(self, terms, variables=None, **kwargs):
...
 
+        for item in compare_list:
+            for key in keys_to_keep:
+                if key not in item.keys():
+                    self.handle_error(msg="Key: '{0}' missing from item in compare_list".format(key))
+
+        for item in api_list:
+            for key in api_keys_to_keep:
+                if key not in item.keys():
+                    self.handle_error(msg="Key: '{0}' missing from item in api_list. Does this object come from the api?".format(key))
```
sean-m-sullivan: Item should be returned as well, so the user can identify what item had the issue.

sean-m-sullivan (changes requested):
Added comment about adding item to error message, otherwise this looks great.

sean-m-sullivan:
I noticed this to yesterday when I started implementing diff outside of tests, found out 3 of my job templates didn't have orgs, out of 101. Was going to get to it this weekend :). 

Awesome that you beat me to it :+1: 
Added one change to make, otherwise looks perfect.

sean-m-sullivan (approved):
LGTM

PR #248, opened 2021-12-17
Extend license role to support pulling from a URL or b64 content

### LINKED ISSUE #214: Extend license role to support pulling the manifest from a web server or as base64 variable
Not always can the manifest be present locally to the execution environment.

I recently extended the corresponding role in AgnosticD to also allow to pull it from a web-server (with or without user/password) or simply Base64 encoded as variable (rather ugly but good enough for a PoC without local storage): https://github.com/redhat-cop/agnosticd/tree/development/ansible/roles/tower-license-injector_cfg

I can merge the code myself but as it might take some time, I thought, someone might be faster...

Tompage1994:
### What does this PR do?
Allows user to set one of `manifest_file`, `manifest_url` or `manifest_content` where this is the manifest file location, a URL of the manifest file, or the content of a manifest (encoded by base64)

The only real changes here are that stuff happens to the file prior to being placed onto Tower

resolves #214

### DIFF
--- a/changelogs/fragments/license.yml
+++ b/changelogs/fragments/license.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - Added new options for adding manifest to Ansible Controller inc. from a URL and from b64 encoded content
+...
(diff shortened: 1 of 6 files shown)


sean-m-sullivan (approved):
LGTM

PR #261, opened 2022-01-18
fix incorrect github link

Tompage1994:
### What does this PR do?
Small fix for the readme links which were pointing to the wrong github URL (controller_configuration.git)

e.g. note that the status icons on the readme are currently broken

N/A

N/A

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Red Hat Communties of Practice Controller Configuration Collection
 
-![Ansible Lint](https://github.com/redhat-cop/controller_configuration/workflows/Ansible%20Lint/badge.svg)
-![Galaxy Release](https://github.com/redhat-cop/controller_configuration/workflows/galaxy-release/badge.svg)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -26,16 +26,16 @@ Click the `Content` button to see the list of content included in this collectio
 
 ## Installing this collection
 
-You can install the redhat_cop controller_configuration collection with the Ansible Galaxy CLI:
+You can install the redhat_cop tower_configuration collection with the Ansible Galaxy CLI:
```
sean-m-sullivan: The links should change, the text should not, we should keep calling it controller configuraiton, but the web links should be tower configuration.

sean-m-sullivan (changes requested):
Need to change back in several places, otherwise agree.

README.md
Tompage1994: My bad. Just did a dumb find and replace without thinking

sean-m-sullivan (approved):
LGTM

PR #264, opened 2022-01-24
changing default no_log to true for user creation

djdanielsson:
### What does this PR do?
changes default no log to true in users role, this task deals with passwords

Automated should be fine

no

none

### DIFF
--- a/roles/users/defaults/main.yml
+++ b/roles/users/defaults/main.yml
@@ -26,5 +26,5 @@ controller_user_accounts: []
 # set this variable to something false and there won't be a default password
 controller_user_default_password: "change_me"
 
-controller_configuration_users_secure_logging: "{{controller_configuration_secure_logging | default('false')}}"
+controller_configuration_users_secure_logging: "{{controller_configuration_secure_logging | default('true')}}"
 ...


ericzolf:
Without having looked in the details, is this really required: it obfuscates the whole task where I would expect that the module already declares the password parameter as being no_log. If not, it should be changed in the module, not in the role IMHO.

djdanielsson:
The actual module does have it no_log [user.py](https://github.com/ansible/awx/blob/8d4d1d594ba1116db94a3303054362a35239758f/awx_collection/plugins/modules/user.py#L127) but because we loop it prints all the items and will show the password. To my knowledge, the only way to prevent password leaks is to hide the whole output.

sean-m-sullivan:
@djdanielsson This is a good catch, and yes it should default to no log, just like we do with credentials, however, the place to change it is in the defaults variable. here
https://github.com/redhat-cop/tower_configuration/blob/9ab8f4dc585b031e63c773222b232ee5c325cfba/roles/users/defaults/main.yml#L29
Like it is done here
https://github.com/redhat-cop/tower_configuration/blob/9ab8f4dc585b031e63c773222b232ee5c325cfba/roles/credentials/defaults/main.yml#L27

djdanielsson:
> however, the place to change it is in the defaults variable. here
> 
> https://github.com/redhat-cop/tower_configuration/blob/9ab8f4dc585b031e63c773222b232ee5c325cfba/roles/users/defaults/main.yml#L29

isn't that where I made the change? or am I missing something...

sean-m-sullivan:
Please disregard my previous comment, I was in the wrong location, this is changing the correct location.

sean-m-sullivan (approved):
LGTM

PR #265, opened 2022-02-02
https and validate certs added to example

ravenawk:
### What does this PR do?
Makes small changes to "examples/configuration_controller.yml": 
  - adding https in the url option 
  - adding the verify_certificates option.

Before those changes were made the playbook would just timeout waiting for a 200 return code.

ansible-lint and yamllint did not show any errors when run against the playbook

### DIFF
--- a/examples/configure_controller.yml
+++ b/examples/configure_controller.yml
@@ -15,8 +15,9 @@
 
     - name: Wait for Controller to come up
       uri:
-        url: "{{ controller_hostname }}/api/v2/ping"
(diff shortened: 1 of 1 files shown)


Tompage1994 (commented):
Hi @ravenawk thanks for the contribution.

Unfortunately this change breaks our Ci because it assumes controller_host contains the `https` in it. I'm going to close this PR because of this reason, but if you think our documentation is missing this then it would be good if you raised another PR there.

PR #267, opened 2022-02-11
Lookup doc fix

sean-m-sullivan:
### What does this PR do?
Added query example to plugin documentation. Ran into issue where the lookup returned a dict instead of a list because it was a singular item, causing plugin issue due to type error.


djdanielsson (approved):

djdanielsson:
looks good to me

PR #268, opened 2022-02-11
Fixed standard Role Usage, role and orgs list pointed to license obje…

silvinux:
Fixed standard Role Usage, role and orgs list pointed to license object instead organizations

### What does this PR do?
Just changed README.md. At the "Playbook Examples"  the role and the data structure organization pointed to another list (license) instead organizations.

It's no needed, because it affects just the README.md file

resolves #<number>


djdanielsson (approved):
oh nice catch

PR #269, opened 2022-02-12
Added variable to control if galaxy_credential should be added and added async task to manage organization object

silvinux:
### What does this PR do?
Added variable "assign_galaxy_organizations_to_org" to control if galaxy_credential should be omited when organization will be created. Galaxy credential must exists before try to  added as a organization credential.
 
Also an async task has been added to manage organization object when there are a bunch of objects to be added.

Automated tests are preferred, but not always doable - especially for infrastructure. Include commands to run your new feature, and also post-run commands to validate that it worked. (please use code blocks to format code samples)

resolves #<number>

### DIFF
--- a/plugins/lookup/controller_object_diff.py
+++ b/plugins/lookup/controller_object_diff.py
@@ -116,17 +116,21 @@ class LookupModule(LookupBase):
                 raise AnsibleLookupError("Unable to find items in api_list")
             return [api_list]
 
-        # Set Keys to keep for each list.
-        keys_to_keep = ["name", "organization"]
-        api_keys_to_keep = ["name", "summary_fields"]
+        # Set Keys to keep for each list. Depending on type
+        if api_list[0]["type"] == "organization":
+            keys_to_keep = ["name"]
+            api_keys_to_keep = ["name"]
+        else:
+            keys_to_keep = ["name", "organization"]
+            api_keys_to_keep = ["name", "summary_fields"]
 
-        # Depending on type, keep additional keys
-        if api_list[0]["type"] == "credential":
-            keys_to_keep.append("credential_type")
-            api_keys_to_keep.append("credential_type")
-        if api_list[0]["type"] == "inventory_source":
-            keys_to_keep.append("inventory")
-            api_keys_to_keep.append("inventory")
+            # Depending on type, keep additional keys
+            if api_list[0]["type"] == "credential":
+                keys_to_keep.append("credential_type")
+                api_keys_to_keep.append("credential_type")
+            if api_list[0]["type"] == "inventory_source":
+                keys_to_keep.append("inventory")
+                api_keys_to_keep.append("inventory")
 
         for item in compare_list:
             for key in keys_to_keep:
@@ -142,10 +146,11 @@ class LookupModule(LookupBase):
         compare_list_reduced = [{key: item[key] for key in keys_to_keep} for item in compare_list]
         api_list_reduced = [{key: item[key] for key in api_keys_to_keep} for item in api_list]
 
-        # Convert summary field name into org name
-        for item in api_list_reduced:
-            item.update({"organization": item["summary_fields"]["organization"]["name"]})
-            item.pop("summary_fields")
+        # Convert summary field name into org name Only if not type organization
+        if api_list[0]["type"] != "organization":
+            for item in api_list_reduced:
+                item.update({"organization": item["summary_fields"]["organization"]["name"]})
+                item.pop("summary_fields")
 
         # Find difference between lists
         difference = [i for i in api_list_reduced if i not in compare_list_reduced]
--- a/roles/master_role_example/tasks/main.yml
+++ b/roles/master_role_example/tasks/main.yml
@@ -19,4 +19,19 @@
     loop_var: "__controller_***********_item"
   no_log: "{{ controller_configuration_*******_secure_logging }}"
   when: controller_************ is defined
+    async: 1000
+  poll: 0
+  register: __controller_***********_job_async
+
+- name: "Configure *********** | Wait for finish the *********** creation"
+  async_status:
+    jid: "{{ __controller_***********_job_async_results_item.ansible_job_id }}"
+  register: __controller_***********_job_async_result
+  until: __controller_***********_job_async_result.finished
+  retries: 10
+  loop: "{{ __controller_***********_job_async.results }}"
+  loop_control:
+    loop_var: __controller_***********_job_async_results_item
+  when: __controller_***********_job_async_results_item.ansible_job_id is defined
+  no_log: "{{ controller_configuration_*******_secure_logging }}"
 ...
--- a/roles/organizations/README.md
+++ b/roles/organizations/README.md
@@ -51,7 +51,7 @@ This role accepts two data models. A simple straightforward easy to maintain mod
 |`notification_templates_error`|""|no|The notifications on error to use for this organization in a list.|
 |`notification_templates_approvals`|""|no|The notifications for approval to use for this organization in a list.|
 |`state`|`present`|no|Desired state of the resource.|
-|`assign_galaxy_credentials_to_org`|""|no|Boolean to indicate whether credentials should be assigned or not. It should be noted that credentials must exist before add it. |
+|`assign_galaxy_credentials_to_org`|`True`|no|Boolean to indicate whether credentials should be assigned or not. It should be noted that credentials must exist before adding it. |
 
 ### Standard Organization Data Structure model
 #### Json Example
--- a/roles/organizations/README.md
+++ b/roles/organizations/README.md
@@ -51,6 +51,7 @@ This role accepts two data models. A simple straightforward easy to maintain mod
 |`notification_templates_error`|""|no|The notifications on error to use for this organization in a list.|
 |`notification_templates_approvals`|""|no|The notifications for approval to use for this organization in a list.|
 |`state`|`present`|no|Desired state of the resource.|
+|`assign_galaxy_credentials_to_org`|""|no|Boolean to indicate weather credentials should be assigned or not. It should be noted that credentials must exist before add it. |
 
 ### Standard Organization Data Structure model
 #### Json Example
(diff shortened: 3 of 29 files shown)


roles/organizations/tasks/main.yml
```diff
@@ -7,7 +7,7 @@
     max_hosts:                          "{{ __controller_organizations_item.max_hosts | default(omit, true) }}"
     instance_groups:                    "{{ __controller_organizations_item.instance_groups | default(omit, true) }}"
     default_environment:                "{{ __controller_organizations_item.default_environment | default(omit, true) }}"
-    galaxy_credentials:                 "{{ __controller_organizations_item.galaxy_credentials | default(omit, true) }}"
+    galaxy_organizations:                 "{{ (__controller_organizations_item.galaxy_organizations | default(omit)) if (assign_galaxy_organizations_to_org is defined and assign_galaxy_organizations_to_org) else omit }}"
```
sean-m-sullivan: This needs to stay as an option, its the relevant field from the module [here](https://github.com/ansible/awx/blob/b26c1c16b985d523c8aa4c35ec9e95bb4f679133/awx_collection/plugins/modules/organization.py#L83)
We can Add a field for the galaxy organization, but want to keep parity with the original module fields.

roles/organizations/README.md
```diff
@@ -51,6 +51,7 @@ This role accepts two data models. A simple straightforward easy to maintain mod
 |`notification_templates_error`|""|no|The notifications on error to use for this organization in a list.|
 |`notification_templates_approvals`|""|no|The notifications for approval to use for this organization in a list.|
 |`state`|`present`|no|Desired state of the resource.|
+|`assign_galaxy_credentials_to_org`|""|no|Boolean to indicate weather credentials should be assigned or not. It should be noted that credentials must exist before add it. |
```
sean-m-sullivan: Typo, fix to whether

roles/organizations/tasks/main.yml
```diff
@@ -24,5 +24,20 @@
   loop: "{{ controller_organizations }}"
   loop_control:
     loop_var: __controller_organizations_item
+  async: 1000
```
sean-m-sullivan, line 27: I wasn't aware a job was made to create an organization, this is interesting, and may be needed, but may have another solution.

roles/projects/tasks/main.yml
```diff
@@ -39,4 +39,19 @@
   loop_control:
     loop_var: __controller_project_item
   no_log: "{{ controller_configuration_projects_secure_logging }}"
+  async: 1000
+  poll: 0
+  register: __projects_job_async
+
+- name: "Configure Controller Projects | Wait for finish the projects creation"
```
sean-m-sullivan: How does this work with the [project_item.wait,](https://github.com/ansible/awx/blob/b26c1c16b985d523c8aa4c35ec9e95bb4f679133/awx_collection/plugins/modules/project.py#L136) 
That should be set to wait for project to be completed, rather then using a seperate task. Might it be easier to set an option to default to always wait, similar to no log?
Like
```
wait: "{{ controller_wait_on_projects | default(__controller_project_item.wait | default(omit, true)) }}"
```

sean-m-sullivan (changes requested):
Left feedback on three items to review/change before merge. 
This looks really good, and had not thought of the organization implications in the object diff and credentials!

roles/organizations/tasks/main.yml
silvinux: I messed up changing it. I didn't mean to change the variable name.

roles/organizations/README.md
silvinux: Fixed.

roles/organizations/tasks/main.yml
silvinux: We normally execute tasks as async to speed up things. when you have a lots objects to add, you can do it like this and will be quicker.

roles/projects/tasks/main.yml
silvinux: As I said before, is just to speed up the process. Then add a task to control the previous task. When you have a buch of objects to add is helpful.

sean-m-sullivan: I think the job Id is what threw me off, thinking ti was waiting on a Tower job :), If this speeds things up, Especially workflows with their nodes!!, we should add it to more objects :)

roles/organizations/tasks/main.yml
```diff
@@ -7,7 +7,7 @@
     max_hosts:                          "{{ __controller_organizations_item.max_hosts | default(omit, true) }}"
     instance_groups:                    "{{ __controller_organizations_item.instance_groups | default(omit, true) }}"
     default_environment:                "{{ __controller_organizations_item.default_environment | default(omit, true) }}"
-    galaxy_credentials:                 "{{ __controller_organizations_item.galaxy_credentials | default(omit, true) }}"
+    galaxy_credentials:                 "{{ (__controller_organizations_item.galaxy_organizations | default(omit)) if (assign_galaxy_organizations_to_org is defined and assign_galaxy_organizations_to_org) else omit }}"
```
sean-m-sullivan: Went and consulted your code in gitlab here It should be 
```
    galaxy_credentials: "{{ (__controller_organizations_item.galaxy_credentials | default(omit)) if (assign_galaxy_credentials_to_org is defined and assign_galaxy_credentials_to_org) else omit }}"
```
The key 
     __controller_organizations_item.galaxy_organizations 
needs to be changed to credentials

sean-m-sullivan:
Found one more thing, a changelog fragment file needs to be created in the 
https://github.com/redhat-cop/tower_configuration/tree/devel/changelogs/fragments

An example can be found in the[ legacy tower changes](https://github.com/redhat-cop/tower_configuration/blob/legacy_tower_v1/changelogs/fragments/playbooks.yml)

Reference here
https://github.com/ansible-community/antsibull-changelog/blob/main/docs/changelogs.rst#changelog-fragment-categories

Particularly a breaking change like so
```
---
breaking_changes:
  - galaxy credentials in the organization role now require assign_galaxy_organizations_to_org to be true. See 
minor_changes:
  - Added asynchronous to organization task to speed up creation.
  - Update to controller_object_diff lookup plugin to better handle organizations.
...
```

sean-m-sullivan:
One last thought, should 'assign_galaxy_organizations_to_org' be set to true by default in the role defaults as that is what most would do, except for the first time creating new galaxy credentials?

roles/organizations/tasks/main.yml
silvinux: Sorry, I missed. I'll take a deeper look tomorrow. Thxs ;)

silvinux:
Regarding 

This is a chicken egg problem. We were discussiong it, and we think that should be not set. The organization is the object that everything should depend on, so basically I think we have this options:

1.- Create Organization, then create the credential and last the credential can be added  to the organization.
2.- Create just the galaxy credentials, then the organizations and last the other credentials.

It's tricky. 

We took the first option, and we've implemented first creating the organizations with variable `assign_galaxy_credentials_to_org` not set, then when the credentials have been created, the variable is set to true after you make sure that the credential is created. 

$ cat roles/controller_casc_implementation/tasks/organizations.yml
```
- name: Include Tasks to config galaxy credentials to Organizations
  include_tasks: organizations.yml
  vars:
    assign_galaxy_credentials_to_org: true
```

$ cat roles/controller_casc_implementation/tasks/credentials.yml
```
- name: Configure Controller Credentials | Launch credential creation
  import_role:
    name: redhat_cop.controller_configuration.credentials

- name: Include Tasks to config galaxy credentials to Organizations
  include_tasks: organizations.yml
  vars:
    assign_galaxy_credentials_to_org: true
```

plugins/lookup/controller_object_diff.py
```diff
@@ -116,17 +116,21 @@ def run(self, terms, variables=None, **kwargs):
...
-        api_keys_to_keep = ["name", "summary_fields"]
-
-        # Depending on type, keep additional keys
-        if api_list[0]["type"] == "credential":
-            keys_to_keep.append("credential_type")
-            api_keys_to_keep.append("credential_type")
-        if api_list[0]["type"] == "inventory_source":
-            keys_to_keep.append("inventory")
-            api_keys_to_keep.append("inventory")
+        # Set Keys to keep for each list. Depending on type
```
sean-m-sullivan, line 119: I just did a comparison to #270 and this looks to be duplicated still, I got that the text was identical sans an extra line at the end of file.

Tompage1994 (changes requested):
Thanks for the contribution.

I'm not 100% sure I like the changes to placing in the galaxy credential here. Personally I usually follow a workflow of calling the credential role first and giving that a slimmed down set of vars (i.e. only the galaxy credentials) and then calling the organisation role before calling the credential role later in the usual way. I wonder whether it would be better, simpler, and prevent a breaking change, to simply update our examples with that workflow (and perhaps add some docs to it)

Regarding the async stuff I do like that, though I would like to see that against all of the roles rather than just some. If you could do that, that would be great.

Thanks

roles/master_role_example/tasks/main.yml
```diff
@@ -19,4 +19,19 @@
     loop_var: "__controller_***********_item"
   no_log: "{{ controller_configuration_*******_secure_logging }}"
   when: controller_************ is defined
+    async: 1000
```
djdanielsson, line 22: looks like your spacing is off here

roles/projects/tasks/main.yml
```diff
@@ -39,19 +39,4 @@
...
-    jid: "{{ __projects_job_async_results_item.ansible_job_id }}"
-  register: __projects_job_async_result
-  until: __projects_job_async_result.finished
-  retries: 10
-  loop: "{{ __projects_job_async.results }}"
-  loop_control:
-    loop_var: __projects_job_async_results_item
-  when: __projects_job_async_results_item.ansible_job_id is defined
-  no_log: "{{ controller_configuration_projects_secure_logging }}"
 ...
```
djdanielsson, line 42: did you mean to remove the async from projects?

djdanielsson (commented):
i have a couple of questions

roles/projects/tasks/main.yml
sean-m-sullivan: no, I had expected to update my merge before this got merged in, I'm working on variables for all the async to merge in, this will be fixed in another PR.

silvinux:
> 0% sure I like the change
I guess you are right, to keep the collection generic that var shouldn't be added. Either way organizations or credentials must be run twice.  cheers

PR #271, opened 2022-02-15
add more async

sean-m-sullivan:
### What does this PR do?
Update to make async more customizable, add role variables to overwrite specific roles, and a global to be abel to set retries and delay. 

Automated tests

### DIFF
--- a/roles/projects/defaults/main.yml
+++ b/roles/projects/defaults/main.yml
@@ -35,7 +35,7 @@ controller_projects: []
 #  notification_templates_error:   []  # optional, list of notification template names
 #  state: present  # optional, choices: present, absent
 
-controller_configuration_projects_secure_logging: "{{controller_configuration_secure_logging | default('false') }}"
+controller_configuration_projects_secure_logging: "{{ controller_configuration_secure_logging | default('false') }}"
 controller_configuration_projects_async_retries: "{{ controller_configuration_async_retries | default(30) }}"
 controller_configuration_projects_async_delay: "{{ controller_configuration_async_delay | default(1) }}"
 ...
--- a/roles/projects/defaults/main.yml
+++ b/roles/projects/defaults/main.yml
@@ -35,5 +35,7 @@ controller_projects: []
 #  notification_templates_error:   []  # optional, list of notification template names
 #  state: present  # optional, choices: present, absent
 
-controller_configuration_projects_secure_logging: "{{controller_configuration_secure_logging | default('false')}}"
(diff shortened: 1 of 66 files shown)


roles/projects/defaults/main.yml
```diff
@@ -35,5 +35,7 @@ controller_projects: []
 #  notification_templates_error:   []  # optional, list of notification template names
 #  state: present  # optional, choices: present, absent
 
-controller_configuration_projects_secure_logging: "{{controller_configuration_secure_logging | default('false')}}"
+controller_configuration_projects_secure_logging: "{{controller_configuration_secure_logging | default('false') }}"
```
djdanielsson: missing space 
controller_configuration_projects_secure_logging: "{{ controller_configuration_secure_logging | default('false') }}"

roles/projects/defaults/main.yml
```diff
@@ -35,5 +35,7 @@ controller_projects: []
...
+controller_configuration_projects_async_retries: "{{ controller_configuration_async_retries | default(30) }}"
```
djdanielsson, line 39: what do you think of something like this
controller_configuration_projects_async_retries: "{{ controller_configuration_async_retries  | default((controller_projects | length)) | default(30) }}"

roles/projects/defaults/main.yml
```diff
@@ -35,5 +35,7 @@ controller_projects: []
...
+controller_configuration_projects_async_delay: "{{ controller_configuration_async_delay | default(1) }}"
```
djdanielsson, line 40: maybe
controller_configuration_projects_async_delay: "{{ controller_configuration_async_delay | default((controller_projects | length)*.5) | default(1) }}"

roles/projects/defaults/main.yml
sean-m-sullivan: Fixed this.

roles/projects/defaults/main.yml
sean-m-sullivan: The problem with this is that unless there are 0 projects, it will never default to 30, and with 10 projects it will default to 10, which I've seen the projects take more then, even with 3 projects. probably best to stick to a number, and let user overwrite. It might make more sense with job templates or other non update items.

roles/projects/defaults/main.yml
sean-m-sullivan: See above.

sean-m-sullivan:
Updated everything, set the minimums, variable overrides, on all the applicable roles

djdanielsson (approved):
LGTM

PR #282, opened 2022-03-14
Add organization support to user creation

ctodea:
### What does this PR do?
Allow setting the organisation when creating users.

The awx collection already supports specifying the organization when creating users (see https://github.com/ansible/awx/blob/devel/awx_collection/plugins/modules/user.py#L141).
 

The change can easily be tested running the playbook below, on an automation controller. 
Before this PR, the `organization` field within `controller_user_accounts` would be ignored, and the user wouldn't be a member of any organization. Variable can be left empty, and has a ` | default(omit)`.

```
- name: Playbook to configure ansible controller
  hosts: automationcontroller
  vars:
    controller_validate_certs: false
    controller_hostname: aap00
    controller_username: admin
    controller_password: change_me
  collections:
    - awx.awx
    - redhat_cop.controller_configuration
  roles:
    - name: organizations
      vars:
        controller_organizations:
          - name: "evil_corp"
            custom_virtualenv: "/var/lib/awx/myvenv"
            max_hosts: 0
            state: present 

    - name: users
      vars:
        controller_user_accounts:
          - username: "ctodea" 
            email: "ctodea@redhat.com"
            state: present
            organization: "evil_corp" # this will make the user a member of the org
```

No

N/A

### DIFF
--- a/roles/users/README.md
+++ b/roles/users/README.md
@@ -58,6 +58,7 @@ This also speeds up the overall role.
 |`last_name`|""|no|str|The last name of the user|
 |`is_superuser`|false|no|bool|Whether the user is a superuser|
 |`is_system_auditor`|false|no|bool|Whether the user is an auditor|
+|`organization`|""|no|str|The name of the organization the user belongs to|
 |`state`|`present`|no|str|Desired state of the resource.|
 |`update_secrets`|true|no|bool| True will always change password if user specifies password, even if API gives $encrypted$ for password. False will only set the password if other values change too.|
 
--- a/roles/users/README.md
+++ b/roles/users/README.md
@@ -58,7 +58,7 @@ This also speeds up the overall role.
 |`last_name`|""|no|str|The last name of the user|
 |`is_superuser`|false|no|bool|Whether the user is a superuser|
 |`is_system_auditor`|false|no|bool|Whether the user is an auditor|
-|`organization`|""|no|str|The name of the organization the user belongs to|
+|`organization`|""|no|str|The name of the organization the user belongs to.<br />Added in awx.awx >= 20.0.0 and ansible.controller >= 4.1.1|
 |`state`|`present`|no|str|Desired state of the resource.|
 |`update_secrets`|true|no|bool| True will always change password if user specifies password, even if API gives $encrypted$ for password. False will only set the password if other values change too.|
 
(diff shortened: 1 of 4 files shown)


roles/users/tasks/main.yml
```diff
@@ -10,6 +10,7 @@
     is_superuser:             "{{ __controller_user_accounts_item.is_superuser | default(__controller_user_accounts_item.superuser | default(omit)) }}"
     is_system_auditor:        "{{ __controller_user_accounts_item.is_auditor | default(__controller_user_accounts_item.is_system_auditor | default(omit)) }}"
     update_secrets:           "{{ __controller_user_accounts_item.update_secrets | default(omit) }}"
+    organization:             "{{ __controller_user_accounts_item.organization | default(__controller_user_accounts_item.organization | default(omit)) }}"
```
djdanielsson: I am not sure I understand the logic here, you are defaulting to __controller_user_accounts_item.organization which is the exact thing you have set as the main var then default omit. what is the point of the 2nd reference?

djdanielsson (changes requested):
please explain the logic before this is approved.

roles/users/README.md
```diff
@@ -58,6 +58,7 @@ This also speeds up the overall role.
 |`last_name`|""|no|str|The last name of the user|
 |`is_superuser`|false|no|bool|Whether the user is a superuser|
 |`is_system_auditor`|false|no|bool|Whether the user is an auditor|
+|`organization`|""|no|str|The name of the organization the user belongs to|
```
Tompage1994: Could you put a note in this to make clear that this will only work in awx.awx >= 20.0.0 and ansible.controller > 2.1.1 (there is no release on controller which has this implemented yet)

Tompage1994 (changes requested):
Looks good. Just the one doc change and then this is good to go

roles/users/tasks/main.yml
sean-m-sullivan: 2nd defaults are for when its different from a simpler default vs the one from export. Currently a user created using your method above, does not have the org show up in the api or in the export, so its not needed. 
simplify to 
```
"{{ __controller_user_accounts_item.organization | default(omit) }}" 
```

ctodea: Fixed.

roles/users/README.md
sean-m-sullivan: Use this code
```
|`organization`|""|no|str|The name of the organization the user belongs to.<br />Added in awx.awx >= 20.0.0 and ansible.controller >= 4.1.1|
```

sean-m-sullivan:
Please also add a fragment file under
https://github.com/redhat-cop/controller_configuration/tree/devel/changelogs/fragments

should be something along the lines of
```
minor_changes:
  - Allow setting the organisation when creating users.
```

Tompage1994 (approved):
LGTM

djdanielsson (approved):
LGTM

PR #287, opened 2022-03-18
make async not have changed status

djdanielsson:
### What does this PR do?
Put in changed_when on all async tasks to not show as changed, also added async to project_update role.

Automated tests to make sure nothing is broken, you could verify by hand that it doesn't show anything as changed on the async task.

none

none


sean-m-sullivan (approved):
LGTM tested this for both ways. Took a while.

Issue #288, opened 2022-03-18
new async status check skips specific types


djdanielsson:
I found that the async status check gets skipped for creating credential types, inventories, and job templates

djdanielsson:
I believe I have found the issue and will submit a PR soon with the fix

PR #290, opened 2022-03-19
attempting to fix issue #288

djdanielsson:
### What does this PR do?
added missing loop control for inventories, job_templates, and credential_types

Automated plus manual check to make sure it does fix the issue as expected

#288 

None

### DIFF
--- a/roles/inventories/tasks/main.yml
+++ b/roles/inventories/tasks/main.yml
@@ -25,6 +25,7 @@
   async: 1000
   poll: 0
   register: __inventories_job_async
+  changed_when: not __inventories_job_async.changed
 
 - name: "Create Controller inventories | Wait for finish the inventories creation"
   async_status:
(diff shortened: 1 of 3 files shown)


roles/inventories/tasks/main.yml
```diff
@@ -34,7 +35,10 @@
   retries: "{{ controller_configuration_inventories_async_retries }}"
   delay: "{{ controller_configuration_inventories_async_delay }}"
   loop: "{{ __inventories_job_async.results }}"
+  loop_control:
+    loop_var: __inventories_job_async_result_item
   when: __inventories_job_async_result_item.ansible_job_id is defined
   delegate_to: "{{ 'localhost' if controller_local else omit }}"
```
sean-m-sullivan, line 41: Can you remove this line, the delegate to makes no sense, I don't know where it came from, and I found that it failed to find that variable when I tried to run it locally. Tests have gone well, will merge when change is done :)

sean-m-sullivan (approved):
Other then the 1 change that we need to put in, LGTM

PR #292, opened 2022-03-22
fix survey_spec not being an applicable option

### LINKED ISSUE #291: Cannot add survey using `survey_spec` option
None of my templates were getting their surveys generated despite the template being able to source the json file with all the questions. By the looks of it, if I had used the ‘survey’ parameter instead of ‘survey_spec’ it would have worked first time, but I think this role should probably try follow the ansible.controller job template module as much as possible and that recommends ‘survey_spec’. Also the Readme says `survey_spec` is the correct option.

```
controller_templates:
  - name: "app_add"
    description: "Add Application"
    playbook: "update_app.yml"
    project: apps
    inventory: appsinv
    credentials:
      - Ansible SSH
    survey_enabled: true
    survey_spec: "{{ lookup('file', 'update_application.json') }}"
```

Tompage1994:
### What does this PR do?
Added survey_spec as an applicable option. The Readme for the JT role states `survey_spec` is the option to use and `survey` was an alias, but the only currently usable option was `survey`

Added a CI test for survey_spec

resolves #291 

N/A

### DIFF
--- a/changelogs/fragments/survey.yml
+++ b/changelogs/fragments/survey.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - Fixes option of `survey_spec` on job_templates role.
+...
--- a/examples/configs/templates.yml
+++ b/examples/configs/templates.yml
@@ -25,6 +25,21 @@ controller_templates:
     credentials: Demo Credential
     verbosity: 0
     inventory: localhost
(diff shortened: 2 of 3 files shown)


sean-m-sullivan (approved):
Looks good. I had saw this before, but its best to use both options.

PR #293, opened 2022-03-23
Fix linting remove only role delegate_to

sean-m-sullivan:
### What does this PR do?
Fix linting remove only role delegate_to
Ran into issue where the delegate_to was failing due to lack of variable existing, not sure where it snuck in.


Tompage1994 (approved):
lgtm

PR #295, opened 2022-03-23
changing no_log: true for inventories to standard var

djdanielsson:
### What does this PR do?
changing no_log: true for inventories to standard var

Automated

none

none

### DIFF
--- a/changelogs/fragments/inventories.yml
+++ b/changelogs/fragments/inventories.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - Fixes inventories hardcoded 'no_log: true' on the async job check task.
+...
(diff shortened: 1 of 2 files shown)


Tompage1994 (changes requested):
nitpick: Could you add a changelog to this as it is a legitimate bug and should be documented

sean-m-sullivan (approved):
Good catch

djdanielsson:
> nitpick: Could you add a changelog to this as it is a legitimate bug and should be documented

ok, I tried to add it to changelog. Let me know if I didn't do it correctly.

Tompage1994 (approved):
Perfect, thanks

Issue #300, opened 2022-03-30
controller_templates missing example, can't get it to work


ikke-t:
Hi,

I have been banging my head to a problem of how to define ```execution_environment``` for ```controller_template```.

If I take out a template from AAP2.1 using the awx collection, it looks like this:

```yaml
---
assets:
    job_templates:
    -   allow_simultaneous: false
        ask_credential_on_launch: false
        ask_diff_mode_on_launch: false
        ask_inventory_on_launch: false
        ask_job_type_on_launch: false
        ask_limit_on_launch: false
        ask_scm_branch_on_launch: false
        ask_skip_tags_on_launch: false
        ask_tags_on_launch: false
        ask_variables_on_launch: false
        ask_verbosity_on_launch: true
        become_enabled: false
        description: ''
        diff_mode: false
        execution_environment:
            name: cool-lab-community
            type: execution_environment
        extra_vars: "---\nansible_python_interpreter:\
            \ /bin/python3\nnet_name: RH-seg-2991\n\
            net_type: dhcp\nmemory_mb: 2048\nnum_cpus:\
            \ 2\nstate: poweredon\n# state: absent\n\
            vm_name: rh-deleteme-01\nvcenter_credentials:\n\
            \  hostname: '{{ lookup(\"env\", \"VMWARE_HOST\"\
            ) }}'\n  username: '{{ lookup(\"env\", \"\
            VMWARE_USER\") }}'\n  password: '{{ lookup(\"\
            env\", \"VMWARE_PASSWORD\") }}'\nuse_aap_inventory:\
            \ true"
        force_handlers: false
        forks: 0
        host_config_key: ''
        inventory:
            name: Demo Inventory
            organization:
                name: Default
                type: organization
            type: inventory
        job_slice_count: 1
        job_tags: ''
        job_type: run
        limit: localhost
        name: test-vm
        natural_key:
            name: test-vm
            organization:
                name: rhops
                type: organization
            type: job_template
        playbook: ensure-vm-state.yml
        project:
            name: Cool-Lab ikke-test
            organization:
                name: rhops
                type: organization
            type: project
        related:
            credentials:
            -   credential_type:
                    kind: cloud
                    name: Red Hat Ansible Automation
                        Platform
                    type: credential_type
                name: aap-creds-rhops-admin
                organization:
                    name: rhops
                    type: organization
                type: credential
            -   credential_type:
                    kind: cloud
                    name: VMware vCenter
                    type: credential_type
[...]

PR #301, opened 2022-03-30
Execution name update

sean-m-sullivan:
### What does this PR do?
Update to use the export for job template with the execution environment name, our readme's and examples still need to be updated, but this should still work.

Fix for #300


djdanielsson (approved):
LGTM

Issue #299, opened 2022-03-30
Issue with async on projects


Tompage1994:
My customer is seeing issues with using v2.1.4 where the project creation is failing to pick up the async job id.

I haven't much more info but wanted to track this so we can test it out. I'm very busy this week so don't have the time to test this myself.

Error they get:

```
TASK [redhat_cop.controller_configuration.projects : Configure Controller Projects | Wait for finish the projects creation] *********************************

failed: [localhost] (item={'failed': 0, 'started': 1, 'finished': 0, 'ansible_job_id': '551356648062.50', 'results_file': '/home/runner/.ansible_async/551356648062.50', 'changed': False, '__controller_project_item': {'name': 'XXX | Playbooks', 'organization': 'XXX', 'scm_branch': 'master', 'scm_clean': 'yes', 'scm_credential': 'XXX | SCM Access', 'scm_delete_on_update': 'yes', 'scm_type': 'git', 'scm_update_on_launch': 'yes', 'scm_update_cache_timeout': 900, 'scm_url': 'ssh://[git@git.example.com/xxx/playbooks.git'](http://git@git.example.com/xxx/playbooks.git')}, 'ansible_loop_var': '__controller_project_item'}) => {"__projects_job_async_results_item": {"__controller_project_item": {"name": "XXX | Playbooks", "organization": "XXX", "scm_branch": "master", "scm_clean": "yes", "scm_credential": "XXX | SCM Access", "scm_delete_on_update": "yes", "scm_type": "git", "scm_update_cache_timeout": 900, "scm_update_on_launch": "yes", "scm_url": "ssh://[git@git.example.com/xxx/playbooks.git](http://git@git.example.com/xxx/playbooks.git)"}, "ansible_job_id": "551356648062.50", "ansible_loop_var": "__controller_project_item", "changed": false, "failed": 0, "finished": 0, "results_file": "/home/runner/.ansible_async/551356648062.50", "started": 1}, "ansible_job_id": "551356648062.50", "ansible_loop_var": "__projects_job_async_results_item", "attempts": 1, "changed": false, "finished": 1, "msg": "could not find job", "results_file": "/root/.ansible_async/551356648062.50", "started": 1, "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}
```

If someone has a chance to take a look that would be super helpful
Thanks

sean-m-sullivan:
Version of ansible and tower?

djdanielsson:
[...]

PR #302, opened 2022-03-31
Add async workaround

sean-m-sullivan:
### What does this PR do?
Add async workaround as discussed in #299
Reduced # of retries on the failing wait for controller to come up, its generally working with this but random things then fail

Tested on cli and on a controller with various execution environments, no issue.

#299


djdanielsson (approved):
looks good

PR #306, opened 2022-04-09
proposed fix for #305

### LINKED ISSUE #305: inventory sources interval
Hi,

First, thank you very much for this excellent solution to simplify IaC for AWX.

When updating the source inventory, I have a problem because I have an inventory that takes more than two minutes to complete. So I noticed that I can adjust the interval to add a delay in the verification of the async task. However I notice that this interval does not seem to be respected. I try with an interval of 10 seconds. So 300 seconds max.

https://github.com/redhat-cop/controller_configuration/blob/b52f3aed66810d7abd35aa8fe5706d3690174a6c/roles/inventory_source_update/tasks/main.yml

Do I understand the option and the use?

```
---
awx_inventory_sources:
  - name: Source XYZ
    description: Description....
    inventory: MyInventory
    credential: MyCredential
    overwrite: true
    overwrite_vars: true
    update_on_launch: false
    update_on_project_update: false
    verbosity: 1
    organization: SAQ
    source: scm
    source_path: inventories/my_stuff_now.yml
    source_project: MyProject
    interval: 10
    wait: true
```

```
    - role: inventory_source_update
      vars:
        controller_inventory_sources: "{{ combined_vars.awx_inventory_sources }}"
      when: combined_vars.awx_inventory_sources is defined
      tags:
        - inventory_sources
```

Thx in advance!

djdanielsson:
### What does this PR do?
Should fix the issue found in 305

Maybe automated is fine

resolves #305 

none

### DIFF
--- a/roles/inventory_source_update/tasks/main.yml
+++ b/roles/inventory_source_update/tasks/main.yml
@@ -6,7 +6,7 @@
     inventory:                "{{ __inventory_source_update_item.inventory }}"
     organization:             "{{ __inventory_source_update_item.organization | default(omit, true) }}"
     wait:                     "{{ __inventory_source_update_item.wait | default(omit) }}"
-    interval:                 "{{ __inventory_source_update_item.interval | default(1) }}"
+    interval:                 "{{ __inventory_source_update_item.interval | default(controller_configuration_inventory_source_update_async_delay) }}"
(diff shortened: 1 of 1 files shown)


sean-m-sullivan (approved):
This looks good to me, we have ways to override, the default is still 1,

PR #312, opened 2022-04-25
Change default for interval for other roles

Tompage1994:
### What does this PR do?
Changed the default value for interval in roles we are using async. Also updated readme to mention default is the delay var

No real functional changes here

N/A

Related to #306 - The same change for the other relevant modules.
Also related to #305


djdanielsson (approved):
oh I did forget to update the readme's on this change, good catch.

PR #313, opened 2022-04-26
Update docs and tests with Tower references to Controller

### LINKED ISSUE #279: Remove references to tower_cli from examples/configs/templates.yml
References to tower_cli should be removed and replaced either with collection usage or perhaps curl/API example (we could possibly use awxkit but it's not supported):

```
$ grep -r tower.cli 
grep: .git/objects/pack/pack-d920d9645030e1328d1dadaceebeabe610a63afb.pack: binary file matches
examples/configs/templates.yml:# tower-cli job_template survey --name="my job template" -f yaml
examples/configs/templates.yml:# https://github.com/ansible/tower-cli/blob/master/docs/source/cli_ref/usage/SURVEYS.rst
examples/configs/templates.yml:# You can retrieve a job along with its survey spec using tower cli as follows:
examples/configs/templates.yml:# tower-cli receive --job_template <Name> --user admin --format yaml
```

Tompage1994:
### What does this PR do?
There were a load of old references to Tower. This removes the vast majority of them. There are some in the examples where they are the names of objects where I didn't want to break anything so left them for now. Also removes the tower-cli reference

Pretty much all doc change but ideally would confirm the ci still works

resolves #279

### DIFF
--- a/examples/configs/controller_auth.yml
+++ b/examples/configs/controller_auth.yml
@@ -1,6 +1,6 @@
-# User may add tower auth creds to this file and encrypt it using `ansible-vault`
+# User may add controller auth creds to this file and encrypt it using `ansible-vault`
 ---
-# controller_hostname: ansible-tower-web-svc-test-project.example.com
+# controller_hostname: ansible-controller-web-svc-test-project.example.com
 controller_username: admin
 controller_password: password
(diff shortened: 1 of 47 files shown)


djdanielsson (approved):
LGTM

PR #314, opened 2022-04-26
Add dispatch role

### LINKED ISSUE #309: Dispatcher Role
This was brought up at the COP meeting, 
We should have a dispatcher role that invokes the rest of the roles. 
so like
redhat_cop.controller_configuration.controller_config_dispatch
will call this role set
https://github.com/redhat-cop/controller_configuration/blob/f99a1c9e020e3a6f1dbb6ace280840397e7f757b/playbooks/configure_controller.yml#L50

Tompage1994:
### What does this PR do?
Adds the dispatcher role discussed in #309.

I have updated the CI to make use of this role. The `roles/groups/tests/test.yml` file also gives a good example of usage

resolves #309 

N/A

### DIFF
--- a/changelogs/fragments/dispatch.yml
+++ b/changelogs/fragments/dispatch.yml
@@ -1,0 +1,4 @@
+---
+major_changes:
+  - Adds dispatch role - A role to run all other roles.
+...
--- a/examples/configure_controller.yml
+++ b/examples/configure_controller.yml
@@ -39,29 +39,7 @@
         - always
 
   roles:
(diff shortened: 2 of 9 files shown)


sean-m-sullivan (approved):
LGTM, I like it, its different then how I thought of using it, the variables definitely make it more versatile.

djdanielsson (approved):
It is better then just a role calling roles. I still am not a huge fan of the idea but I don't have good enough reasons to not approve it.

PR #327, opened 2022-06-18
fix default behavior for controller_host in Manage Workflows task

adonisgarciac:
### What does this PR do?
Configure the same behavior for controller_host parameter in Manage Workflows than the rest of roles with the similar type of task. I had an issue with this because I set controller_host as a var from the main playbook and it is working properly for all objects in the tower/controller but fail for workflows because the behavior of controller_host parameter is different.

N/A

N/A

N/A


Tompage1994 (approved):
LGTM. Not sure why this was set like this.

PR #329, opened 2022-07-01
default omit interval option so it will match upstream default

Tompage1994:
### What does this PR do?
Omits the intervals because the upstream default has changed and this previously wouldn't have followed. It makes more sense for us to omit so the awx.awx/ansible.controller default is used instead

N/A

N/A

https://github.com/ansible/awx/pull/12094


djdanielsson (approved):
LGTM

sean-m-sullivan (approved):
Good catch, LGTM

PR #333, opened 2022-08-16
CasC roles Automation-Iberia

silvinux:
### What does this PR do?
 - Add Configuration as Code implementation ansible role - A role to implement objects roles under a hierarchical and scalable directory structure.
 - Add Configuration as Code desired state role ansible role - A role to manage the Controller's configuration desired state
 - Add Configuration as Code from aap ansible role - A role to export and convert all  Controller's object configuration in yaml files to be consumed with previous roles.

We’ve added yaml files in the examples directory, which can be used as a baseline to perform tests. All Files should be modified according to your needs, controller host, user, password and credentials to be used, among other things.

Procedure to perform tests:
```bash
$ git clone [git@github.com](mailto:git@github.com):automationiberia/controller_configuration.git
$ cd controller_configuration/examples
$ tree -L 4 org_vars/
org_vars/
├── env
│   ├── dev
│   │   └── configure_connection_controller_credentials.yml
│   └── prod
│       └── configure_connection_controller_credentials.yml
└── OrgExample
    └── env
        ├── common
        │   ├── controller_credential_types.d
        │   ├── controller_groups.d
        │   ├── controller_instance_groups.d
        │   ├── controller_inventories.d
        │   ├── controller_job_templates.d
        │   ├── controller_organizations.d
        │   ├── controller_projects.d
        │   ├── controller_roles.d
        │   ├── controller_schedules.d
        │   ├── controller_teams.d
        │   ├── controller_users.d
        │   └── controller_workflow_job_templates.d
        ├── dev
        │   ├── controller_credentials.d
        │   ├── controller_execution_environments.d
        │   ├── controller_hosts.d
        │   ├── controller_inventory_sources.d
        │   └── controller_settings.d
        └── prod
            ├── controller_credentials.d
            ├── controller_execution_environments.d
            ├── controller_hosts.d
            ├── controller_inventory_sources.d
            └── controller_settings.d

30 directories, 2 files

$ vim orgs_vars/env/dev/configure_connection_controller_credentials.yml
$ ansible-vault encrypt orgs_vars/env/dev/configure_connection_controller_credentials.yml

$  ansible-playbook roles/casc_implementation/tests/config-controller.yml --tags organizations,credentials -e "{orgs: OrgExample, dir_orgs_vars: orgs_vars, env: dev }" -e @orgs_vars/env/dev/configure_connection_controller_credentials.yml --vault-password-file ./.vault_pass.txt

$  ansible-playbook roles/casc_desired_state/tests/desired-state.yml --tags desired_organizations,desired_state_credentials -e "{orgs: OrgExample, dir_orgs_vars: orgs_vars, env: dev }" -e @orgs_vars/env/dev/configure_connection_controller_credentials.yml --vault-password-file ./.vault_pass.txt

$ ansible-playbook roles/casc_from_aap/tests/extract_from_aap.yml -e '{dir_orgs_vars: orgs_vars, env: dev, input_tag: [all]}' -e @orgs_vars/env/dev/configure_connection_controller_credentials.yml --vault-password-file ./.vault_pass.txt
```


roles/casc_desired_state/.travis.yml
```diff
@@ -0,0 +1,30 @@
+---
```
Tompage1994: Everywhere else we have gotten rid of this file. Could you also please

Tompage1994 (changes requested):
Thanks for the contribution on this. I like the idea of some of this but I think there may be opportunities to slightly adapt some of it to fit into the collection as it is today slightly better whilst offering the same functionality you're adding here. Right off the bat I'll say I'm definitely not a fan of the file tree method for including variables but I can see some will want to use that and that's fine. That said, I don't like there is some level of enforcing here. Up until now we have been completely agnostic with how you want to include your variables and I think we should stay that way. That said, we could definitely offer a role which does the loading of the vars from a file tree as you've got here. That way we can reuse.

My thoughts on the desired state role (casc_desired_state), we could add the functionality to each already existing role without too much work. I can't think off the top of my head, but I'm sure we could essentially optionally run the second block you have in each of the tasks files in that role in each individual endpoint role in the collection.

With casc_implementation, I feel like if you take out the loading of the vars (which as previously mentioned we could pull into its own role) then it's essentially doing the same thing as the `dispatch` role.

The casc_from_aap could potentially be part of the same role as loading from the file tree structure (or its own role to create the structure as exists here. 

To summarise, I think having 1 or 2 roles to create and read the file tree structure is a good idea, and the desired_state role I think could be dispersed into the other roles. If you guys have time its probably worth jumping on a call to discuss these options with everyone else involved in this.

djdanielsson:
I agree Tom, I am not a fan of the tree method either. If Sean or I can ever get around to rewriting the diff module I think most of this won't be needed in the future, but because I don't think that will be very soon and it does add value for some who want this, I am fine with merging it. I also am not a fan of the names of the roles casc on each one of them seems pointless when this whole collection is about config as code and makes the names very long in the process.

roles/casc_desired_state/.travis.yml
silvinux: Yeap, sure. We'll do it.

silvinux:
We're not a big fan either of file tree method, but when customers have really big configurations with a lot of objects, you need to provide them a way to load it, and the tree method provides you a scalable way of do it. Sure, we can talk about this, we love to do it. 

Regarding the name of the roles, we can shorten it. 

Cheers,

ivarmu:
> Thanks for the contribution on this. I like the idea of some of this but I think there may be opportunities to slightly adapt some of it to fit into the collection as it is today slightly better whilst offering the same functionality you're adding here. Right off the bat I'll say I'm definitely not a fan of the file tree method for including variables but I can see some will want to use that and that's fine. That said, I don't like there is some level of enforcing here. Up until now we have been completely agnostic with how you want to include your variables and I think we should stay that way. That said, we could definitely offer a role which does the loading of the vars from a file tree as you've got here. That way we can reuse.
> 
> My thoughts on the desired state role (casc_desired_state), we could add the functionality to each already existing role without too much work. I can't think off the top of my head, but I'm sure we could essentially optionally run the second block you have in each of the tasks files in that role in each individual endpoint role in the collection.
> 
> With casc_implementation, I feel like if you take out the loading of the vars (which as previously mentioned we could pull into its own role) then it's essentially doing the same thing as the `dispatch` role.
> 
> The casc_from_aap could potentially be part of the same role as loading from the file tree structure (or its own role to create the structure as exists here.
> 
> To summarise, I think having 1 or 2 roles to create and read the file tree structure is a good idea, and the desired_state role I think could be dispersed into the other roles. If you guys have time its probably worth jumping on a call to discuss these options with everyone else involved in this.


The desired_state role is separated from the implementation one because it's deleting the objects that are not defined at our CasC input files and to avoid removing a project before the job template using that project, reducing afterwards issues with orphan objects. The execution order for the implementation role is just the opposite one, where the project must exist prior to the job template creation. So, the main difference between both implementation and desired_state roles is the order of the tasks, as they are importing the same roles from the collection (`redhat_cop.controller_configuration.job_templates` for example). Another reason is that, for our customers, it's clearer to have this concepts clearly separated instead of differentiating them through a tag, for example.

The input directory structure is only a proposal as, in fact, the path that is really useful to every role task is, for example, `      with_filetree: "{{ dir_orgs_vars }}/{{ orgs }}/env/common/controller_job_templates.d/"`. Inside that path, we let to define all the job_templates as needed/wanted, from a simple file containing 1000 job templates definitions, to a more elaborated hierarchy of sub-directories and files... so that path could be a variable with the recommended value as it's default one, but easily over-writable by the end user.

Tompage1994:
> > Thanks for the contribution on this. I like the idea of some of this but I think there may be opportunities to slightly adapt some of it to fit into the collection as it is today slightly better whilst offering the same functionality you're adding here. Right off the bat I'll say I'm definitely not a fan of the file tree method for including variables but I can see some will want to use that and that's fine. That said, I don't like there is some level of enforcing here. Up until now we have been completely agnostic with how you want to include your variables and I think we should stay that way. That said, we could definitely offer a role which does the loading of the vars from a file tree as you've got here. That way we can reuse.
> > My thoughts on the desired state role (casc_desired_state), we could add the functionality to each already existing role without too much work. I can't think off the top of my head, but I'm sure we could essentially optionally run the second block you have in each of the tasks files in that role in each individual endpoint role in the collection.
> > With casc_implementation, I feel like if you take out the loading of the vars (which as previously mentioned we could pull into its own role) then it's essentially doing the same thing as the `dispatch` role.
> > The casc_from_aap could potentially be part of the same role as loading from the file tree structure (or its own role to create the structure as exists here.
> > To summarise, I think having 1 or 2 roles to create and read the file tree structure is a good idea, and the desired_state role I think could be dispersed into the other roles. If you guys have time its probably worth jumping on a call to discuss these options with everyone else involved in this.
> 
> The desired_state role is separated from the implementation one because it's deleting the objects that are not defined at our CasC input files and to avoid removing a project before the job template using that project, reducing afterwards issues with orphan objects. The execution order for the implementation role is just the opposite one, where the project must exist prior to the job template creation. So, the main difference between both implementation and desired_state roles is the order of the tasks, as they are importing the same roles from the collection (`redhat_cop.controller_configuration.job_templates` for example). Another reason is that, for our customers, it's clearer to have this concepts clearly separated instead of differentiating them through a tag, for example.
> 
> The input directory structure is only a proposal as, in fact, the path that is really useful to every role task is, for example, ` with_filetree: "{{ dir_orgs_vars }}/{{ orgs }}/env/common/controller_job_templates.d/"`. Inside that path, we let to define all the job_templates as needed/wanted, from a simple file containing 1000 job templates definitions, to a more elaborated hierarchy of sub-directories and files... so that path could be a variable with the recommended value as it's default one, but easily over-writable by the end user.

I see what you mean. There is definitely some additional complexity here that I hadn't thought about. Let's set up a call for next week to go through this

sean-m-sullivan (approved):
LGTM this is good promising start!

roles/casc_desired_state/tasks/main.yml
```diff
@@ -0,0 +1,214 @@
...
+    - desired_state_groups
+    - desired_state_hosts
+    - desired_state_inventory_sources
+    - desired_state_inventories
+    - desired_state_projects
+    - desired_state_credentials
+    - desired_state_organizations
+
+- name: Include Tasks to configure Controller Workflow Job Templates
+  include_tasks: workflow_job_templates_desired_state.yml
```
Tompage1994: Will this actually work? What if the project doesn't exist yet?

Tompage1994:
Thanks all for coming on the call today. The target outcomes following the discussion are as follows:

- casc_desired_state: We can pretty much keep this as is. I think there is some potential tidying up to be done here but nothing specific for now. I guess we can decouple the filetree part of this from this role and instead you would first run the `filetree_read` role followed by this role. I would rename this role something like `dispatch_diff`. I think this role could eventually go away if we implement the diff in each role individually based on a variable, but we felt it isn't quite ready for that yet.

- casc_from_aap: This will essentially become a `filetree_create` role. One comment I have (which I forgot to raise) is that the way this is currently done feels a little clunky. I wonder if there is a better way other than templates to create each of the files. If not, it works for now and simply renaming the role will be good enough here.

- casc_implementation: We can remove the actual implementation part of this role and just keep the top block of each task file (where it loads the filetree into the variables). We can rename the role as something like `filetree_read`. The suggested solution would then be for a user to run `filetree_read` followed by `dispatch`


CC @sean-m-sullivan @djdanielsson please confirm you agree with this write up

djdanielsson:
> Thanks all for coming on the call today. The target outcomes following the discussion are as follows:
> 
> * casc_desired_state: We can pretty much keep this as is. I think there is some potential tidying up to be done here but nothing specific for now. I guess we can decouple the filetree part of this from this role and instead you would first run the `filetree_read` role followed by this role. I would rename this role something like `dispatch_diff`. I think this role could eventually go away if we implement the diff in each role individually based on a variable, but we felt it isn't quite ready for that yet.
> * casc_from_aap: This will essentially become a `filetree_create` role. One comment I have (which I forgot to raise) is that the way this is currently done feels a little clunky. I wonder if there is a better way other than templates to create each of the files. If not, it works for now and simply renaming the role will be good enough here.
> * casc_implementation: We can remove the actual implementation part of this role and just keep the top block of each task file (where it loads the filetree into the variables). We can rename the role as something like `filetree_read`. The suggested solution would then be for a user to run `filetree_read` followed by `dispatch`
> 
> CC @sean-m-sullivan @djdanielsson please confirm you agree with this write up

yes, sounds good to me.

roles/casc_desired_state/tasks/main.yml
ivarmu: Yes, it works, because... I've realised right now that the name of `casc_desired_state` for that role is not adequate at all, as that role is only erasing the objects at the AAP that are not defined as code. That's like this because the order issues we've discussed before, so the complete desired state idea is given by the execution of the two roles: `casc_desired_state` and `casc_implementation`. `dispatch_diff` should be a good name, but maybe `drop_diffs` should be better one.

silvinux:
Hi @sean-m-sullivan @djdanielsson @Tompage1994 , we've updated the PR with the changes with discussed yesterday. Please take a look and give some feedback. Thanks

changelogs/fragments/casc_roles.yml
```diff
@@ -0,0 +1,4 @@
+major_changes:
+  - Adds Configuration as Code implementation role - A role to implement objects roles under a hierarchical and scalable directory structure.
+  - Adds Configuration as Code desired state role - A role to manage the Controller's configuration desired state
```
djdanielsson: update the names of the roles in the change logs

djdanielsson (changes requested):
so I skimmed some of it and this was the one thing that jumped out at me. I am assuming you tested it after you made the changes to make sure it didn't break anything and if you say it still works then after updating the change log stuff I think I would be ready to approve it.

changelogs/fragments/casc_roles.yml
silvinux: We'll do it. Yes, we have tested, we did some changes in "user side", but nothing in the collection. Thanks

adonisgarciac: Thanks @djdanielsson  for you review, good appreciation. I've already changed it.

djdanielsson (commented):
I was just thinking about it, I wonder if we should have a writeup somewhere in a readme explaining who would want to use these roles vs just the normal ones

adonisgarciac:
> I was just thinking about it, I wonder if we should have a writeup somewhere in a readme explaining who would want to use these roles vs just the normal ones

I believe that adding an explanation about how use these roles vs the normal ones can be confusing. I mean, these roles take tasks from current roles of this collection which already have documentation. I think that we need to improve a little more our explanation in our READMEs of the new roles to clarify how to use them (we have this task in our to-do).

Also, it would be a good idea to add a brief explanation about using these new roles in the main README of this collection. What do you think?

examples/templates/vars/controller_execution_environments.yml
```diff
@@ -7,6 +7,6 @@ controller_execution_environments:
     image: ""
     organization: "{{ which_org }}"
     credential: ""
-    pull: "" # always, missing, or never
```
sean-m-sullivan, line 10: Not sure what happened here, can we revert?

silvinux: We were fixing lintenring issues and we didn't notice that threre were too many spaces. We've fixed. 

$ yamllint -c .github/workflow-config/.yamllint.yml examples/templates/vars/controller_execution_environments.yml
examples/templates/vars/controller_execution_environments.yml
  10:14     warning  too few spaces before comment  (comments)

silvinux: We were fixing lintenring issues and we didn't notice that threre were too many spaces. We've fixed.

$ yamllint -c .github/workflow-config/.yamllint.yml examples/templates/vars/controller_execution_environments.yml
examples/templates/vars/controller_execution_environments.yml
10:14 warning too few spaces before comment (comments)

ivarmu:
We've added some documentation about how to use the new roles and examples. Could you please review it?

ansible.cfg
```diff
@@ -1,4 +1,22 @@
 [defaults]
```
Tompage1994: I don't think you meant to commit this file

roles/filetree_create/vars/main.yml
```diff
@@ -0,0 +1,22 @@
+---
+# vars file for filetree_create
+output_path: "/tmp/filetree_output"
```
Tompage1994: I'm not sure this should be here? Surely the user may want to set their own location for this instead so it should be in defaults?

roles/filetree_read/tasks/main.yml
```diff
@@ -0,0 +1,171 @@
...
+      set_fact:
+        controller_oauthtoken: "{{ authtoken_res.json.token }}"
+        oauthtoken_url: "{{ authtoken_res.json.url }}"
+      no_log: true
+
+  tags:
+    - always
+
+- name: Include Tasks to config Settings
+  include_tasks: settings.yml
```
Tompage1994: Could we tidy this up by looping over a list of all the things we want to include and just variablising here?

Tompage1994 (changes requested):
You've done exactly what I wanted to see which is great! Thanks.

There are a couple of minor points which could be tidied up now but then we should be good

roles/filetree_create/vars/main.yml
silvinux: We have changed, to defaults, because it should be defined there.

roles/filetree_read/tasks/main.yml
silvinux: Hi @Tompage1994 , I don't think is needed, but to be align with collection, I've adapted filetree_read and object_diff roles with dispatch role. Please take a look and give us feedback. Thanks in advance.

README.md
```diff
@@ -118,6 +118,14 @@ Controller token module would be invoked with this code:
 
 ```
 
+### Automate the Automation
+Every Ansible Controller instance has it's own particularities and needs. Every administrator team has it's own practices and costumes. This collection allows to addapt to every needs, from small to large scale, having the objects distributed across multiple environments and leveraging Automation Webhook that can be used to link a Git repository and Ansible automation natively.
+
+A complete example of how to use all of the roles present in the collection is available at the following [README.md](examples/automatetheautomation/README.md), where all the phases to allow CI/CD for the Controller Configuration are provided.
```
sean-m-sullivan: Fix whitespace end of line.

roles/filetree_read/tests/config-controller-filetree.yml
```diff
@@ -0,0 +1,36 @@
...
+                - organizations
+                - credentials
+          vars:
+            assign_galaxy_credentials_to_org: false
+            controller_configuration_dispatcher_roles:
+              - {role: organizations, var: controller_organizations, tags: organizations}
+              - {role: credentials, var: controller_credentials, tags: credentials}
+
+  roles:
+    - {role: redhat_cop.controller_configuration.filetree_read }
```
sean-m-sullivan: not sure if it wants space at end removed, or a similar space at the beginning for both these lines.

sean-m-sullivan (approved):
I like the changes, this looks great, put in some comments about the linting, then it should be good to merge!

README.md
silvinux: Fixed!.

sean-m-sullivan (approved):
LGTM, going for one other approval and we can merge :) 👍

djdanielsson (approved):
LGTM

ivarmu:
@sean-m-sullivan @djdanielsson @Tompage1994, is there anything pending for this PR to get merged?

Thanks!

sean-m-sullivan:
I was giving @Tompage1994 a chance to review he was out yesterday, we should merge today, was thinking release as well

Tompage1994 (approved):
Yes, I'm happy with this. 

Thanks for all the hard work on this!

PR #334, opened 2022-08-29
Update templates.yml

IPvSean:
just adding more example...

### What does this PR do?
Adds an example with execution_environment

### How should this be tested?
Automated tests are preferred, but not always doable - especially for infrastructure. Include commands to run your new feature, and also post-run commands to validate that it worked. (please use code blocks to format code samples)

### Is there a relevant Issue open for this?
Provide a link to any open issues that describe the problem you are solving.
resolves #<number>

### Other Relevant info, PRs, etc.
Please provide link to other PRs that may be related (blocking, resolves, etc. etc.)

PR #335, opened 2022-08-30
Fix CI

Tompage1994:
### What does this PR do?
Fixes broken CI where a non existant EE was referenced

CI should pass

N/A

#334 introduced a CI failure

### DIFF
--- a/examples/configs/differential_items.yml
+++ b/examples/configs/differential_items.yml
@@ -78,18 +78,18 @@ differential_items:
         organization: Default
         credential_type: "Source Control"
         state: absent
-  - name: inventory_sources
-    with_present: false
-    differential_test_items:
-      - name:  RHVM-02
-        organization: Default
-        inventory: 3
-        state: absent
-    expected_test_result:
-      - name: RHVM-01-Default
-        organization: Default
-        inventory: 3
-        state: absent
+#  - name: inventory_sources
+#    with_present: false
+#    differential_test_items:
+#      - name:  RHVM-02
+#        organization: Default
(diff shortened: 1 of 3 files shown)


Tompage1994:
Confirmed this does fix this issue but there is another issue with the diff section

Tompage1994:
Think I've found the fix. No idea why the number has changed. Not sure I fully understand why it is a number (I guess ID)

djdanielsson (approved):
Is there a reason we use a number instead of a name?

Tompage1994:
ok, this still doesn't work. The number seems to be different each time which means this isn't going to work. I assume this is all because the controller_api lookup returns the inventory id rather than the name. not sure what we can really do about that in this case. @sean-m-sullivan any ideas?

djdanielsson (changes requested):
you should be able to use name instead of number... I think worst case we take out the diff tests and wait until it gets rewritten to add it back in

sean-m-sullivan:
The problem is in the test itself, its getting back numbers from the API, and in the future we could look it up, or something like that, its probably best just to disable the test till we can fix it.

Tompage1994:
Ok, I've done that. We shall see how much further it gets now

Tompage1994:
It failed much later this time because a field has been deprecated and removed.

See https://github.com/ansible/awx/pull/12366

sean-m-sullivan (approved):
lGTM

djdanielsson (approved):
seems to be working now, sweet!

PR #338, opened 2022-08-30
update link vars

sean-m-sullivan:
### What does this PR do?
Update link vars tested the galaxy build/install that was erring before due to bad symlinks.


djdanielsson (approved):
LGTM

PR #347, opened 2022-09-08
fix example filetre_create with wrong checks

adonisgarciac:
### What does this PR do?
By mistake, we added some asserts in the examples of filetree_create role. These asserts depend of variables declared in the role itself, then, this example would be failed. In this PR we remove some vars and asserts which are already in the role:

https://github.com/redhat-cop/controller_configuration/blob/devel/roles/filetree_create/defaults/main.yml#L5-L8
https://github.com/redhat-cop/controller_configuration/blob/devel/roles/filetree_create/tasks/main.yml#L3-L19
https://github.com/redhat-cop/controller_configuration/blob/devel/roles/filetree_create/vars/main.yml

Run test playbook

### DIFF
--- a/roles/filetree_create/README.md
+++ b/roles/filetree_create/README.md
@@ -30,29 +30,7 @@ Example Playbook
 - hosts: localhost
   connection: local
   gather_facts: false
-  vars:
-    controller_hostname: "{{ vault_controller_hostname }}"
-    controller_username: "{{ vault_controller_username }}"
-    controller_password: "{{ vault_controller_password }}"
-    controller_validate_certs: "{{ vault_controller_validate_certs }}"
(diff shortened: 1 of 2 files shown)


djdanielsson (approved):
before I merge, does this prevent the current release of the role from working? trying to understand if we need to release another version ASAP to make the roles functional or not.

adonisgarciac:
> before I merge, does this prevent the current release of the role from working? trying to understand if we need to release another version ASAP to make the roles functional or not.

In my opinion it is not necessary a new release. This mistake is only present in the examples so we can use the roles and prove them anyway.

sean-m-sullivan (approved):

PR #348, opened 2022-09-08
Export the dynamic inventories (with inventory_sources defined) without the hosts and without the groups.

ivarmu:
### What does this PR do?
Currently, the inventory object itself is not exported, so this is fixing that.

- remove the filter for inventories with inventory sources
 
- the organization's name can have an /, so we need to use another character when creating the destination paths
 
- don't export hosts that have has_inventory_sources to true
 
- increase the max_objects limit. ToDo: make a variable for that
 
- don't need the hosts list when the inventory has inventory_sources
 
- don't need the hosts list when the inventory has inventory_sources
 
- the organization's name can have an /, so we need to use another character when creating the destination paths
 
- DO need the hosts list when the inventory has inventory_sources
 
- DO NOT need the hosts list when the inventory has inventory_sources. Unaffordable by now: out of resources at large environments

`ansible-playbook filetree_create.yml -e '{input_tag: [inventory], output_path: ~/iam/filetree_create.2.1.7/}'` 

No issue opened for this bug.

N/A

### DIFF
--- a/roles/filetree_create/tasks/groups.yml
+++ b/roles/filetree_create/tasks/groups.yml
@@ -8,10 +8,10 @@
 - name: "Add current groups to the output yaml file: {{ output_path }}/current_groups_<INVENTORY_NAME>.yaml"
   template:
     src: "templates/current_groups.j2"
-    dest: "{{ output_path }}/current_groups_{{ current_groups_asset_value.name }}.yaml"
+    dest: "{{ output_path }}/current_groups_{{ current_groups_asset_value.name | regex_replace('/','_') }}.yaml"
     mode: '0644'
   loop: "{{ current_organization_inventories }}"
   loop_control:
     loop_var: current_groups_asset_value
-    label: "{{ output_path }}/current_groups_{{ current_groups_asset_value.name }}.yaml"
+    label: "{{ output_path }}/current_groups_{{ current_groups_asset_value.name | regex_replace('/','_') }}.yaml"
 ...
--- a/roles/filetree_create/tasks/groups.yml
+++ b/roles/filetree_create/tasks/groups.yml
@@ -14,4 +14,5 @@
   loop_control:
     loop_var: current_groups_asset_value
     label: "{{ output_path }}/current_groups_{{ current_groups_asset_value.name | regex_replace('/','_') }}.yaml"
+  when: current_groups_asset_value.groups | length > 0
 ...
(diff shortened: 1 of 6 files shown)


adonisgarciac (commented):
I think that it would be a good idea to add a conditional in include_tasks tasks:

- name: "Set the inventory's inventory sources"
- name: "Set the inventory's hosts"
- name: "Set the inventory's groups"

I'm not sure but it seems that these tasks will always create a exported file with the variables controller_hosts, configure_tower_groups and controller_inventory_sources even though the inventory does not have host, group or inventory_source. Am i right? 

Example of conditional that i say:

- name: "Set the inventory's inventory sources"
  include_tasks: "inventory_sources.yml"
  vars:
    output_path: "{{ local_output_path }}/{{ current_inventory_sources.key }}"
    current_organization_inventories: "{{ current_inventory_sources.value }}"
  loop: "{{ current_inventories | default({}) | dict2items }}"
  loop_control:
    loop_var: current_inventory_sources
    label: "{{ local_output_path }}/{{ current_inventory_sources.key }}"
  when: current_organization_inventories.inventory_sources | lenght > 0

ivarmu:
> I think that it would be a good idea to add a conditional in include_tasks tasks:
> 
> * name: "Set the inventory's inventory sources"
> * name: "Set the inventory's hosts"
> * name: "Set the inventory's groups"
> 
> I'm not sure but it seems that these tasks will always create a exported file with the variables controller_hosts, configure_tower_groups and controller_inventory_sources even though the inventory does not have host, group or inventory_source. Am i right?
> 
> Example of conditional that i say:
> 
> * name: "Set the inventory's inventory sources"
>   include_tasks: "inventory_sources.yml"
>   vars:
>   output_path: "{{ local_output_path }}/{{ current_inventory_sources.key }}"
>   current_organization_inventories: "{{ current_inventory_sources.value }}"
>   loop: "{{ current_inventories | default({}) | dict2items }}"
>   loop_control:
>   loop_var: current_inventory_sources
>   label: "{{ local_output_path }}/{{ current_inventory_sources.key }}"
>   when: current_organization_inventories.inventory_sources | lenght > 0

You are right, it was creating _empty_ files for hosts and groups on those inventories having external sources. I've fixed that.

@adonisgarciac eagle eye! :+1:

adonisgarciac (approved):

djdanielsson (approved):
LGTM

PR #349, opened 2022-09-08
Playbook update

sean-m-sullivan:
### What does this PR do?
Removes cancel job from the example playbooks


zcampo:
Looks good to me.  This is to address what I ran in to when trying to launch a job directly from the controller_configuration.yml playbook.

Job would launch, details would print, then immediately cancel the job since it had a when statement based on the job being launched.

branic:
LGTM

branic (approved):

PR #351, opened 2022-09-09
Let the collection to use both awx.awx and ansible.controller collections, as per user decision

ivarmu:
add variable to set the plugin name to use everywhere

### What does this PR do?
Right now, the tasks in filetree_create and object_diff roles are using the `ansible.controller` collection to make use of the `controller_api` plugin, so anyone trying to use the `awx.awx` collection will get in trouble as the `ansible.controller` collection should not be present. To let the end users to chose freely what collection they want to use, I've modified the code to get that collection name from a variable, which must be defined with the correct value to work successfully.

```
$ cat filetree_create.yml
---
- hosts: localhost
  connection: local
  gather_facts: false
  vars:
    controller_hostname: "{{ vault_controller_hostname }}"
    controller_username: "{{ vault_controller_username }}"
    controller_password: "{{ vault_controller_password }}"
    controller_validate_certs: "{{ vault_controller_validate_certs }}"
    controller_api_plugin: "ansible.controller.controller_api"
  pre_tasks:
    - name: "Check if the required input variables are present"
      assert:
        that:
          - input_tag is defined
          - (input_tag | type_debug) == 'list'
        fail_msg: 'A variable called ''input_tag'' of type ''list'' is needed: -e ''{input_tag: [organizations, projects]}'''
        quiet: true

    - name: "Check if the required input values are correct"
      assert:
        that:
          - tag_item in valid_tags
        fail_msg: "The valid tags are the following ones: {{ valid_tags | join(', ') }}"
        quiet: true
      loop: "{{ input_tag }}"
      loop_control:
        loop_var: tag_item
  roles:
    - redhat_cop.controller_configuration.filetree_create
...
$ ansible-playbook filetree_create.yml -e '{input_tag: [inventory], output_path: ~/iam/filetree_create.2.1.7 }'
```

No issue opened for this.

N/A

### DIFF
--- a/plugins/lookup/controller_object_diff.py
+++ b/plugins/lookup/controller_object_diff.py
@@ -46,20 +46,47 @@ options:
 """
 
 EXAMPLES = """
+# Set the controller_api_plugin manually:
+- name: Set the name for the collection to provide the plugin controller_api
+  set_fact:
+    controller_api_plugin: "awx.awx.controller_api"
+
+# Set the controller_api_plugin automatically:
+- block:
+    - name: "Check if the collection ansible.controller is installed"
+      set_fact:
+        ansible_controller_collection_installed: "{{ lookup('pipe', 'ansible-galaxy collection list | grep -i ansible.controllersss') }}"
+  rescue:
+    - name: "Check if the collection awx.awx is installed"
+      set_fact:
+        awx_awx_collection_installed: "{{ lookup('pipe', 'ansible-galaxy collection list | grep -i awx.awx') }}"
+  always:
+    - name: "Set the collection providing the controller_api lookup plugin"
+      set_fact:
+        controller_api_plugin: "{{ ('ansible.controller.controller_api' if ansible_controller_collection_installed is defined) | default('awx.awx.controller_api' if awx_awx_collection_installed is defined) | default('NONE') }}"
+    - name: "Fail if no collection is detected"
+      fail:
+        msg: "One of the following collections is required to be installed: 'ansible.controller' or 'awx.awx'."
+      when: controller_api_plugin is match('NONE')
+    - name: "Show the plugin we are using"
+      debug:
+        msg: "Using the 'controller_api' plugin from: {{ controller_api_plugin }}"
+
+# Examples for the lookup plugin
 - name: Get the organization ID
   set_fact:
-    controller_organization_id: "{{ lookup('awx.awx.controller_api', 'organizations', query_params={ 'name': 'Default' },
+    controller_organization_id: "{{ lookup(controller_api_plugin, 'organizations', query_params={ 'name': 'Default' },
       host=controller_hostname, username=controller_username, password=controller_password, verify_ssl=false) }}"
 
 - name: "Get the API list of all Projects in the Default Organization"
   set_fact:
-    controller_api_results: "{{ lookup('awx.awx.controller_api', 'projects', query_params={ 'organization':
+    controller_api_results: "{{ lookup(controller_api_plugin, 'projects', query_params={ 'organization':
       controller_organization_id.id } ,host=controller_hostname, username=controller_username,
       password=controller_password, verify_ssl=false) }}"
 
 - name: "Get the API in a list form. Useful for making sure the results of one item is set to a list.
   set_fact:
-    controller_api_results: "{{ query('awx.awx.controller_api', 'inventories', query_params={ 'organization':
+    controller_api_results: "{{ query(controller_api_plugin, 'inventories', query_params={ 'organization':
       controller_organization_id.id } ,host=controller_hostname, username=controller_username,
       password=controller_password, verify_ssl=false) }}"
 
--- a/plugins/lookup/controller_object_diff.py
+++ b/plugins/lookup/controller_object_diff.py
@@ -46,50 +46,20 @@ options:
 """
 
 EXAMPLES = """
-# Set the controller_api_plugin manually:
-- name: Set the name for the collection to provide the plugin controller_api
-  set_fact:
-    controller_api_plugin: "awx.awx.controller_api"
-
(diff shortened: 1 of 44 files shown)


ivarmu:
@sean-m-sullivan @Tompage1994 @djdanielsson I think it would be a good idea to have a new release based on that PR, as the current version is tied to the `ansible.controller`collection and will not work if the user only {wants to,can} have `awx.awx` collection installed. What do you think?

Tompage1994 (changes requested):
I don't necessarily like this solution but it may be the best we can do. I'm not sure if there is some magic we could do to look up what collection we are using. 

That said, I'm happy to go with this, at least for now. My only thought is that we should probably default to ansible.controller rather than awx.awx.

ivarmu:
> I don't necessarily like this solution but it may be the best we can do. I'm not sure if there is some magic we could do to look up what collection we are using.
> 
> That said, I'm happy to go with this, at least for now. My only thought is that we should probably default to ansible.controller rather than awx.awx.

I've set it to `awx.awx` to be consistent with the collection already used in the example two lines after my addition. We can change it if you preffer, but the rest of the examples and documentation is based in `awx.awx` as well...

ivarmu:
If you agree, I could add this to automatically detect what collection is installed, and default to `ansible.controller` if both of them are present:

https://docs.ansible.com/ansible/latest/collections/community/general/collection_version_lookup.html

or the following to run `ansible-galaxy collection list | grep ansible.controller` if we don't want to depend on `community.general`:

https://docs.ansible.com/ansible/latest/collections/ansible/builtin/pipe_lookup.html#ansible-collections-ansible-builtin-pipe-lookup

ivarmu:
> If you agree, I could add this to automatically detect what collection is installed, and default to `ansible.controller` if both of them are present:
> 
> https://docs.ansible.com/ansible/latest/collections/community/general/collection_version_lookup.html
> 
> or the following to run `ansible-galaxy collection list | grep ansible.controller` if we don't want to depend on `community.general`:
> 
> https://docs.ansible.com/ansible/latest/collections/ansible/builtin/pipe_lookup.html#ansible-collections-ansible-builtin-pipe-lookup

To avoid depending on `community.general` collection, I can integrate the following code if it look good for you:

```
---
- hosts: localhost
  connection: local
  gather_facts: false
  tasks:
    - block:
        - name: "Check if the collection ansible.controller is installed"
          set_fact:
            ansible_controller_collection_installed: "{{ lookup('pipe', 'ansible-galaxy collection list | grep -i ansible.controllersss') }}"
      rescue:
        - name: "Check if the collection awx.awx is installed"
          set_fact:
            awx_awx_collection_installed: "{{ lookup('pipe', 'ansible-galaxy collection list | grep -i awx.awx') }}"
    - name: "Set the collection providing the controller_api lookup plugin"
      set_fact:
        collection_name: "{{ ('ansible.controller.controller_api' if ansible_controller_collection_installed is defined) | default('awx.awx.controller_api' if awx_awx_collection_installed is defined) | default('NONE') }}"
    - debug:
        msg:
          - "collection_name: {{ collection_name }}"
...
```

djdanielsson:
I like that solution to figure out which collection is being used

ivarmu:
On Monday I'm working on it :blush:

plugins/lookup/controller_object_diff.py
```diff
@@ -46,20 +46,50 @@
...
+        msg: "One of the following collections is required to be installed: 'ansible.controller' or 'awx.awx'."
+      when: controller_api_plugin is match('NONE')
+    - name: "Show the plugin we are using"
+      debug:
+        msg: "Using the 'controller_api' plugin from: {{ controller_api_plugin }}"
+
+# Examples for the lookup plugin
 - name: Get the organization ID
   set_fact:
-    controller_organization_id: "{{ lookup('awx.awx.controller_api', 'organizations', query_params={ 'name': 'Default' },
```
sean-m-sullivan: Lets leave this as is, as the plugin can be used outside the roles, the plugin itself should be self documenting.

roles/filetree_create/README.md
```diff
@@ -15,7 +15,10 @@ Role Variables
 
 The following variables are required for that role to work properly:
 
-- **`output_path`**: The path to the output directory where all the generated `yaml` files with the corresponding Objects as code will be written to. The default path is `/tmp/filetree_output`.
+| Variable Name | Default Value | Required | Description |
+| :------------ | :-----------: | :------: | :---------- |
+| **`controller_api_plugin`**: | N/A | yes | Full path for the controller_api_plugin to be used. <br/> Can have two possible values: <br/>&nbsp;&nbsp;- awx.awx.controller_api             # For the community Collection version <br/>&nbsp;&nbsp;- ansible.controller.controller_api  # For the Red Hat Certified Collection version|
```
sean-m-sullivan: in addition to updating the readme's. Can we put a default value for this in each of the roles. So if they do not use our default playbooks with detection it doesn't fail due to the plugin not being defined. Should default to the ansible.controller collection version.

plugins/lookup/controller_object_diff.py
ivarmu: Done

roles/filetree_create/README.md
ivarmu: I've added the default values for both `filetree_create` and `object_diff` roles.

ivarmu: I was thinking that a better place to add the auto-detection for the plugin to be used should be the `main.yml` of each role, so the default value shouldn't be needed. I've updated the PR with these changes.

ivarmu:
@Tompage1994 @djdanielsson @sean-m-sullivan, I'm not sure if there's anything pending in this PR.

sean-m-sullivan (approved):
LGTM, thanks for making the changes.

PR #352, opened 2022-09-10
Added query_controller_api_max_objects to filetree_create

silvinux:
### What does this PR do?
Added variable ´query_controller_api_max_objects´ to handle the maximum number of objects to return from the list when using query at controller_api. If a list view returns more an max_objects an exception will be raised. By default is [1]max_objects=1000, when you are working in large environments this number could be short, that why it would be better to create a variable for it. 

[1].-https://docs.ansible.com/ansible/3/collections/awx/awx/tower_api_lookup.html#parameter-max_objects

### DIFF
--- a/roles/filetree_create/defaults/main.yml
+++ b/roles/filetree_create/defaults/main.yml
@@ -25,4 +25,7 @@ controller_workflow_job_templates: []
 
 # Output directory path
 output_path: "/tmp/filetree_output"
+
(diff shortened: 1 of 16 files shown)


silvinux:
Sorry, I needed to rebase.

sean-m-sullivan (approved):
Lgtm, agree, set it, allow override

PR #357, opened 2022-09-16
Remove with filetree

adonisgarciac:
### What does this PR do?
We don't want to depend of community.general so i are trying to remove every dependency with it. In this PR we remove with_filetree lookup plugin of community.general collection.

Run test playbook


sean-m-sullivan (approved):
LGTM

PR #363, opened 2022-09-22
fix a relation between role name and description

vojtechDB:
### What does this PR do?
fix a relation between role name and the description


sean-m-sullivan (approved):
LGTM

PR #365, opened 2022-09-26
Improve oauthtoken

adonisgarciac:
### What does this PR do?
This PR try to improve the managment of oauthtoken for filetree_create, filetree_read and object_diff roles and some other fixes which they were seen during tests:

1. oauthtoken var is renamed because follow naming of the rest of roles: oauthtoken -> controller_oauthtoken
2. Control if controller_oauthtoken is given, the roles will not create the oauth token
3. If oauth token is created, it will be deleted at the end.
4. Remove generate token in filetree_read because this role only is to read variable definitions.
5. Add check what collection ansible.controller or ansible.awx is installed for object_diff role
6. fix controller_object_diff.py: it was I was handling credential_type as it would have organization.

N/A

I think that, after this PR will be merged, I should generate a new release version because we have a lot of new features and fixes and we already have customers testing CasC using these roles.


silvinux (commented):
LGTM. Awesome job!.

djdanielsson (approved):
LGTM

PR #367, opened 2022-09-28
trying to fix #366

### LINKED ISSUE #366: project scm_branch bug
if you pass 'scm_branch: ""  ' and in the UI or old code had set scm_branch to something else this does not over ride the UI and set it to blank/default which should be the result.

djdanielsson:
### What does this PR do?
changed scm_branch default(omit, true) to default(omit) to hopefully allow users to provide "" and have this update controller to go back to default branch if currently not set to that.

tested by hand once to verify this solves the issue

resolves #366 

none


silvinux:
LGTM

silvinux (approved):
LGTM

PR #376, opened 2022-10-06
Fix issue in CI where lookup not found

Tompage1994:
### What does this PR do?
Fixes broke CI where the `redhat_cop.controller_configuration.controller_object_diff` is not found because the collection isn't built so it needs to look the plugin up locally.

CI will run

No

N/A


silvinux (approved):
LGTM

PR #377, opened 2022-10-07
Removing readme comments from default files

JonathanBouligny:
### What does this PR do?

Removes the comments explaining what the different variables in defaults/main.yml do because they are often out of date. Refer people to the readme instead.


djdanielsson (changes requested):
Just delete all # possible fields stuff

PR #382, opened 2022-10-10
remove auth from roles

adonisgarciac:
### What does this PR do?

We remove from roles the authentication tasks because roles themselves shouldn't do it. Authentication should be done by the playbook which use the roles or by the user sending the oauth token as a extra var.

We found problems with auth in roles because we need to use more than one role in the same playbook and authentication in roles broke our workflow.


sean-m-sullivan (approved):
these changes look good, and agreed on leaving it to the playbook, LIke that you included the information in the readme.

ivarmu (approved):

silvinux (approved):
LGTM. I'll merge this because is causing some issues.

PR #391, opened 2022-10-19
Update ansible-lint version and fix lint errors

branic:
### What does this PR do?

Updates ansible-lint to version 6.8.2 and fixes new found linting errors.

The `fqcn[action]` check has also been disabled as the collection depends on either the `awx.awx` or `ansible.controller` collection so the FQCN of the modules cannot be specified.

Automated tests are preferred, but not always doable - especially for infrastructure. Include commands to run your new feature, and also post-run commands to validate that it worked. (please use code blocks to format code samples)

No

Nnoe

### DIFF
--- a/roles/filetree_create/tasks/inventory.yml
+++ b/roles/filetree_create/tasks/inventory.yml
@@ -7,7 +7,7 @@
                                  return_all=true, max_objects=query_controller_api_max_objects)
                         }}"
 
-- name: "Create the <ORGANIZATION_NAME>/inventories output directory for inventories in {{ output_path | regex_replace('/', '_') }}"
+- name: "Create the <ORGANIZATION_NAME>/inventories output directory for inventories in {{ output_path }}"
   ansible.builtin.file:
     path: "{{ __path }}"
     state: directory
@@ -21,7 +21,7 @@
     loop_var: needed_path
     label: "{{ __path }}"
 
-- name: "Add current inventories to the <ORGANIZATION_NAME>/inventories output yaml file in {{ output_path | regex_replace('/', '_') }}"
+- name: "Add current inventories to the <ORGANIZATION_NAME>/inventories output yaml file in {{ output_path }}"
   ansible.builtin.template:
     src: "templates/current_inventories.j2"
     dest: "{{ __dest }}"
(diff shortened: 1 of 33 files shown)


roles/filetree_create/tasks/inventory.yml
```diff
@@ -7,7 +7,7 @@
                                  return_all=true, max_objects=query_controller_api_max_objects)
                         }}"
 
-- name: "Create the output directory for inventories: {{ output_path }}/<ORGANIZATION_NAME>/inventories"
+- name: "Create the <ORGANIZATION_NAME>/inventories output directory for inventories in {{ output_path | regex_replace('/', '_') }}"
```
ivarmu: output_path can't be regex_replaced entirely because it contains the full (relative) path, which must show the directories correctly. That treatment has been done at other places, where the output_path is constructed.

roles/filetree_create/tasks/inventory.yml
```diff
@@ -21,7 +21,7 @@
     loop_var: needed_path
     label: "{{ __path }}"
 
-- name: "Add current inventories to the output yaml file: {{ output_path }}/<ORGANIZATION_NAME>/inventories"
+- name: "Add current inventories to the <ORGANIZATION_NAME>/inventories output yaml file in {{ output_path | regex_replace('/', '_') }}"
```
ivarmu: output_path can't be regex_replaced entirely because it contains the full (relative) path, which must show the directories correctly. That treatment has been done at other places, where the output_path is constructed.

ivarmu (changes requested):
LGTM. Only make the requested changes to let the paths show correctly in the tasks' names.

Great job!

roles/filetree_create/tasks/inventory.yml
branic: Removed the regex_replace where used in the name of a task.

roles/filetree_create/tasks/inventory.yml
branic: I've removed the `regex_replace` for `output_path` when it is used in the name of a task.

silvinux (approved):
LGTM

djdanielsson:
I have set it to merge once @ivarmu approves the fixes

ivarmu (approved):
Thanks for these changes!

PR #400, opened 2022-10-26
add settings to filetree_create and custom_virtualenv for jt if tower

adonisgarciac:
### What does this PR do?

Add settings to filetre_create and add custom_virtualenv for job_templates if it is tower.


silvinux (commented):
LGTM, but before approve, I'd like to consider to get just the settings that has changed, don't you think? instead of getting everything. The endpoint to get just what has changed is "/api/v2/settings/changed/".

silvinux (approved):
LGTM!

PR #402, opened 2022-10-27
add links to other repos

sean-m-sullivan:
### What does this PR do?
add links to other repos


djdanielsson (changes requested):
I think it looks good, my one thought is we should keep the collection requirements above this because that still is commonly missed and if we move it down further it will just be missed more

djdanielsson (approved):
LGTM

sean-m-sullivan:
Fixed in terms of Order, Was thinking of a SPLASHY image or something, but this will do.

PR #408, opened 2022-11-10
adding option to skip adding default ee to org

djdanielsson:
### What does this PR do?

adding option to skip adding default ee to org for first time org creation

manual test should be done.

no

none


Tompage1994 (changes requested):
Seems to make sense.

Could this option be added to the readme for the role and mention its use case.

Can you also add to the changelog please

silvinux:
Awesome!!! this solve the chicken and egg problem for that!!!

Tompage1994 (approved):

PR #410, opened 2022-11-15
Update CI to release infra collection

sean-m-sullivan:
### What does this PR do?
When we do a new release this uploads collection builds to the release. This will require all new releases to be numbered as in 2.3.0 not v2.3.0, unless someone has a better sed, but that should be our standard practice anyway.

Automated tests

In action: https://github.com/sean-m-sullivan/controller_configuration/releases/tag/1.3.1


djdanielsson (approved):
LGTM

PR #416, opened 2022-11-21
Protect delete org

adonisgarciac:
# What does this PR do?

Allow user to set protect_not_empty_orgs to avoid remove orgs which are not empty. 
Only superuser can remove orgs after run object diff.

# How should this be tested?

### DIFF
--- a/roles/object_diff/tasks/organizations.yml
+++ b/roles/object_diff/tasks/organizations.yml
@@ -25,13 +25,13 @@
 
     - name: "Sets Organization differences"
       vars:
-        __api_url_users: "api/v2/organizations/{{ __org.name}}/users/"
-        __api_url_admins: "api/v2/organizations/{{ __org.name}}/admins/"
-        __api_url_inventories: "api/v2/organizations/{{ __org.name}}/inventories/"
-        __api_url_teams: "api/v2/organizations/{{ __org.name}}/teams/"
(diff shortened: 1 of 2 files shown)


roles/object_diff/tasks/organizations.yml
```diff
@@ -1,17 +1,48 @@
...
+        __api_url_projects: "api/{{ controller_api_version }}/organizations/{{ __org.name | urlencode }}/projects/"
+        __api_url_job_templates: "api/{{ controller_api_version }}/organizations/{{ __org.name | urlencode }}/job_templates/"
+        __api_url_workflow_job_templates: "api/{{ controller_api_version }}/organizations/{{ __org.name | urlencode }}/workflow_job_templates/"
+      ansible.builtin.set_fact:
+        controller_organizations: "{{ controller_organizations | combine(__org) }}"
+      loop: "{{ __organizations_difference }}"
+      loop_control:
+        loop_var: __org
+      when: protect_not_empty_orgs is not defined or not protect_not_empty_orgs or
+        ( {{ query(controller_api_plugin, __api_url_users, host=controller_hostname, oauth_token=controller_oauthtoken, verify_ssl=controller_validate_certs) | length == 0 }} and
```
ivarmu: These queries can be executed at `vars` section and at `when` only need to treat the resulting booleans.

ivarmu (changes requested):

PR #417, opened 2022-11-22
Protect delete org

adonisgarciac:
# What does this PR do?
Allow user to set protect_not_empty_orgs to avoid remove orgs which are not empty.
Only superuser can remove orgs after run object diff.

# How should this be tested?

### DIFF
--- a/roles/object_diff/tasks/organizations.yml
+++ b/roles/object_diff/tasks/organizations.yml
@@ -24,34 +24,28 @@
                                      }}"
 
     - name: "Set list __list_orgs_empty when protect_not_empty_orgs"
-      vars:
-        __bool_empty_users: "{{ query(controller_api_plugin, 'api/' + controller_api_version + '/organizations/' + (__org.name | urlencode) + '/users/',
-                          host=controller_hostname, oauth_token=controller_oauthtoken, verify_ssl=controller_validate_certs) | length == 0 }}"
-        __bool_empty_admins: "{{ query(controller_api_plugin, 'api/' + controller_api_version + '/organizations/' + (__org.name | urlencode) + '/admins/',
-                           host=controller_hostname, oauth_token=controller_oauthtoken, verify_ssl=controller_validate_certs) | length == 0 }}"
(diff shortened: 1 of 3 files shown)


roles/object_diff/tasks/organizations.yml
```diff
@@ -1,17 +1,63 @@
...
+        __bool_empty_job_templates: "{{ query(controller_api_plugin, 'api/' + controller_api_version + '/organizations/' + (__org.name | urlencode) + '/job_templates/',
+                                  host=controller_hostname, oauth_token=controller_oauthtoken, verify_ssl=controller_validate_certs) | length == 0 }}"
+        __bool_empty_workflow_job_templates: "{{ query(controller_api_plugin, 'api/' + controller_api_version + '/organizations/' + (__org.name | urlencode) + '/workflow_job_templates/',
+                                           host=controller_hostname, oauth_token=controller_oauthtoken, verify_ssl=controller_validate_certs) | length == 0 }}"
+      ansible.builtin.set_fact:
+        __list_empty_orgs: "{{ __list_empty_orgs | default([]) + [__org.name] }}"
+      loop: "{{ __organizations_difference }}"
+      loop_control:
+        loop_var: __org
+      when: protect_not_empty_orgs is defined and protect_not_empty_orgs and
```
ivarmu: if all the `when` conditions are related with an `and`, the following is preferred:
```
      when:
        - protect_not_empty_orgs is defined
        - protect_not_empty_orgs
        - __bool_empty_users
        - __bool_empty_admins
        - __bool_empty_inventories
        - __bool_empty_teams
        - __bool_empty_projects
        - __bool_empty_job_templates
        - __bool_empty_workflow_job_templates
```

ivarmu: The following condition:
```
      when: protect_not_empty_orgs is defined and protect_not_empty_orgs and
```
should be in the `when` section of a `block` to ensure no API calls will be made in case that variable is not defined or is defined to `false`.

ivarmu (changes requested):

ivarmu (approved):

Issue #407, opened 2022-11-04
Add newly added promptable on launch vars to job_launch for AAP 2.3


ffirg:
In AAP 2.3 we have added everything that we could as a prompt on launch option when running job templates.
This has added more variables as per https://github.com/ansible/awx/pull/12875
Can we get these added to job_launch/tasks/main.yml ?
Will need to same exercise for workflows as well I suspect.

Thanks!

ffirg:
I believe the JT vars list should be:

 execution_environment
 forks
 instance_groups
 job_slice_count

sean-m-sullivan:
Slightly Behind, but needs testing, this is the WIP to get this done.
#406

ffirg:
Nice! 

I added:

```
    execution_environment:    "{{ __job_launch_item.execution_environment | default(omit, true) }}"
    forks:                    "{{ __job_launch_item.forks | default(omit, true) }}"
    job_slice_count:          "{{ __job_launch_item.job_slice_count | default(omit, true) }}"
    instance_groups:          "{{ __job_launch_item.instance_groups | default(omit, true) }}"
```

forks and job_slice_count doesn't seem to be supported? The others worked.

Example config vars:
```
---
controller_launch_jobs:
  - name: Demo Job Template
    execution_environment: 1    # Minimal EE
    inventory: 1                # Demo Inventory
    instance_groups: [1,2]      # run on default, or then control plane

  - name: Demo Job Template
    execution_environment: 2    # Ansible Engine 2.9 EE
    inventory: 2                # SunCavanaugh Industries
    instance_groups: [4]        # newIG instance group
```

sean-m-sullivan:
Need to do launch jobs, I missed that one!

sean-m-sullivan:
need to remove the default value for job slice on launch jobs as it interferes with backwards compatibility.

sean-m-sullivan:
From the PR that was done, this is now complete, closing

PR #422, opened 2022-11-24
[#407] change default behavior of job_slice_count in job_launch

adonisgarciac:
# What does this PR do?

Change default behavior of job_slice_count. 

job_slice_count is not set by default because is only allowed when ask_job_slice_count_on_launch is true but by default it is false. 

I think it is better to set job_slice_count without default 1 and give the responsibility to user who has to know when set it up due to ask_job_slice_count_on_launch is true in the JT. Moreover, JT can have a default value which is 1 if user does not change it, so I think even although user will not send job_slice_count with ask_job_slice_count_on_launch=true, it will work.

Also it will work fine in AAP 2.2 and Tower.

https://github.com/redhat-cop/controller_configuration/issues/407


ivarmu (approved):
LGTM.

Let's @djdanielsson and @sean-m-sullivan to review and approve.

silvinux (approved):
LGTM.

Tompage1994 (approved):
Completely agree. We shoudln't be being opinionated here. The underlying module doesn't have a default so we should just be omitting this as we do with most other options.

PR #424, opened 2022-11-24
remove OBJECT DIFF from task name

adonisgarciac:
# What does this PR do?
remove OBJECT DIFF from task name because it does not have sense. When tasks are running we can see role name in the task itself.

# How should this be tested?


ivarmu (approved):
LGTM.

PR #425, opened 2022-11-28
Fix loop object diff organizations

silvinux:
# What does this PR do?

Fix loop object diff organizations. It was populated the variable as a dictionary when it should be a list. Also, ´controller_organizations´ variable came from file tree role

# How should this be tested?

Create organizations that are not defined as code and run objectdiff.


adonisgarciac (approved):
LGTM! Thanks @silvinux for the fixes!! You rock!!

djdanielsson (approved):
LGTM

PR #431, opened 2022-12-01
avoid duplicate load vars in filetree_read

adonisgarciac:
# What does this PR do?

Init controller variables to avoid repeated elements when filetree_read is called N times.


sean-m-sullivan:
Adding the two maintainers into this for their thoughts / approval, it looks good to me, but want them to verify it wouldn't break something.

ivarmu (approved):
It's great! LVGTM!

PR #432, opened 2022-12-02
add some new exports in filetree_create

adonisgarciac:
# What does this PR do?
Add content to export with filetree_create: labels, notifications, instance_groups and applications.

# How should this be tested?

Export objects with filetree_create and try to import with filtree_read and dispatch.

### DIFF
--- a/roles/filetree_create/templates/current_inventory_sources.j2
+++ b/roles/filetree_create/templates/current_inventory_sources.j2
@@ -21,4 +21,22 @@ controller_inventory_sources:
     credential: "{{ inventory_source.credential }}"
 {% endif %}
 {% endfor %}
+    notification_templates_error:
+{% for notification_error  in query(controller_api_plugin, inventory_source.related.notification_templates_error,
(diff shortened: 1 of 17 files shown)


roles/filetree_create/templates/current_inventory_sources.j2
```diff
@@ -20,5 +20,23 @@ controller_inventory_sources:
 {% if inventory_source.credential %}
     credential: "{{ inventory_source.credential }}"
 {% endif %}
+    notification_templates_error:
```
ivarmu: This can lead on a result like the following line:

```
notification_templates_error:
```

which should be, either absent, either like follows:
```
notification_templates_error: []
```

You can run the query at the playbook task (vars section) to have that information available for the j2, so you can write it down only if it'll be populated.

ivarmu (changes requested):
Repeat the commented at every similar process in the PR.

ivarmu (approved):
LGTM. Thanks a lot!!!

PR #434, opened 2022-12-05
Update to reusable workflows

sean-m-sullivan:
# What does this PR do?
Updates github workflows to be reusable.
Idea is to make these changes to all 4 infra collection repos, so that when we make a change, it is automatically reflected in all 4 repositories. Currently there is a Issue to create a repo to store it in redhat_cop, for now its the repo I setup.
Adds sanity tests
Adds tests before releases and revamps release and changelog workflows
Fixes errors in lookup plugin discovered by sanity tests.

# How should this be tested?
Automated tests
Results here
Release with attachments
https://github.com/sean-m-sullivan/controller_configuration/releases/tag/1.0.11
Release Workflow
https://github.com/sean-m-sullivan/controller_configuration/actions/runs/3617309300


.github/workflows/pre-commit.yml
```diff
@@ -1,25 +1,35 @@
 ---
 # This workflow action will run pre-commit, which will execute ansible and yaml linting
 # See .pre-commit-config.yaml for what hooks are executed
-name: Yaml and Ansible Lint
+name: Sanity and Linting
```
Tompage1994: Probably shouldn't go with 'sanity' and rather use 'confidence' or something like that here.

changelogs/fragments/workflow.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
```
Tompage1994: Probably don't need this . I think the changelog should be kept to functionality changes

.github/workflows/pre-commit.yml
```diff
@@ -1,25 +1,35 @@
...
+
+on:
+  pull_request:
+  push:
+  schedule:
+    - cron: "0 6 * * *"
 
 jobs:
   pre-commit:
+    uses: sean-m-sullivan/infra_git_ci_cd/.github/workflows/pre-commit.yml@main
```
Tompage1994, line 15: Not sure I like this being under your namespace. Any way we can get this under redhat-cop?

Tompage1994 (changes requested):

.github/workflows/pre-commit.yml
sean-m-sullivan: https://github.com/redhat-cop/org/issues/537
this is temporary until that is created

changelogs/fragments/workflow.yaml
sean-m-sullivan: I can remove that, it was put in mainly to test the changelog workflow changes.

.github/workflows/pre-commit.yml
sean-m-sullivan: Changed

Tompage1994 (approved):

djdanielsson (approved):
I need to checkout the action itself at some point but the changes here seem fine

PR #449, opened 2022-12-15
fix bug variable name obj diff orgs

adonisgarciac:
# What does this PR do?

Fix a bug with the variable org.name in the queries to get orgs empty list.

# How should this be tested?

Launch object_diff.

### DIFF
--- a/roles/object_diff/tasks/organizations.yml
+++ b/roles/object_diff/tasks/organizations.yml
@@ -53,7 +53,9 @@
       loop: "{{ __organizations_difference }}"
       loop_control:
         loop_var: __organizations_difference_item
-      when: __organizations_difference is defined and (protect_not_empty_orgs is not defined or not protect_not_empty_orgs or __organizations_difference_item.name in __list_empty_orgs | default([]))
+      when:
+        -  __organizations_difference is defined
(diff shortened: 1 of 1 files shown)


silvinux (approved):
LGTM!

roles/object_diff/tasks/organizations.yml
```diff
@@ -53,7 +53,7 @@
       loop: "{{ __organizations_difference }}"
       loop_control:
         loop_var: __organizations_difference_item
-      when: __organizations_difference is defined and (protect_not_empty_orgs is not defined or not protect_not_empty_orgs or __organizations_difference_item.name in __list_empty_orgs)
+      when: __organizations_difference is defined and (protect_not_empty_orgs is not defined or not protect_not_empty_orgs or __organizations_difference_item.name in __list_empty_orgs | default([]))
```
ivarmu: I'm trying to understand what is the purpose of that change... I don't see it

adonisgarciac: It is because you can have protect_not_empty_orgs =  true but any empty organization. Then, you will not have __list_empty_orgs defined by the previous task

ivarmu (approved):

PR #451, opened 2022-12-21
filetree_create: fix export objects to add extra variables

adonisgarciac:
# What does this PR do?

Add or fix variables and extra_vars to some kind of objects like: inventory, inventory_source, hosts, groups, jt, wfjt and notifications

# How should this be tested?

Launch filetree_create to export all objects.


ivarmu (approved):
LGTM

PR #452, opened 2022-12-22
relocate team and user roles when export with filetree_create

adonisgarciac:
# What does this PR do?

relocate team and user roles when export with filetree_create

# How should this be tested?

export roles of teams and users with filetree_create role.


ivarmu (approved):
LGTM!

PR #454, opened 2022-12-26
added roles to object_diff

ivarmu:
# What does this PR do?

Adds the object type `roles` to the object_diff stuff. This way, the 'desired state' idea can be implemented for `roles`.

# How should this be tested?

Tested manually

N/A

/N/A


sean-m-sullivan (approved):
LGTM, Needed to test thinking we might need to add tests for this in terms of the controller api diffs, but they passed for existing tests.

PR #460, opened 2023-01-06
Update README.md with correct badge

Tompage1994:
# What does this PR do?

When going through other repos and fixing links I realised this badge was from the old CI jobs. I've now updated to the correct badge for the current CI

# How should this be tested?

Should mark as passing when CI passes

N/A

N/A


branic (approved):
LGTM!

PR #462, opened 2023-01-10
Replaced the 'warning' call for the 'v' call to show the information only when verbosity is required

### LINKED ISSUE #384: Unexpected warning in diff plugin
A customer has flagged up that there is a warning which looks like it's just there for info. This should probably be adjusted to  be displayed on verbose or something. The link to the line is below.

https://github.com/redhat-cop/controller_configuration/blob/05ee872f7fdcc1f471aab58dcf4765c157ec26a2/plugins/lookup/controller_object_diff.py#L183

ivarmu:
# What does this PR do?

Replaced the 'warning' call for the 'v' call to show the information only when verbosity is required.

# How should this be tested?

Tested manually.

resolves #384 

N/A

### DIFF
--- a/plugins/lookup/controller_object_diff.py
+++ b/plugins/lookup/controller_object_diff.py
@@ -248,8 +248,8 @@ class LookupModule(LookupBase):
                 item.update({"organization": item["summary_fields"]["organization"]["name"]})
                 item.pop("summary_fields")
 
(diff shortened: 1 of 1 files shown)


Tompage1994 (approved):
This seems sensible

PR #463, opened 2023-01-10
fixes a reserved name collision

### LINKED ISSUE #385: Project_update Role Issue
Hi Team,

I just found one glitch in project_update role. In project_update role below condition is not working properly. Seems item.**update** is a built in function or method that why its not taking this variable as Boolean type. Hope if we change the variable name means it will work as expected.
  when:
    - controller_projects is defined
    **- __project_update_update_item.update**
    - __project_update_update_item.state | default('present') != "absent"
    -
https://github.com/redhat-cop/controller_configuration/blob/devel/roles/project_update/tasks/main.yml
    
**__project_update_update_item.update** this is current 
**__project_update_update_item._update** if we change like this means the condition will work 

Also we need to change the "update" variable in project role. Kindly correct me if am wrong means

ivarmu:
# What does this PR do?

Renamed the field `update` to `do_update` to avoid colliding with the Python dict update method

# How should this be tested?

Manually tested the problem itself, but not the complete functionality. The change seems quite simple.

resolves #385 

N/A

### DIFF
--- a/changelogs/fragments/project_updates.yml
+++ b/changelogs/fragments/project_updates.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - Renamed the field `update` to `do_update` to avoid colliding with the Python dict update method
+...
--- a/changelogs/fragments/project_updates.yml
+++ b/changelogs/fragments/project_updates.yml
@@ -1,4 +1,4 @@
 ---
 minor_changes:
-  - Renamed the field `update` to `do_update` to avoid colliding with the Python dict update method
+  - Renamed the field `update` to `update_project` to avoid colliding with the Python dict update method
 ...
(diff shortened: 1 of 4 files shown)


sean-m-sullivan:
So generally I am in faver of this change.
however I am looking at other workings
Should it be project_update to match the project module option
https://github.com/ansible/awx/blob/devel/awx_collection/plugins/modules/project.py#L157

which updates the project after a project role, should it match or be its own option like this update would make it.

Thoughts?

ivarmu:
We can use `update_project` as is defined in the file you mention, I'm pushing the changes tomorrow morning 😴

sean-m-sullivan (approved):
LGTM

PR #464, opened 2023-01-12
renamed all controller_workflow_job_templates to controller_workflows

### LINKED ISSUE #439: Incorrect variable name for workflow objects in several roles
Both "filetree_read" and "object_diff" roles use the following default var for workflow objects:

```yaml
controller_workflow_job_templates: []
```

this needs to change to:

```yaml
controller_workflows: []
```

otherwise the following task will fail due to undefined var if no workflow is defined:

```yaml
- name: "Find the difference of Workflow Job Templates between what is on the Controller versus CasC on SCM"
  ansible.builtin.set_fact:
    __workflow_job_templates_difference: "{{ lookup('redhat_cop.controller_configuration.controller_object_diff',
                                                    api_list=__controller_api_workflow_job_templates,
                                                    compare_list=controller_workflows,
                                                    with_present=false, set_absent=true)
                                          }}"
```

ivarmu:
# What does this PR do?

Renamed all controller_workflow_job_templates to controller_workflows.

# How should this be tested?

Tested manually. Automated tests are OK also.

resolves #439

N/A

### DIFF
--- a/changelogs/fragments/filetree_create.yml
+++ b/changelogs/fragments/filetree_create.yml
@@ -1,5 +1,6 @@
 ---
 minor_changes:
+  - Renamed variable controller_workflow_job_templates to controller_workflows (the previos one was not used at all).
   - Improve template to export settings with filetree_create role. Settings will be in yaml format.
   - Add or fix some variables or extra_vars exported from objects like notifications, inventory, inventory_source, hosts, groups, jt or wjt.
 ...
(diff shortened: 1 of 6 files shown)


Tompage1994 (changes requested):
LGTM, but can you add a changelog entry please

Tompage1994 (approved):

PR #469, opened 2023-01-17
Allow state to be undefined on workflow_job_templates when  using simplified_workflow_nodes

### LINKED ISSUE #297: workflow error
I am trying to figure out if there is a bug in the role or if I have something wrong in the vars.

```
---
controller_workflows:
  - name: test_workflow
    description: testing issue
    extra_vars: ''
    survey_enabled: false
    allow_simultaneous: true
    ask_variables_on_launch: false
    ask_inventory_on_launch: false
    ask_scm_branch_on_launch: false
    ask_limit_on_launch: true
    webhook_service: ''
    organization: Default
    schedules: []
    simplified_workflow_nodes:
      - all_parents_must_converge: false
        identifier: node001
        unified_job_template: first_template
        credentials: []
        success_nodes: []
        failure_nodes:
          - node002
        always_nodes: []
      - all_parents_must_converge: false
        identifier: node002
        unified_job_template: second_template
        credentials: []
        success_nodes: []
        failure_nodes: []
        always_nodes: []
    notification_templates_started: []
    notification_templates_success: []
    notification_templates_error: []
    notification_templates_approvals: []
    survey_spec: {}
...
```

I get this error
```
FAILED! => {"msg": "The conditional check '__workflow_loop_item.state == \"present\"' failed. The error was: error while evaluating conditional (__workflow_loop_item.state == \"present\"): 'dict object' has no attribute 'state'\n\nThe error appears to be in '/home/user/.ansible/collections/ansible_collections/redhat_cop/controller_configuration/roles/workflow_job_templates/tasks/main.yml': line 62, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n# Create links between workflow node\n- name: loop over nodes in schema to add to workflow templates\n  ^ here\n"}
```

I did test removing this line https://github.com/redhat-cop/controller_configuration/blob/8028d7105a183da44915714754ab8f2ebe0d7995/roles/workflow_job_templates/tasks/main.yml#L69 and that works but I am not sure if that is the correct fix or if something else is wrong.

spyrexd:
# What does this PR do?

The `state` key is optional for workflow_job_templates unless you are using simplified_workflow_nodes.
This change allows the state key to remain optional when using simplified_workflow_nodes.

# How should this be tested?

Automated tests have been added

resolves #297 

### DIFF
--- a/roles/workflow_job_templates/tasks/main.yml
+++ b/roles/workflow_job_templates/tasks/main.yml
@@ -75,5 +75,5 @@
     loop_var: __workflow_loop_item
   when:
     - __workflow_loop_item.simplified_workflow_nodes is defined
-    - __workflow_loop_item.state == "present"
+    - (__workflow_loop_item.state | default('present')) ==  "present"
 ...
--- a/.github/tests/configs/differential_items.yml
+++ b/.github/tests/configs/differential_items.yml
@@ -138,4 +138,7 @@ differential_items:
       - name: Simple workflow schema2
         organization: Default
         state: absent
+      - name: Simple workflow schema no state defined
+        organization: Default
+        state: absent
 ...
--- a/.github/tests/configs/workflows.yml
+++ b/.github/tests/configs/workflows.yml
@@ -73,4 +73,47 @@ controller_workflows:
     notification_templates_success: []
     notification_templates_error: []
     notification_templates_approvals: []
+  - name: Simple workflow schema no state defined
+    description: a basic workflow
+    extra_vars: {}
+    survey_enabled: false
+    allow_simultaneous: false
+    ask_variables_on_launch: false
+    inventory:
+    limit:
+    job_tags:
+      - stuff
+      - stuff2
+    skip_tags:
+      - stuff3
+    ask_labels_on_launch: true
+    ask_skip_tags_on_launch: true
+    labels:
+      - Prod
+    scm_branch:
+    ask_inventory_on_launch: false
+    ask_scm_branch_on_launch: false
+    ask_limit_on_launch: false
+    organization: Default
+    schedules: []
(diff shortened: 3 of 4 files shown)


silvinux (commented):
LGTM!

roles/workflow_job_templates/tasks/main.yml
```diff
@@ -75,5 +75,5 @@
     loop_var: __workflow_loop_item
   when:
     - __workflow_loop_item.simplified_workflow_nodes is defined
-    - __workflow_loop_item.state == "present"
+    - __workflow_loop_item.state is undefined or __workflow_loop_item.state == "present"
```
Tompage1994: I think this isn't the clearest like this. Would it be better to set this as 
```
__workflow_loop_item.state != 'absent'
```

Tompage1994 (changes requested):

roles/workflow_job_templates/tasks/main.yml
adonisgarciac: I think it will not work because state is not defined by default in objects within configuration as code. State = present is  the default value, this PR try to remove the dependence of declare the state of workflows. With your solution, it will fail with an undefined variable.

Tompage1994: In that case you could do:

```
- (__workflow_loop_item.state | default('present')) == "present"
```

Just thinking it otherwise isn't entirely clear

spyrexd: Pushed changes to the way the default value is evaluated.

sean-m-sullivan (approved):
LGTM, good change and catch.

Tompage1994 (approved):

PR #476, opened 2023-01-24
Fix controller password query

adonisgarciac:
# What does this PR do?

Fix one query with controller_password to change it and set oauth_token=controller_oauthtoken.


silvinux (approved):
LGTM!

PR #478, opened 2023-01-26
add no_log everywhere controller_api_plugin is used

### LINKED ISSUE #446: filetree_read and object_diff roles output is exposing credentials in clear text
We need to add the secure logging variables to both filetree_read and object_diff roles to avoid having credentials in clear text in console output.

ivarmu:
# What does this PR do?

Add no_log everywhere controller_api_plugin is used to avoid to expose sensitive information in case of crashes.

# How should this be tested?

Tested manually.

resolves #446 

N/A


silvinux (approved):
LGTM!. This change will avoid that the password is not stored in clear in a stdout.

Tompage1994 (changes requested):
lgtm, but please add a changelog fragment

Tompage1994 (approved):
Thanks

PR #479, opened 2023-01-26
Added task to simplify the assignment of  galaxy/ee to org

### LINKED ISSUE #477: Change dispatch to run orgs twice for after creds/ee creation
# Summary
Change dispatch to run orgs twice for after creds/ee creation

# Issue Type
- feature update

silvinux:
# What does this PR do?

Solve the chicken and egg problem to add galaxy credentials and execution environments to organizations in the dispatch role. Basically, through variables we can create the organizations and credentials in the first run, then a tasks has been include to add the  galaxy credentials and execution environments to organizations.

# How should this be tested?

Tested manually.

resolves #477

### DIFF
--- a/changelogs/fragments/dispatch_creds_ee_orgs_creation.yml
+++ b/changelogs/fragments/dispatch_creds_ee_orgs_creation.yml
@@ -1,7 +1,6 @@
 ---
 minor_changes:
-  - Set the following variables to false when run all roles at dispatch role:
-      - assign_galaxy_credentials_to_org: false
-      - assign_default_ee_to_org: false
-  - Add task to add Galaxy credentials and Execution Environments to Organization
+  - Set the variables to assign_galaxy_credentials_to_org and assign_default_ee_to_org to false in the task to run all roles at dispatch role.
+  - Add task to add Galaxy credentials and Execution Environments to Organization.
+  - Adapt filetree_read role tests playbook config-controller-filetree.yml.
 ...
--- a/changelogs/fragments/dispatch_creds_ee_orgs_creation.yml
+++ b/changelogs/fragments/dispatch_creds_ee_orgs_creation.yml
@@ -1,0 +1,7 @@
+---
+minor_changes:
+  - Set the following variables to false when run all roles at dispatch role:
+      - assign_galaxy_credentials_to_org: false
+      - assign_default_ee_to_org: false
+  - Add task to add Galaxy credentials and Execution Environments to Organization
+...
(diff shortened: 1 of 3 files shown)


sean-m-sullivan:
I don't like it, but it should work, trying to get something else working, so going to put this one on hold till I figure it out
https://github.com/redhat-cop/controller_configuration/compare/devel...sean-m-sullivan:controller_configuration:dispatch_update?expand=1

We should add tests around this like I've illustrated in the above PR.

sean-m-sullivan:
when I add in the tests in that PR preview I linked above, I get failures, even with yours, its still trying to lookup the EE/Creds, will look at it, but we need to add this situation to tests so we get good results on it.

silvinux:
Ok thanks, I will look it deeper. Cheers.

silvinux:
Hello Sean,

I've done some tests just adding the variables in the dictionary as you mentioned, but in this approach, these variables being inside the dictionary, it is necessary to iterate them to apply them only when the dictionary item contains the role organizations, complicating the role. 

``` 
  1 ---
  2 controller_configuration_dispatcher_roles:
  3   - {role: settings, var: controller_settings, tags: settings}
  4   - {role: organizations, var: controller_organizations, tags: organizations, assign_galaxy_credentials_to_org: false, assign_default_ee_to_org: false}
...
  9   - {role: credentials, var: controller_credentials, tags: credentials}
 10   - {role: credential_input_sources, var: controller_credential_input_sources, tags: credential_input_sources}
 11   - {role: notification_templates, var: controller_notifications, tags: notification_templates}
 12   - {role: projects, var: controller_projects, tags: projects}
 13   - {role: inventories, var: controller_inventories, tags: inventories}
...
 26   - {role: roles, var: controller_roles, tags: roles}
 27   - {role: organizations, var: controller_organizations, tags: organizations}
 28 ...
 ``` 

 Something like this:
 ``` 
   tasks:
    - debug:
        msg:
          - "{{ test_var }}"
      loop: "{{ controller_configuration_dispatcher_roles }}"
      loop_control:
        loop_var: __role
      vars:
        test_var: "{{ __role.assign_to_org if (__role.assign_to_org is defined) else omit }}"
``` 

It is somewhat similar to what I have done in PR, however I believe that by trying to simplify this process we are forcing users to a solution, moving away from the agnostic solution that roles are looking for. I think that the solution we have now, from outside the role, and applying those variables from the playbook is a more agnostic solution, and gives the user flexibility when using the roles.

sean-m-sullivan (approved):

sean-m-sullivan (approved):
Okay merging this in as its the best option for fixing it, I've tried the complicated other options and none have worked.

PR #481, opened 2023-01-30
add new object types to object_diff role

adonisgarciac:
# What does this PR do?

Add new object types to object_diff role: applications, execution environments, instance groups, notifications and schedules.

Particular cases: 
- Execution Environments: They will be removed /listed only EE of organization. So, EE which are globaly available (no org), they won't listed or removed with object_diff (disared state of CasC).
- Instance Groups:  They will be filtered by user_capabilities.delete: true to avoid that it try to remove some default instance group.
- Schedules: They doesn't have organization so, object_diff role for this kind of object will search for JT and WFJT of its organization and then, it keep schedules which are in some of the JT and WFTJ. 

# How should this be tested?

Try to add new objects to the controller and launch the object_diff role to check if applications, execution environments, instance groups, notifications and schedules appear now.

### DIFF
--- a/changelogs/fragments/object_diff_role_and_plugin.yml
+++ b/changelogs/fragments/object_diff_role_and_plugin.yml
@@ -1,0 +1,3 @@
+minor_changes:
+  - "Add new type of objects for object_diff role:  applications, execution environments, instance groups, notifications and schedules"
+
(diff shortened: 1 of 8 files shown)


Tompage1994 (approved):
LGTM

Tompage1994 (changes requested):
Just spotted there are a bunch of linter errors you'll need to fix.

Also can you please add a changelog for this.

silvinux (commented):
LGTM! GOOD job!

Tompage1994 (approved):
LGTM

PR #491, opened 2023-02-07
#482 change behavior of labels when parameter is empty

adonisgarciac:
# What does this PR do?

Change the behavior of labels when parameter is empty. Labels can be an empty array because it is wanted to remove the labels of a previous labeled JT/WFJT. But, as it was set until now, empty label was not sent.

# How should this be tested?

Create an object with labels and after that change to remove the labels.

https://github.com/redhat-cop/controller_configuration/discussions/482


djdanielsson:
So we had some failures but I think they aren't related.... so I am tempted to approve... I will approve but wait for another person to review before merging.

djdanielsson (approved):

sean-m-sullivan:
I was more thinking we'd default to empty list, either way I want to test this manually to see how it shows up in things, especially without automated tests. But I have haven't had the time, will get to this if I can

sean-m-sullivan (approved):
Double checked behavior, didn't show up in change, nor did change existing labels. Approved.

PR #500, opened 2023-02-14
#496 fix to omit labels if it is not defined

### LINKED ISSUE #496: How can related labels be defined as code for JT/WFJT/WFNodes
https://github.com/redhat-cop/controller_configuration/blob/devel/roles/workflow_job_templates/tasks/add_workflows_schema.yml#L23

Why we are allowing something like __workflow_loop_node_item.related.labels? How can a JT/WF/WFNode object be defined as code with related labels?

It is getting an error because it has default([]) so it will always send a value even it is not defined and for versions previous of AAP 2.3, labes are not in workflow_job_template_node.

I would change these lines:

Before:
```
    labels:                               "{{ __controller_template_item.labels | default(__controller_template_item.related.labels | default([]) | map(attribute='name') | list) | default(omit) }}"
```

After:
```
    labels:                               "{{ __controller_template_item.labels | default(omit) }}"
```

adonisgarciac:
# What does this PR do?

As labels are defined, they always will send something even though they are not defined as code.  It is due to default([]) defined for "related.labels". With this change, labels will be omited if they are not defined as code.

# How should this be tested?

Create JT, WFJT or Schedules as code with defined labels and after that, empty the labels defining it:

labels: []

resolves #496

### DIFF
--- a/roles/job_templates/tasks/main.yml
+++ b/roles/job_templates/tasks/main.yml
@@ -52,7 +52,7 @@
     webhook_service:                      "{{ __controller_template_item.webhook_service | default(omit, true) }}"
     webhook_credential:                   "{{ __controller_template_item.webhook_credential | default(omit, true) }}"
     scm_branch:                           "{{ __controller_template_item.scm_branch | default(omit, true) }}"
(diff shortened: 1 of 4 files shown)


silvinux (commented):
LGTM! Nice and easy solution.

ivarmu (approved):
LGTM

PR #503, opened 2023-02-16
Fix linting issues and set to warn on broken lint rule

Tompage1994:
# What does this PR do?
Fixes lint issues caused by update to ansible-lint 6.13.0

# How should this be tested?
CI should pass

N/A

Note that I've set jinja[invalid] to warn for now because the new version has caused a regression in this rule which I've raised a bug for: https://github.com/ansible/ansible-lint/issues/3048


branic (approved):
LGTM

PR #504, opened 2023-02-16
avoid to create orgs during drop_diff

adonisgarciac:
# What does this PR do?

Due to conditional "when: __controller_organizations is defined", if there are not any org to remove, it was sent the controller_organizations variable read from filetree_read. Then, organizations will try to be created during drop_diff process. 

With this change, if any org has to be removed, an empty list will be sent for controller_organizations.


sean-m-sullivan (changes requested):
Looks like a good change to me, But we need to put in a changelog fragment for this as the behavior will change.

silvinux (approved):
LGTM!

sean-m-sullivan (approved):

PR #508, opened 2023-02-20
Fix linting issues and set to warn on broken lint rule

Tompage1994:
# What does this PR do?
Fix linting issues and set to warn on broken lint rule

# How should this be tested?
CI

resolves #[number]

https://github.com/redhat-cop/controller_configuration/pull/503 made the skip global, this makes it so that the rule is only skipped for certain tasks

As per comment https://github.com/redhat-cop/aap_utilities/pull/136#issuecomment-1433378274


djdanielsson (approved):
LGTM

branic (approved):
LGTM

PR #511, opened 2023-02-21
Add argument specs for roles, remove master template role and update readmes

Tompage1994:
# What does this PR do?
There's a lot to take in here and it took AGES to put together but I think its all correct. Any typos or minor mistakes can be cleaned up in future PRs

# How should this be tested?
Reading it... (sorry)

resolves #509 

We will need to do the same on the other collections too 😢


Tompage1994:
Realised I missed the filetree and dispatch stuff. Was meant to come back to it. I'll do those ones in a separate PR tomorrow

sean-m-sullivan (approved):
This looks good, trying to get that last test to pass, this is awesome work Tom.

PR #525, opened 2023-03-07
Avoid double pipeline on PRs

ericzolf:
# What does this PR do?

I got annoyed by unstable pipelines running twice on each PR. The changes should make sure that the pipeline only triggers on very specific pushes. The devel branch is anyway protected so that it can only be modified by pull requests

# How should this be tested?

Create a PR and enjoy to have e.g. pre-commit running only once. You can also create a branch (or a tag) ending in underscore to see how the pipeline triggers.

n/a

n/a

### DIFF
--- a/.github/workflows/ci_testing.yaml
+++ b/.github/workflows/ci_testing.yaml
@@ -1,6 +1,14 @@
 ---
 name: Run Test Playbooks on AWX workflow
-on: [push, pull_request]
+on:
+  push:
+    branches: ['*_']
+    tags:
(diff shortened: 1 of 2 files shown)


Tompage1994 (changes requested):
The CI should probably also run on the main branch (i.e. once its been merged) so that everything can be confirmed as working.

It can probably also run on all tags as we don't push tags except releases anyway and it not causing a double run if we do it on any tags anyway

sean-m-sullivan (approved):
Nice Addition, THanks!.

PR #528, opened 2023-03-08
Update pre-commit.yml

Tompage1994:
# What does this PR do?
Adds on to previous PR where double CI runs were prevented. I want to add this so it still runs on the devel branch when pushed (i.e. a PR is merged) as then the commit is verified for us and the state of the repo can be said to be good.

# How should this be tested?
runs when PR merged

No

#525 related


sean-m-sullivan (approved):
Good change

PR #534, opened 2023-03-12
update testing to use docker compose and versions

sean-m-sullivan:
# What does this PR do?
update testing to use docker compose and versions, Right now there are problems with devel, and that will be added later.
Right now the current test runs against the latest release which is 21.13.0, which this test also does.

# How should this be tested?
With the automated tests


djdanielsson (approved):
LGTM

PR #535, opened 2023-03-13
Cleanup and add devel test

sean-m-sullivan:
# What does this PR do?
Adds back in devel test, A PR was pushed last night that fixed the issue in the awx repo. Removed some things that I had added that turned out not to be needed.


djdanielsson (approved):
LGTM

PR #537, opened 2023-03-13
Infra update 

sean-m-sullivan:
# What does this PR do?
Updates references from redhat_cop to infra, mainly the readmes

In draft as I figure in how to unbreak the breaking change.

# How should this be tested?

resolves #[number]


djdanielsson (approved):
I am good with this if you want to do it now or wait until 2.4 when we officially drop redhat_cop namespace

silvinux (commented):
LGTM! Awesome. But I think we should wait until the 2.4 is released, basically there are a lot of customers/users that might don't know about the change.

sean-m-sullivan:
@silvinux we are looking at doing the change now so we can provide this in a release with the 2.4 bundle. So this is prepping for release.

silvinux:
> @silvinux we are looking at doing the change now so we can provide this in a release with the 2.4 bundle. So this is prepping for release.

It's Ok, I'll try to contact them.

PR #547, opened 2023-03-20
Fix dispatch to use general vars instead of specific hostvars

Klaas-:
# What does this PR do?
Change the when clause for variable lookup of the dispatch role to vars instead of hostvars

# How should this be tested?
This should not modify behavior except for also including vars that are for example set on the play level.

No


Tompage1994 (approved):
Perfect, I knew there were some issues with variable loading when not set against a host but I couldn't find a better way.

I never knew that there was a special variable called `vars`. Every day's a school day! Thanks

PR #549, opened 2023-03-22
Move CI test code to codebase rather than .github

Tompage1994:
# What does this PR do?
Since we changed the CI to run on 'pull_request_target', the CI only runs against the tests set on devel, but we should probably be able to update the tests and them be run. This change moves them into the tests directory which is probably more appropriate

# How should this be tested?
CI will run after its been merged (not ideal, i know but this should resolve that)

No

https://github.com/redhat-cop/ah_configuration/pull/246


djdanielsson (approved):
LGTM

Tompage1994:
You'll have to ignore the markdown link check on this as the location has moved and technically doesn't exist at that URL right now but will once merged.

PR #553, opened 2023-03-24
Up delay on projects async to prevent failures

Tompage1994:
# What does this PR do?
Increases delay on the project creation async because it regularly fails there randomly (presumably because its slow to do the project pull)

# How should this be tested?
CI will be more stable

No
N/A


Tompage1994:
Lol. There is a failure in the markdown link checker because of some kind of DNS error. The link should work (I can navigate to it) just it seems theres some weirdness going on somewhere that we will probably need to ignore for now

djdanielsson (approved):
LGTM

Issue #548, opened 2023-03-20
project and project_update defaults for update_project


parrjd:
The default value used for update_project in both project and project_update breaks backward compatibility with older playbooks and vars used with controller_configuration.

currently both project, and project_update have the default for update_project set to default to false, which breaks older playbooks written around controller_configuration.  

Also having the update_project var default to false really does not make much sense, since the name of the role is project_update, and one would expect it to update the projects, but if the var update_project is not explicitly included in the project definition it will skip since update_project defaults to false.

Recommend project default for update_project remain false, and change the default for project_update to default to true.  This would preserve behavior with running older playbooks that have not been updated to support the changes as projects would be for loading new/updating projects, and a call to project_update would  succeed since it would default update_project to true causing the update to happen when calling the project_update role

sean-m-sullivan:
Can you point to use where it breaks, and with what versions so we can better understand the problem?

sean-m-sullivan:
Idea is to set the default in project update role to True
https://github.com/redhat-cop/controller_configuration/blob/3839b01d32f9d792159a12c8ed13f54bbe49b093/roles/project_update/tasks/main.yml#L24

We should also then remove project update in dispatch to stop it from running twice when its set to true manually.

djdanielsson:
I don't understand how this is breaking code, as for changing default to true vs false... after some thought I don't care to much either way. but I would like to understand how this breaks code and if so how far back are we talking?

parrjd:
With the merge of "fixes a reserved name collision #463" a new when statement was added fixing the name collision, but that fix changed the behavior of project_updates since before the fix not having update_project defined allowed the project to be updated.  



The example code include will not result in the project update being executed as the update_project is listed as an optional value, but it defaults to false which will block the project update as expected.  

controller_projects:
  - name: Test Project
    scm_type: git
    scm_url: https://github.com/ansible/tower-example.git
    scm_branch: master
    scm_clean: true
    description: Test Project 1
    organization: Satellite
    wait: true

sean-m-sullivan:
@parrjd notice on the PR to fix.

parrjd:
That does not look to address the Isse since the default is still false for project_update

sean-m-sullivan:
@parrjd you are right, I've fixed it 3 minutes after I posted ><

sean-m-sullivan:
as the above has been merged, closing issue.

PR #560, opened 2023-04-04
remove project update from dispatch

sean-m-sullivan:
# What does this PR do?
After some testing and looking at the problem, this removes the project update role from dispatch, under normal conditions the project role will update the project. 
  - Removed project_update from dispatch. This is because with bringing update_project option in line with the module options, it was running twice both in project and project update. Since both roles use the same variable controller_projects.
  - If someone wants to have the old behavior, or only update projects with dispatch, the dispatch variable controller_configuration_dispatcher_roles can be overwritten and customized.

Also redid the ics file to use LF line formating instead of CRLF to pass sanity

#548


ivarmu (approved):
LGTM

PR #562, opened 2023-04-04
Solve issue `Object diff for roles removing most roles`

### LINKED ISSUE #552: Object diff for roles removing most roles
# Summary

There are several things not working with object diff for roles. In our case, it marks most of the roles for removal.
From what I can tell, it will only work for target_teams, job_templates and workflows, as there are the only ones expanded here:
https://github.com/redhat-cop/controller_configuration/blob/devel/plugins/lookup/controller_object_diff.py#L221-L243

This means it will not work for projects, inventories, credentials.

Furthermore, all keys of the compared objects must match: https://github.com/redhat-cop/controller_configuration/blob/devel/plugins/lookup/controller_object_diff.py#L279
So even for job_templates, or target_teams, this will not work if we add "organization" or "lookup_organization", as these are not returned by the api in the role object - and the keys will not match.

# Issue Type

- Bug Report

# Ansible, Collection, Controller details

redhat_cop.controller_configuration: 2.2.5

# STEPS TO REPRODUCE
Here are a couple of test scenarios, where the entries will be marked for removal (state: absent):

1. Role will be set to state: absent, because the target_team is not handled - workaround is to use target_teams list
```yaml
---
controller_roles:
  - user: myuser
    target_team: myteam
    role: member
```

2. Inventories not expanded
```yaml
---
controller_roles:
  - team: myteam
    inventories:
      - inv1
      - inv2
    role: use
```
3. Projects not expanded
```yaml
---
controller_roles:
  - team: myteam
    projects:
      - project1
      - project2
    role: use
```
4. Credentials not expanded
```yaml
---
controller_roles:
  - team: myteam
    credentials:
      - secret1
      - secret2
    role: use
```

4. lookup_organization is an extra key not handled by object diff
```yaml
---
controller_roles:
  - team: myteam
    lookup_organization: Acme
    job_templates:
      - jd1
      - jd2
    role: execute
```

4. organization is an extra key not handled by object diff
```yaml
---
controller_roles:
  - user: myuser
    organization: Acme
    target_teams:
      - teams1
    role: member
```

ivarmu:
# What does this PR do?

Tries to solve issue #552:

- Added treatment for:
  - target_team
  - job_template
  - workflow
  - inventory
  - projects
  - credentials

# How should this be tested?

Manually tested. Automatic tests also passed.

resolves #552 

N/A


ivarmu:
Pending for some extra info asked at the issue #552 .

sean-m-sullivan:
So was this a bug where if we add spaces to that are that it removes. And is this a WIP as it says still not tested?

ivarmu:
The spaces additions/removals are basically sthetic and/or required by the lintering checkings.

I checked manually all the changes and also run locally all the tests in [`tests/configure_controller.yml`](https://github.com/redhat-cop/controller_configuration/blob/devel/tests/configure_controller.yml). I've updated the description acordingly.

I've also improved the `*_expanded` staff in the lookup plugin and also added new fields in the roles comparisons.

sean-m-sullivan (approved):
LGTM, Needed time to review the python changes

PR #563, opened 2023-04-04
An inventory source is not ever assigned to an organization

ivarmu:
# What does this PR do?

When exporting an inventory source, if it has not associated organization, the filetree_create role will fail as no organization field is found. This is fixing that error and setting a ToDo organization name instead.

# How should this be tested?

Automatic tests.

N/A

N/A

### DIFF
--- a/roles/filetree_create/templates/current_inventory_sources.j2
+++ b/roles/filetree_create/templates/current_inventory_sources.j2
@@ -3,7 +3,9 @@ controller_inventory_sources:
 {% for inventory_source in current_inventory_sources_asset_value %}
   - name: "{{ inventory_source.name }}"
     description: "{{ inventory_source.description }}"
-    organization: "{{ inventory_source.summary_fields.organization.name | default('ToDo: The Inventory Source \'' + inventory_source.name + '\' must belong to an organization') }}"
+{% if inventory_source.summary_fields.organization %}
+    organization: "{{ inventory_source.summary_fields.organization.name }}"
+{% endif %}
     source: "{{ inventory_source.source | default('ToDo: The source of the inventory_source was originally missing and must be specified',true) }}"
 {% if inventory_source.source_project %}
     source_project: "{{ inventory_source.summary_fields.source_project.name }}"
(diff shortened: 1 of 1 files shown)


djdanielsson:
I am not sure I like this...

ivarmu:
> I am not sure I like this...

Have in mind that the `filetree_create` is intended to export the current objects from an already configured instance of Tower/Controller, so it's out of our control the original state for that objects... even if the inventory_source has been defined as code and deployed, this field is non mandatory (see https://github.com/redhat-cop/controller_configuration/blob/9478b34cf212666dd9c5a324be97387da7ee5be5/roles/inventory_sources/tasks/main.yml#L8) so... maybe it should be better to drop the entire line instead of setting the ToDo organization?

sean-m-sullivan:
The ONLY use of organization in the inventory source is to lookup the parent inventory, So if the ORG does not exist, it should not matter, its if the parent inventory exists is the key question.

The actual use case if for cases where the user has access to multiple orgs where they have identical inventory names.

I am not sure what scenario this is trying to fix, where filetree create, creates an inventory source with no org, which should be fixed in create, or user error where they added it manuallly?

ivarmu:
This is trying to fix the error thrown by the filetree_create when is exporting an inventory_source that is not assigned to any organization. That scenario is currently breaking the execution and it shouldn't, as it is a real and feasible case.

adonisgarciac:
I'm not sure in what cases it's possible to have an inventory_source without organization. I mean, organization is mandatory for inventories and inventory is mandatory for inventory_sources. Then, inventory_source should always have inventory_source.summary_fields.organization.name. 

I don't understand why awx/controller module accepts organization for inventory_source https://docs.ansible.com/ansible/latest/collections/awx/awx/inventory_source_module.html#parameter-organization when if you set an organization different than inventory's organization, it will fail.

On the other hand, controller_object_diff is trying to compare organization but it could fail because organization is not mandatory for inventory_source:

https://github.com/redhat-cop/controller_configuration/blob/devel/plugins/lookup/controller_object_diff.py#L151-L153
https://github.com/redhat-cop/controller_configuration/blob/devel/plugins/lookup/controller_object_diff.py#L183-L186

sean-m-sullivan (approved):

sean-m-sullivan:
Until we move to export, the summary fields is the only way to look at this, it should not be possible to have an inventory without an org, so this should always work.

PR #565, opened 2023-04-07
Add option for default ee from org export

sean-m-sullivan:
# What does this PR do?
Add option for default ee from org export


roles/organizations/tasks/main.yml
```diff
@@ -6,7 +6,7 @@
     custom_virtualenv:                  "{{ __controller_organizations_item.custom_virtualenv | default(omit, true) }}"
     max_hosts:                          "{{ __controller_organizations_item.max_hosts | default(omit, true) }}"
     instance_groups:                    "{{ __controller_organizations_item.instance_groups | default(omit, true) }}"
-    default_environment:                "{{ (__controller_organizations_item.default_environment | default(omit)) if (assign_default_ee_to_org is defined and assign_default_ee_to_org) else omit }}"
+    default_environment:                "{{ (__controller_organizations_item.default_environment.name | default(__controller_organizations_item.execution_environment | default(omit, true))) if (assign_default_ee_to_org is defined and assign_default_ee_to_org) else omit }}"
```
djdanielsson: Do we want omit true?

sean-m-sullivan: I don't remember anymore, and I was just keeping the original

djdanielsson: I think remove true

djdanielsson (approved):
LGTM

Issue #495, opened 2023-02-09
disussion #482 - It should be reviewed every parameter of each role to configure the right default behavior


adonisgarciac:
# Summary

Following the disscusion #482, I want to bring it to an issue. 

The roles of this collection have omit or omit,true in their parameters and it is not the default behavior of that parameters. It doesn't follow the principle of configuration as code.

I mean, if you change in the GUI  of controller some parameter and you don't set this parameter as code, actually you won't have this object in configuration as code in full.

In my opinion, if you want to have your object(s) as code, you have to have every parameter as code. If you don't have some of them, you will have a default value even though you have already configured a value for it through the GUI 

This approach will do that this collection will be closer to a "desired state" behavior.

adonisgarciac:
NOTE: It could break if it is using an AAP without the latest version because it will pass always every parameter and the target AAP (without latest version) could not have some of them.

Maybe it isn't a problem if it is using the collection of awx/ansible.controller/ansible.tower which fit with the version of awx/tower/controller. I believe that as the awx/ansible collection don't have that parameter, it won't send to the AWX/tower/controller.

djdanielsson:
We will most likely keep omit for most things so I don't think it should be an issue but is worth noting

adonisgarciac:
Why do you prefer to keep omit for most things? Do you prefer that users could change parameters from the GUI and not change them with the Configuration as Code if they have not defined this parameter?

I'm wondering how we want to address this behavior. Setting a default value will be more disruptive but it will be closer to a desired state because you will have to have everything as code, otherwise, a default value will be sent to the API and your manual changes done through the GUI will be lost.

Sorry if this is not good to be an issue, we could continue in the discussion again.

Tompage1994:
The reason we went with `default(omit)` (forgetting the usage of `(omit,true)`) was because we wanted the module defaults to be used. The only sensible thing we can do otherwise is default to what the module defaults to but then if the module's default changes then we would need to reflect this (and its pretty hard to keep track of that). 

Until I've seen a proposed example solution to this I'd be hesitant to go down this path.

adonisgarciac:
> (and its pretty hard to keep track of that).

Yes, this is the most problematic thing. 

> The reason we went with `default(omit)` (forgetting the usage of `(omit,true)`) was because we wanted the module defaults to be used

Ok, but now if someone change manually a parameter through the GUI, it will persist if you have not configured as code.

[...]

PR #567, opened 2023-04-08
Add option to roles to enforce defaults

sean-m-sullivan:
# What does this PR do?
As per the discussion an option has been added to multiple roles to enforce defaults. 
This option enforces module/api defaults in order to prevent config drift. This makes it so if an option is NOT specified in a configuration it enforces the default value. It is not enabled by default.

I have started with two roles that have multiple options enabled for review. The default environment is not set in orgs as there is an underlying issue with the module and the API where the default value is null, and its hard to pass that through the role->module->api.

I did try to do inline if/else like we do galaxy creds and it did not work
I tried to make my own default filter, but it does not allow for a variable to not be defined. 
In the end I settled on a simple lookup plugin to perform the if/else logic and thus provide the default with one value to then work on.

# How should this be tested?
This has been turned on in automated tests

#495


changelogs/fragments/enforce_defaults.yml
```diff
@@ -0,0 +1,4 @@
+---
+minor_changes:
```
djdanielsson, line 2: I wouldn't call this a minor change I think

roles/applications/tasks/main.yml
```diff
@@ -5,11 +5,11 @@
     name:                           "{{ __application_item.name | mandatory }}"
     new_name:                       "{{ __application_item.new_name | default(omit, true) }}"
     organization:                   "{{ __application_item.organization | mandatory }}"
-    description:                    "{{ __application_item.description | default(omit, true) }}"
+    description:                    "{{ __application_item.description | default(lookup('infra.controller_configuration.enforce_defaults', enforce_default=controller_configuration_applications_enforce_defaults , default_value='', omit_value=omit), true) }}"
```
djdanielsson, line 8: Do we want that final true? (And all the others)

djdanielsson (changes requested):
Questions

roles/applications/tasks/main.yml
sean-m-sullivan: The final true is the same as it was before the 
| default(omit, true)
its now
| default(lookup, true)

So you want to keep it, as it gives a null value rather then an _omit_placeholder_234324732472342348

changelogs/fragments/enforce_defaults.yml
sean-m-sullivan: Its a feature addition that isn't a breaking change, 
https://github.com/ansible-community/antsibull-changelog/blob/main/docs/changelogs.rst#changelog-fragment-categories

minor_changes
    This category should mention all new features, like plugin or module options. This section should not appear in patch releases according to semantic versioning.
    
It's a new feature added, an option.

sean-m-sullivan: If I were to have made it NON optional, enforced by default, I'd say major breaking change.

roles/applications/tasks/main.yml
djdanielsson: I thought the decision was to remove the true to all of them though?

sean-m-sullivan: I've been using this as a reference
https://github.com/redhat-cop/controller_configuration/issues/132
The default omit true or false is a weird beast, if you remove the true, then it sometimes sends strings "" to the option, which with defaulted values is what we wanted, but with True it sets null, which has been the past behavior, I believe if we did remove true, it would be a breaking change at this point. 

The previous conversation in the discussion linked in the PR, was more about how to go about doing the default values or the "Enforced" state, which we've bypassed the need to remove the true, as we have it the OLD way AND the Enforced way using this plugin.

djdanielsson (approved):
Can't wait to try it out

PR #572, opened 2023-04-17
Default options

sean-m-sullivan:
# What does this PR do?
Updated a set of roles to enforce defaults, this should be the last one of the trio to bring it to every role, I've tested every option for this.

# How should this be tested?

#495 

resolves #[number]


Tompage1994 (approved):
LGTM

djdanielsson (approved):
LGTM

PR #573, opened 2023-04-18
Update aap_config_as_code_public_meeting.ics

Tompage1994:
# What does this PR do?
Updates calendar invite
I've also generated it separately so it isn't linked to our internal RH calendar invite so it is more testable for us since it tries to break everything otherwise

# How should this be tested?
Download the ics and check it makes the correct event

N/A

Note that we have moved the time slot for the community meeting to 1300UTC every 4th Thursday


djdanielsson (approved):
LGTM

PR #578, opened 2023-04-20
Update galaxy.yml.j2

aoyawale:
updating names.

Just updating name.


sean-m-sullivan (approved):
LGTM, happy to update

Issue #332, opened 2022-07-29
Compatibility with awx.awx.export / awxkit


luckylittle:
Please make this role compatible with `awx.awx.export` / `awxkit`. It does not work out of the box:

Steps to reproduce:

1. Run the export playbook:

```yaml
---
- hosts: localhost
  tasks:
  - name: Export all assets
    awx.awx.export:
      organizations: 'all'
    register: orgs_register

  - name: Store organizations as YAML in ./configs
    ansible.builtin.copy:
      content: "{{ orgs_register.assets | to_yaml }}"
      dest: "/home/redhat/Projects/awxkit/configs/organizations.yml"
```

2. It exports the following example file:

```yaml
organizations:
- default_environment: null
  description: This is one Business Unit in DEV
  max_hosts: 0
  name: NAB_BU_1_dev
  natural_key: {name: NAB_BU_1_dev, type: organization}
  related:
    execution_environments: []
    galaxy_credentials:
    - credential_type: {kind: galaxy, name: Ansible Galaxy/Automation Hub API Token,
        type: credential_type}
      name: Ansible Galaxy
      organization: null
      type: credential
    notification_templates: []
    notification_templates_approvals: []
    notification_templates_error: []
    notification_templates_started: []
    notification_templates_success: []
- default_environment: null
  description: Another business unit, but in PROD
  max_hosts: 0
  name: NAB_BU_2_prod
  natural_key: {name: NAB_BU_2_prod, type: organization}
  related:
    execution_environments: []
    galaxy_credentials:
    - credential_type: {kind: galaxy, name: Ansible Galaxy/Automation Hub API Token,
        type: credential_type}
      name: Ansible Galaxy
      organization: null
      type: credential
    notification_templates: []
    notification_templates_approvals: []
    notification_templates_error: []
    notification_templates_started: []
    notification_templates_success: []
```

3. Remove these two organizations from AAP

4. Run the playbook to configure AAP using the export from step 1:

```yaml
---
- name: Playbook to configure Ansible Controller
  hosts: localhost
  connection: local
  collections:
    - redhat_cop.controller_configuration
  pre_tasks:

    - name: Include vars from configs directory
      include_vars:
        dir: "{{ controller_configs_dir | default((lookup('env','CONTROLLER_CONFIGS_DIR') == '') | ternary('./configs', lookup('env','CONTROLLER_CONFIGS_DIR'))) }}"
        ignore_files: [controller_config.yml.template]
[...]

PR #581, opened 2023-04-21
Add alias's for applicable roles to use the variables set by the awx cli export

sean-m-sullivan:
# What does this PR do?
Add alias's for applicable roles to use the variables set by the awx cli export.

# How should this be tested?
Converted the export configs to use this as they are what would be used.

#332 


djdanielsson (approved):
I'm sure a lot of people will like this update, looks good.

PR #583, opened 2023-04-24
Export differential module new and improved.

sean-m-sullivan:
# What does this PR do?
This PR Steals the awx.awx.export module, and parts of the lookup difference module and shoves them together.

It is built to be compatible with either the awx.awx or ansible.controller module. 

It only works for the fields that work with the awx cli export, but instead of reducing fields, and comparing direct lists, it does a search and pop on wanted fields and keeps all the fields in place. 

It's built to be robust and easy to add to. Though roles may be tricky still.

# How should this be tested?
Automated tests included in PR

redhat-cop/aap_configuration_extended#31 
In addition when
https://github.com/ansible/awx/issues/13868
Is acted on/improved, we can easily add fields.
Also teams on redhat-cop/infra.aap_configuration#468 

Is dependent on tooling update to pass sanity tests:
https://github.com/redhat-cop/ansible_collections_tooling/pull/22


ivarmu (approved):
LGTM. Initial testings works really well. Waiting on https://github.com/ansible/awx/issues/13868 to let `roles` to be more easily compared and treated!

djdanielsson (approved):
it would be nice if it didn't require the code to be in export format but it's a great start! hopefully we can get the export tool to clean up out it works

Issue #584, opened 2023-04-25
Mandatory rrule in shedules


jpcarmona:
# Summary

Hello!

Could it be that the "rrule" parameter in the "schedules" role does not have to be mandatory?
https://github.com/redhat-cop/controller_configuration/blob/devel/roles/schedules/tasks/main.yml#L7

In awx.awx and ansible.controller collections do not appear as required:
https://docs.ansible.com/ansible/latest/collections/awx/awx/schedule_module.html#parameter-rrule
https://console.redhat.com/ansible/automation-hub/repo/published/ansible/controller/content/module/schedule/?version=4.3.0

# Issue Type

- Bug Report

# Ansible, Collection, Controller details

```console
ansible.controller              4.3.0  
ansible.utils                   2.5.0  
awx.awx                         21.14.0 
community.general               4.4.0  
infra.controller_configuration  2.3.1_devel 
```

# ERROR

```console (error)

TASK [infra.controller_configuration.schedules : Add Controller Schedule] ************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"msg": "Mandatory variable 'rrule'  not defined."}

```

# STEPS TO REPRODUCE

```yaml (playbook/task)

- name: "Delete schedule"
  ansible.builtin.include_role:
    name: "schedules"
  vars:
    controller_schedules:
      - name: "schedule_to_delete"
        state: absent

```

Great job on this collection!
Best regards

djdanielsson:
you are correct it shouldn't be set to mandatory, we will try and get that updated in the next release

PR #586, opened 2023-04-27
Rrule fix

sean-m-sullivan:
# What does this PR do?

Fix the schedule rrule to not be mandatory.

As referenced here:
https://github.com/ansible/awx/blob/devel/awx_collection/plugins/modules/schedule.py#L24

Fixing #584


djdanielsson (approved):
LGTM

PR #590, opened 2023-04-28
Revised enforced defaults

sean-m-sullivan:
# What does this PR do?
This removes the lookup plugin from the enforced defaults instead relying on if else. This should simplify things and make it run faster. 

# How should this be tested?
Automated tests are designed around this change.

resolves #[number]


djdanielsson (approved):
LGTM

PR #592, opened 2023-04-29
fix ci

sean-m-sullivan:
# What does this PR do?
Tried multiple ways to fix this, but this was what I went with. user password is now being exported as a field and ""

# How should this be tested?

resolves #[number]


djdanielsson (approved):
Whatever works at this point lol

Issue #355, opened 2022-09-16
Update license role to use the new subscription module


sean-m-sullivan:
The current license role is a bit of a workaround for manifests, but we should use the new subscription module to make it easier to use. 

example playbook:
```
---
- name: Get subscription for controller
  hosts: localhost
  connection: local
  gather_facts: false
  collections:
    - awx.awx
  tasks:
    - name: Get subscriptions with a filter
      awx.awx.subscriptions:
        username: redhat_username
        password: password
        filters:
          product_name: Red Hat Ansible Automation Platform
          support_level: Self-Support
        controller_host: https://controller.node
        controller_username: admin
        controller_password: secret123
        validate_certs: false
      register: subscription

    - name: Attach to a pool
      awx.awx.license:
        pool_id: "{{ subscription.subscriptions[0].pool_id }}"
        controller_host: https://controller.node
        controller_username: admin
        controller_password: secret123
        validate_certs: false
...
```
 If a manifest file is provided and specified, use it, otherwise use the pool entry, the product name in the example should work for most people subscriptions as default

sean-m-sullivan:
Closing with the update

PR #593, opened 2023-04-29
update liscense role to use subscription module

sean-m-sullivan:
# What does this PR do?
update liscense role to use subscription module

# How should this be tested?

Since awx does not do liscense its impossible for automated tests, however this role has been tested using the example and the test role.

#355


djdanielsson (approved):
LGTM

Issue #594, opened 2023-05-04
The Roles role uses incorrect values when it starts after the Object_diff role


dmnagornyi:
# Summary

When you run the Roles role after the Object_diff role, all roles that should be removed fail with an error:

```
"value of role must be one of: admin, read, member, execute, adhoc, update, use, approval, auditor, project_admin, inventory_admin, credential_admin, workflow_admin, notification_admin, job_template_admin, execution_environment_admin, got: notification admin"
```

Same for "job template admin", "execution environment admin" and other roles with "_" sign in the name.

For some reason, all "_" characters are replaced with a space character for role values. There is no problem with the "member" role, since it only consists of one word.

# Issue Type

- Bug Report

# Ansible, Collection, Controller details

```console (paste below)
ansible --version
ansible [core 2.14.3]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.9/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.9.13 (main, Nov  9 2022, 13:16:24) [GCC 8.5.0 20210514 (Red Hat 8.5.0-15)] (/usr/bin/python3.9)
  jinja version = 3.1.2
  libyaml = True

ansible-galaxy collection list
# /root/.ansible/collections/ansible_collections
Collection                     Version
------------------------------ -------
ansible.controller             4.3.0
infra.controller_configuration 2.3.1

Controller version

```

- ansible installation method: OS package from AAP installer

# OS / ENVIRONMENT

Red Hat Enterprise Linux release 8.7 (Ootpa)

# Desired Behavior

Those roles that are missing in the code and were defined by the Object_diff role are removed by the Roles role after the list of missing objects is generated.

# Actual Behavior

Playbook crashed due to invalid value.

```console (error)
[...]

PR #595, opened 2023-05-04
fix issue #594 for roles with blank spaces in the name

adonisgarciac:
# What does this PR do?

Fix issue #594. Role name has blank spaces for role types:

"project_admin"
"inventory_admin"
"credential_admin"
"workflow_admin"
"notification_admin"
"job_template_admin"
"execution_environment_admin"

It can be checked going to the API and look for some role with one of this type.

This role name is setting as role in the plugin controller_object_diff.py:

`                item.update({"role": item["name"].lower()})
`

And this value is not valid so we need no change the blank spaces for underscores:

`                item.update({"role": item["name"].lower().replace(" ", "_")})
`

# How should this be tested?

```
>>> my_string = "This is a sample String"
>>> my_string = my_string.lower().replace(" ", "_")
>>> print(my_string)
this_is_a_sample_string
```

 #594

### DIFF
--- a/plugins/lookup/controller_object_diff.py
+++ b/plugins/lookup/controller_object_diff.py
@@ -202,7 +202,7 @@ class LookupModule(LookupBase):
             for item in api_list_reduced:
                 if item["resource_type"] == "organization":
                     item.update({"organizations": [item[item["resource_type"]]]})
-                item.update({"role": item["name"].lower()})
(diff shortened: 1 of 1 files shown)


adonisgarciac:
I can't add reviewers but I'd like @ivarmu can check this change :smiley:

ivarmu (approved):
LGTM

djdanielsson (approved):
LGTM

PR #597, opened 2023-05-06
add options to roles role

sean-m-sullivan:
# What does this PR do?
add options to roles role. This is not live in a published awx.awx collection but it has been merged.
Adds teams/users as options so that lists can be consolidated.

# How should this be tested?
Additional tests should be added later, adding now will fail CI.

https://github.com/ansible/awx/pull/13930


djdanielsson (approved):
LGTM

PR #602, opened 2023-05-12
Controller Schedule: forks, timeout and limit can only be set if template is configured as "prompt on launch"

Klaas-:
# What does this PR do?
Controller Schedule: forks, timeout and limit can only be set if template is configured as "prompt on launch" for these arguments. This restores the default(omit, true), but maybe it should be default(omit)

# How should this be tested?
Create a schedule, do not modify an existing one as this works (guess this is a awx.awx.schedule module or even awx api error)

no


sean-m-sullivan (approved):
LGTM, we are working on moving these to modules.

PR #603, opened 2023-05-15
Fix job_template notifications

Klaas-:
# What does this PR do?
I think this is what you meant to do :) Please check, in the current devel version code looks like this:

https://github.com/redhat-cop/controller_configuration/blob/a764ec7a64725ce35662a7dd2f655b1d9b589432/roles/job_templates/tasks/main.yml#L57-L59

That means the first if will always evaluate to `''` or `__controller_template_item.related.notification_templates_started` never to `__controller_template_item.notification_templates_started`

I think you meant to leave out the `else ''` which is what this PR does.

# How should this be tested?

Create a job template with a notification, currently it does not work, after PR it works again, I have not tested the .related. code path.

no


sean-m-sullivan (approved):
Needed to test this, it needs to be updated elsewhere, but accepting this change.

Klaas-:
I did not check all roles, but just noticed it here :)

PR #613, opened 2023-06-08
Make 'with_present' variable

Rickmarges:
# What does this PR do?

This PR makes the 'with_present' option variable (with default false). The actual plugin uses true as default, but the object_diff role uses a hardcoded 'false' value.

# How should this be tested?

The example given in the object_diff role should still work as designed. This will only remove differences that were added by hand.

```
ansible-playbook drop_diff.yml --tags ${CONTROLLER_OBJECT} -e "{orgs: ${ORGANIZATION}, dir_orgs_vars: orgs_vars, env: ${ENVIRONMENT} }" --vault-password-file ./.vault_pass.txt -e @orgs_vars/env/${ENVIRONMENT}/configure_connection_controller_credentials.yml ${OTHER}
```

But after this change you could provide the variable `include_present_state` to also include the present state.
So the output of the objects to manage contain the items to remove, as well as the items that need to be added


sean-m-sullivan (approved):
Good catch on the change, I didn't think of this.

PR #615, opened 2023-06-09
Fix the diff functionality when using with_present

Rickmarges:
# What does this PR do?

The PR extends on the previous PR #613 

While testing with `with_present: true` we noticed no items were created when the api does not provide any results.
This is a small rework of the diff calculation to also add the `state: present` items whenever the api is empty.

# How should this be tested?

There are  a few cases to test the behavior of `with_present: true`

1. Empty API and empty configuration
2. Filled API and empty configuration
3. Empty API and filled configuration
4. Filled API and filled configuration

In all of these cases the diff functionality works as expected.

### DIFF
--- a/plugins/lookup/controller_object_diff.py
+++ b/plugins/lookup/controller_object_diff.py
@@ -105,6 +105,14 @@ class LookupModule(LookupBase):
     def warn_callback(self, warning):
         self.display.warning(warning)
 
+    def create_present_list(self, compare_list):
+        if not compare_list:
+            return [compare_list]
+
+        for item in compare_list:
+            item.update({"state": "present"})
+        return compare_list
+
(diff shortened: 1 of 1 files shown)


plugins/lookup/controller_object_diff.py
```diff
@@ -105,6 +105,14 @@ def handle_error(self, **kwargs):
     def warn_callback(self, warning):
         self.display.warning(warning)
 
+    def create_present_list(self, api_list, compare_list):
```
sean-m-sullivan: Why are we even passing the api list if we do nothing with it. we should just cut that entire part out.

sean-m-sullivan (approved):
LGTM, just trying to cut down on operations, keeping it simple :)

PR #642, opened 2023-07-06
add organizations tag to second dispatch task

adonisgarciac:
# What does this PR do?

The second task of dispatch role was not being executed when organizations tag was sent. With this change, organizations will be created properly when organizations tag is present in the execution.

# How should this be tested?

Add a new organization using the organizations tag in the execution.


djdanielsson (approved):
LGTM

sean-m-sullivan (approved):

PR #647, opened 2023-07-13
Change lookup to query to ensure list

Rickmarges:
# What does this PR do?

When first configuring an organization with only 1 inventory defined, the configuration fails with the message:
`argument 'controller_inventories' is of type <class 'dict'> and we were unable to convert to list`

The diff_role tasks use the `lookup` plugin to calculate the differences, which doesn't necessarily create a list when it recieves one item. This change will move from `lookup` to `query`, which will always return a list (see relevant link for more info)

# How should this be tested?

1. Create an empty organization (without inventories)
2. Define 1 inventory for created organization
3. Run the configuraion with diff_role and `with_present: true`

https://docs.ansible.com/ansible/latest/plugins/lookup.html#forcing-lookups-to-return-lists-query-and-wantlist-true

### DIFF
--- a/changelogs/fragments/change_from_lookup_to_query.yml
+++ b/changelogs/fragments/change_from_lookup_to_query.yml
@@ -1,0 +1,5 @@
+---
+bugfixes:
+  - 'Fix addition of `state: present` when `with_present: true`'
+minor_changes:
+  - Change from lookup to query in the object_diff task files
--- a/plugins/lookup/controller_object_diff.py
+++ b/plugins/lookup/controller_object_diff.py
@@ -67,7 +67,7 @@ EXAMPLES = """
 
 - name: "Find the difference of Project between what is on the Controller versus curated list."
   set_fact:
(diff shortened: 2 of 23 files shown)


Rickmarges:
We're still testing this further. We ran into some issues, but are not yet sure if they're related to this.
I will post an update ASAP

Rickmarges:
So, I've investigated the issue a bit more. Due to the last return statement (`return [difference]`) in combination with the `query` plugin from Ansible, caused certain lists to be set to a nested loop (`[[...]]`) instead of a single one (`[...]`).

With the currently added 2 commits, this issue should be resolved. And in my opinion this is ready to be merged once approved :)

ivarmu (approved):
LGTM

sean-m-sullivan (approved):
@ivarmu this looks good, looks like we need a changelog though as this is a breaking change, as well as the present/compare list bugfix,

sean-m-sullivan (changes requested):
Changelog needed,

sean-m-sullivan (approved):

PR #652, opened 2023-07-21
update Bugfixes

sean-m-sullivan:
# What does this PR do?

remove colon in the bugfix that snuck through due to the great pyyaml disturbance


sean-m-sullivan:
This incorporates an earlier PR here, and and update to tooling to fix precommit/sanity tests.

djdanielsson (approved):

PR #653, opened 2023-07-25
add controller_configuration_async_dir var

benformosa:
# What does this PR do?

Creates a new variable `controller_configuration_async_dir` on all roles which use the async keyword on tasks, which sets the `ansible_async_dir` variable on those tasks, and sets the default to `null`, from `/tmp/.ansible_async`.

This is effectively the same change made on infra.ah_configuration in ansible/automation_hub_collection/pull/273 and ansible/automation_hub_collection/pull/276.

I haven't had the same issue with this collection as I had with infra.ah_configuration, but it has the same design, so it makes sense that it should work the same.

# How should this be tested?

I haven't tested this thoroughly, but it seems to work OK.

See also https://github.com/redhat-cop/controller_configuration/issues/299#issuecomment-1493322272

### DIFF
--- a/changelogs/fragments/async.yml
+++ b/changelogs/fragments/async.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - Add option to change async directory, and set the default to null. /tmp/.ansible_async was a workaround while the default was broken previously.
+...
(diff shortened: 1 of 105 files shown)


djdanielsson (approved):
LGTM

PR #658, opened 2023-07-31
Add roles option to roles role to allow setting multiple roles in one item

### LINKED ISSUE #606: Roles controller support for multiple `role` per assignment
As per the README found in controller roles/roles, `role` which defines what role to give from a set of possible choice is a STR.
Which means that for instance, you'll need many instances of this snippet to assign multiple role to teams on organization:
```
controller_roles:
  - teams: X
    organization: Y
    role: auditor
  - teams: X
    organization: Y
    role: project_admin
 - teams: X
    organization: Y
    role: execute
    ....
```

It'd be much easier if `role` would support a list

Also the `role` description in the README mentions `target_team` twice but not `organization`, looks like a typo.

Tompage1994:
# What does this PR do?
Add roles option to roles role to allow setting multiple roles in one item rather than repeating entire sections of code

# How should this be tested?
Added a new automated test

resolves #606 

N/A

### DIFF
--- a/changelogs/fragments/roles.yml
+++ b/changelogs/fragments/roles.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - Added roles option to roles role to allow setting multiple roles in one item rather than repeating entire sections of code
+...
--- a/roles/roles/README.md
+++ b/roles/roles/README.md
@@ -79,6 +79,7 @@ This also speeds up the overall role.
 |`users`|""|no|list|The users for which the role applies|
 |`team`|""|no|str|The team for which the role applies|
 |`teams`|""|no|list|The teams for which the role applies|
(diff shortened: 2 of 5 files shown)


djdanielsson (approved):
This looks like a great improvement

ivarmu (approved):
Great!

ivarmu (changes requested):
Maybe you can reduce code/task duplication with the following idea:
```yaml
---
- hosts: localhost
  connection: local
  gather_facts: false
  vars:
    - data:
        - name: row1
          roles:
            - role1
            - role2
        - name: row2
          role: roleA
  tasks:
    - debug:
        msg:
          - "{{ item.1 if item.1 is defined else item.role }}"
      loop: "{{ (data | selectattr('role', 'defined')) + (data | subelements(['roles'], skip_missing=true)) }}"
...
```

ivarmu:
Yeah, I'm late... I was thinking about this... what do you think?

Tompage1994:
> Yeah, I'm late... I was thinking about this... what do you think?

Yes, that makes sense and it should work. I'll look at implementing another change here

PR #662, opened 2023-08-01
Deduplicate role task in roles role

### LINKED ISSUE #661: Improvement: De-duplicate task in roles role
Maybe you can reduce code/task duplication with the following idea:
```yaml
---
- hosts: localhost
  connection: local
  gather_facts: false
  vars:
    - data:
        - name: row1
          roles:
            - role1
            - role2
        - name: row2
          role: roleA
  tasks:
    - debug:
        msg:
          - "{{ item.1 if item.1 is defined else item.role }}"
      loop: "{{ (data | selectattr('role', 'defined')) + (data | subelements(['roles'], skip_missing=true)) }}"
...
```

_Originally posted by @ivarmu in https://github.com/redhat-cop/controller_configuration/pull/658#pullrequestreview-1556989197_

Tompage1994:
# What does this PR do?

De-duplicates the role task in the roles role by using defaults and a more clever loop

Unlike the example in the issue, I've simplified slightly as it would have been messy doing the `if/else` stuff in the task. Instead I've used defaults which is much neater. 

# How should this be tested?
CI already added

resolves #661

There hasn't been a release since the PR which this is updating so no additional changelog required.

### DIFF
--- a/roles/roles/tasks/main.yml
+++ b/roles/roles/tasks/main.yml
@@ -1,5 +1,5 @@
 ---
-- name: Create Roles Based Access Entry on Controller
+- name: Create Role Based Access Entry on Controller
   role:
     user:                           "{{ __controller_role_item.0.user | default(__controller_role_item.user) | default(omit, true) }}"
     users:                          "{{ __controller_role_item.0.users | default(__controller_role_item.users) | default(( [] if controller_configuration_role_enforce_defaults else omit), true) }}"
@@ -40,7 +40,7 @@
   register: __controller_role_job_async
   changed_when: not __controller_role_job_async.changed
   vars:
-    ansible_async_dir: '/tmp/.ansible_async'
+    ansible_async_dir: '{{ controller_configuration_async_dir }}'
 
 - name: "Configure Roles | Wait for finish the Roles creation"
   ansible.builtin.async_status:
--- a/roles/roles/tasks/main.yml
+++ b/roles/roles/tasks/main.yml
@@ -1,28 +1,28 @@
 ---
 - name: Create Roles Based Access Entry on Controller
   role:
-    user:                           "{{ __controller_role_item.0.user | default(omit, true) }}"
-    users:                          "{{ __controller_role_item.0.users | default(( [] if controller_configuration_role_enforce_defaults else omit), true) }}"
-    team:                           "{{ __controller_role_item.0.team | default(omit, true) }}"
-    teams:                          "{{ __controller_role_item.0.teams | default(( [] if controller_configuration_role_enforce_defaults else omit), true) }}"
-    role:                           "{{ __controller_role_item.1 | mandatory }}"
-    target_team:                    "{{ __controller_role_item.0.target_team | default(omit, true) }}"
-    target_teams:                   "{{ __controller_role_item.0.target_teams | default(( [] if controller_configuration_role_enforce_defaults else omit), true) }}"
-    inventory:                      "{{ __controller_role_item.0.inventory | default(omit, true) }}"
-    inventories:                    "{{ __controller_role_item.0.inventories | default(( [] if controller_configuration_role_enforce_defaults else omit), true) }}"
-    job_template:                   "{{ __controller_role_item.0.job_template | default(omit, true) }}"
-    job_templates:                  "{{ __controller_role_item.0.job_templates | default(( [] if controller_configuration_role_enforce_defaults else omit), true) }}"
-    workflow:                       "{{ __controller_role_item.0.workflow | default(omit, true) }}"
-    workflows:                      "{{ __controller_role_item.0.workflows | default(( [] if controller_configuration_role_enforce_defaults else omit), true) }}"
-    credential:                     "{{ __controller_role_item.0.credential | default(omit, true) }}"
-    credentials:                    "{{ __controller_role_item.0.credentials | default(( [] if controller_configuration_role_enforce_defaults else omit), true) }}"
-    organization:                   "{{ __controller_role_item.0.organization | default(omit, true) }}"
-    organizations:                  "{{ __controller_role_item.0.organizations | default(( [] if controller_configuration_role_enforce_defaults else omit), true) }}"
-    lookup_organization:            "{{ __controller_role_item.0.lookup_organization | default(omit, true) }}"
-    project:                        "{{ __controller_role_item.0.project | default(omit, true) }}"
-    projects:                       "{{ __controller_role_item.0.projects | default(( [] if controller_configuration_role_enforce_defaults else omit), true) }}"
-    instance_groups:                "{{ __controller_role_item.0.instance_groups | default(( [] if controller_configuration_role_enforce_defaults else omit), true) }}"
-    state:                          "{{ __controller_role_item.0.state | default(controller_state | default('present')) }}"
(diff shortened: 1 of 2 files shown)


djdanielsson (approved):
LGTM

ivarmu (approved):
Looks better, but there's a situation that wouldn't be desired... if the input contains the following... it would silently bypass the wrong formatted entry (it is missing both `role` and `roles` parameters):

```
controller_roles:
  - team: "team1"
    credential: "credential1"
```

It's not a stopper, but may need to add an extra check for that situation...

roles/roles/tasks/main.yml
```diff
@@ -42,50 +42,6 @@
...
-  when: not __controller_role_item.roles is defined
-  loop_control:
-    loop_var: __controller_role_item
-  no_log: "{{ controller_configuration_role_secure_logging }}"
-  async: 1000
-  poll: 0
-  register: __controller_role_job_async
-  changed_when: not __controller_role_job_async.changed
-  vars:
-    ansible_async_dir: '{{ controller_configuration_async_dir }}'
```
ivarmu: Could you maintain these lines?

ivarmu (changes requested):

ivarmu:
> Looks better, but there's a situation that wouldn't be desired... if the input contains the following... it would silently bypass the wrong formatted entry (it is missing both `role` and `roles` parameters):
> 
> ```
> controller_roles:
>   - team: "team1"
>     credential: "credential1"
> ```
> 
> It's not a stopper, but may need to add an extra check for that situation...

Adding the following `fail` task would be enough to raise a failure when the input items has a lack of both `role` and `roles` fields:

```yaml
---
- hosts: localhost
  connection: local
  gather_facts: false
  vars:
    - data:
        - name: row1
          roles:
            - role1
            - role2
        - name: row2
          role: roleA
        - name: row3
  tasks:
    - fail:
        msg: "The following input is incomplete: {{ current_bad_input }}"
      vars:
        bad_input: "{{ data | selectattr('role', 'undefined') | selectattr('roles', 'undefined') }}"
      loop: "{{ bad_input }}"
      loop_control:
        loop_var: current_bad_input

    - debug:
        msg:
          - "{{ item.1 if item.1 is defined else item.role }}"
      loop: "{{ (data | selectattr('role', 'defined')) + (data | subelements(['roles'], skip_missing=true)) }}"
...
```

What do you think?

Tompage1994:
> > Looks better, but there's a situation that wouldn't be desired... if the input contains the following... it would silently bypass the wrong formatted entry (it is missing both `role` and `roles` parameters):
> > ```
> > controller_roles:
> >   - team: "team1"
> >     credential: "credential1"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It's not a stopper, but may need to add an extra check for that situation...
> 
> Adding the following `fail` task would be enough to raise a failure when the input items has a lack of both `role` and `roles` fields:
> 
> ```yaml
> ---
> - hosts: localhost
>   connection: local
>   gather_facts: false
>   vars:
>     - data:
>         - name: row1
>           roles:
>             - role1
>             - role2
>         - name: row2
>           role: roleA
>         - name: row3
>   tasks:
>     - fail:
>         msg: "The following input is incomplete: {{ current_bad_input }}"
>       vars:
>         bad_input: "{{ data | selectattr('role', 'undefined') | selectattr('roles', 'undefined') }}"
>       loop: "{{ bad_input }}"
>       loop_control:
>         loop_var: current_bad_input
> 
>     - debug:
>         msg:
>           - "{{ item.1 if item.1 is defined else item.role }}"
>       loop: "{{ (data | selectattr('role', 'defined')) + (data | subelements(['roles'], skip_missing=true)) }}"
> ...
> ```
> 
> What do you think?

I think you can go without as the `mandatory` filter on the input essentially does that job. If you don't specify `role` or `roles` you get the following output:

```
fatal: [localhost]: FAILED! => {"msg": "Mandatory variable 'role'  not defined."}
```
I think that will be fine to be honest.

roles/roles/tasks/main.yml
Tompage1994: They are still there for the task that remains it's just hidden in the diff because nothing has changed. If you open up the full file you'll see it.

Unless I'm missing something...

ivarmu: I'm sorry... you are right. I saw the tests were failling and thought in this at the first place:

https://github.com/redhat-cop/controller_configuration/actions/runs/5728794937/job/15524141848?pr=662

ivarmu:
> I think that will be fine to be honest.

Ok, just to be sure that my concern is clear... given the following example:

```yaml
---
- hosts: localhost
  connection: local
  gather_facts: false
  vars:
    - data:
        - name: row1
          roles:
            - role1
            - role2
        - name: row2
          role: roleA
        - name: row3
  tasks:
    # - fail:
    #     msg: "The following input is incomplete: {{ current_bad_input }}"
    #   vars:
    #     bad_input: "{{ data | selectattr('role', 'undefined') | selectattr('roles', 'undefined') }}"
    #   loop: "{{ bad_input }}"
    #   loop_control:
    #     loop_var: current_bad_input

    - debug:
        msg:
          - "{{ item.1 if item.1 is defined else item.role }}"
      loop: "{{ (data | selectattr('role', 'defined')) + (data | subelements(['roles'], skip_missing=true)) }}"
...
```
This will skip the `data` -> `row3` item, as loop is selecting only the list items having defined either of `role` or `roles`, discarding any posible item without both of them, so the `mandatory` term will not have the opportunity to shine:

```console
ansible-playbook /tmp/test.yaml
[WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost does not match 'all'

TASK [debug] *********************************************************************************************************************************************************************************************************************************
ok: [localhost] => (item={'name': 'row2', 'role': 'roleA'}) => {
    "msg": [
        "roleA"
    ]
}
ok: [localhost] => (item=[{'name': 'row1', 'roles': ['role1', 'role2']}, 'role1']) => {
    "msg": [
        "role1"
    ]
}
ok: [localhost] => (item=[{'name': 'row1', 'roles': ['role1', 'role2']}, 'role2']) => {
    "msg": [
        "role2"
    ]
}

localhost                  : ok=1    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   

```

Do we want to let that kind of incomplete configuration entries in the input configuration as code?

Tompage1994:
> 
That doesn't happen because instead of testing if `role` is defined, I've tested if `roles` is undefined, meaning if neither is set then it will be picked up and will spit out the error.

The only possible issue is when BOTH are defined, but in that case it will just use the `roles` option and `role` will be ignored

This:
```
tasks:
  - name: Include roles role at ../roles/roles
      ansible.builtin.include_role:
        name: ../roles/roles
      vars:
        controller_roles:
          - user: user1
            job_template: Demo Job Template
            roles:
              - read
              - execute
          - inventory: Demo Inventory
            user: user1
            role: read
          - job_template: Demo Job Template
```

Outputs this:
```
...
fatal: [localhost]: FAILED! => {
    "msg": "Mandatory variable 'role'  not defined."
}
```

ivarmu (approved):

PR #668, opened 2023-08-09
Use version for PR in the action

Tompage1994:
# What does this PR do?
Makes it so that the PR is actually getting tested rather than the devel branch

# How should this be tested?
In the action run you can see which version of the repo is being checked out.

N/A

resolves #[number]

This is equivalent to the changes in galaxy_collection. We should wait to make sure that is working there before this is merged


Tompage1994:
Working in galaxy_collection so should be good

djdanielsson (approved):
LGTM

sean-m-sullivan (approved):

Issue #671, opened 2023-08-12
Assigning Default Environment to Organization


willtome:
# Summary
I'm not sure if this is intended behavior however when assigning the default environment to an organization, it requires a dictionary instead of a string. Either role or documentation needs to be updated.

# Issue Type

- Bug Report

# Ansible, Collection, Controller details

```console (paste below)
ansible --version

Controller versionansible [core 2.15.0]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/lab-user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.9/site-packages/ansible
  ansible collection location = /home/lab-user/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.9.7 (default, Sep 13 2021, 08:18:39) [GCC 8.5.0 20210514 (Red Hat 8.5.0-3)] (/usr/bin/python3.9)
  jinja version = 3.1.2
  libyaml = True

ansible-galaxy collection list

# /home/lab-user/test/collections/ansible_collections
Collection                     Version
------------------------------ -------
amazon.aws                     2.2.0  
ansible.posix                  1.3.0  
ansible.utils                  2.7.0  
awx.awx                        22.6.0 
community.general              4.6.1  
infra.controller_configuration 2.5.0  
kubernetes.core                2.3.1  

Controller version: 4.4.1

```

- ansible installation method: OS package

# OS / ENVIRONMENT

RHEL 8.6

# Desired Behavior

When an organization is defined as documented, the default_environment is assigned to the org.
```
  controller_organizations:
      - name: Default
        default_environment: 'product-demos'
        galaxy_credentials:
          - Ansible Galaxy
          - Automation Hub
```

However, looking at [this line](https://github.com/redhat-cop/controller_configuration/blob/8b97ad07815a921bc8a33871e05b609c97966107/roles/organizations/tasks/main.yml#L9) it appears that the default_environment parameter is expecting a dictionary. When the variables are defined like this it works as expected.

```
    controller_organizations:
      - name: Default
        default_environment: 
          name: 'product-demos'
        galaxy_credentials:
          - Ansible Galaxy
          - Automation Hub
```

sean-m-sullivan:
It should work both ways, is it not working with option 1

The first example you shared is the simplified data model that we originally started with.
[...]

Issue #670, opened 2023-08-11
redhat_cop.controller_configuration.license usage: parameters are mutually exclusive: manifest|pool_id


sean-m-sullivan:
### Discussed in https://github.com/redhat-cop/controller_configuration/discussions/667

<div type='discussions-op-text'>

<sup>Originally posted by **jrigglem** August  9, 2023</sup>
To automate the registration portion of AAP, I'm using this collection.role and would like to use the subscription model vs. a manifest.

When specifying the pool_id instead of a manifest name, the play is erroring out with an error which indicates manifest is being set.  

[reg_yml.txt](https://github.com/redhat-cop/controller_configuration/files/12304240/reg_yml.txt)
[output.txt](https://github.com/redhat-cop/controller_configuration/files/12304258/output.txt)

</div>

PR #672, opened 2023-08-12
Org environment fix

sean-m-sullivan:
# What does this PR do?
bugfixes:
  - Fixed issue with organization role not acceppting default environments option correctly.
  - Fixed issue with licence role not operating properly, when a controller never had credentials provided for subscription lookup. See Role Readme for proper usuage.

#671 
#670 

#667 

@willtome 
@jrigglem


roles/license/README.md
```diff
@@ -59,23 +61,20 @@ The module and this role can use either a manifest file, or lookup the subscript
...
+|`redhat_subscription_password`|""|yes|str|Red Hat or Red Hat Satellite password to get available subscriptions. Global var not part of controller_license data structure.|
 |`filters`|"default values"|no|str|dict of filters to use to narrow the subscription. See example below for how to use this.|
 |`support_level`|"Self-Support"|no|str|DEPRECATED - changed to `manifest_file` (still works as an alias)|
 |`list_num`|0|no|int|List index of the subscription to use, if you want to overide the default, it is recomended to use the filters to limit the pools found.|
-|`pool_id`|""|no|str|Red Hat or Red Hat Satellite pool_id to attach to, setting this will skip the lookup.|
+|`pool_id`|""|no|str|Red Hat or Red Hat Satellite pool_id to attach to.|
 |`force`|`False`|no|bool|By default, the license will only be applied if controller is currently unlicensed or trial licensed. When force=true, the license is always applied.|
+|`controller_license_use_lookup`|`False`|no|bool|Whether or not to lookup subscriptions. Global var not part of controller_license data structure.|
```
Tompage1994: This shouldn't be in this bit of the variable listing but should instead be further up with the rest of the global vars

roles/license/README.md
```diff
@@ -59,23 +61,20 @@ The module and this role can use either a manifest file, or lookup the subscript
...
 
-For further details on fields see <https://docs.ansible.com/automation-controller/latest/html/userguide/credential_plugins.html>
-
 ### License Variables for using Red Hat Subscription
 
 |Variable Name|Default Value|Required|Type|Description|
 |:---:|:---:|:---:|:---:|:---:|
-|`redhat_subscription_username`|""|no|str|Red Hat or Red Hat Satellite username to get available subscriptions.|
-|`redhat_subscription_password`|""|no|str|Red Hat or Red Hat Satellite password to get available subscriptions.|
+|`redhat_subscription_username`|""|yes|str|Red Hat or Red Hat Satellite username to get available subscriptions. Global var not part of controller_license data structure.|
```
Tompage1994: Likewise, these should be above with the rest of the global vars

Tompage1994: Also, I'm pretty certain these variables are not required if using a manifest. So they should be set as required: false

Tompage1994: Can you also add any global vars to the `meta/argument_spec.yml` file for this role please

Tompage1994 (changes requested):

roles/license/README.md
sean-m-sullivan: So there is no global section above, There are two sections used

Manifest
Subscription

so put each one in each, and in the subscription section, and maybe we should break it out as two seperate roles? or how should we structure the readme better. I have updated the argspec

Tompage1994 (approved):
LGTM

roles/license/defaults/main.yml
```diff
@@ -1,5 +1,6 @@
 ---
 controller_configuration_license_secure_logging: "{{ controller_configuration_secure_logging | default('false') }}"
+controller_license_use_lookup: false
```
Tompage1994: Don't think this is needed actually?

changelogs/fragments/licence_org.yml
```diff
@@ -0,0 +1,5 @@
+---
```
Tompage1994, line 1: We probably need a minor_change around the `use_lookup` change now

Tompage1994 (changes requested):
Very minor points

silvinux (commented):
LGTM!

Tompage1994 (approved):

djdanielsson (approved):
LGTM

Tompage1994:
CI broken. Once #674 is merged this can be rebased and tested

PR #675, opened 2023-08-16
Fix broken CI because of request_timeout added to awx

Tompage1994:
# What does this PR do?
Fix broken CI because of request_timeout added to awx

New release of awx.awx caused the CI to fail because the `request_timeout` in the module_utils which are used

# How should this be tested?
CI starts passing

N/A

Failing CI: https://github.com/redhat-cop/controller_configuration/actions/runs/5875280508/job/15931315104#step:9:286


djdanielsson (approved):
LGTM

PR #677, opened 2023-08-17
Add to description of request_timeout option that awx.awx>=22.7.0 is required

Tompage1994:
# What does this PR do?
We need to have latest awx.awx if the request_timeout option is to be added to the plugin because it relies upon the module_utils from awx.awx where the option is defined. I'm pretty sure it'll spit out an error if the version isn't correct (haven't tested) and at best it'll just not do anything.

# How should this be tested?
Doc change only

Addition to #675 
^ This hasn't been released so the same changelog fragment is fine


djdanielsson (approved):
LGTM

PR #680, opened 2023-08-22
sanity rework

sean-m-sullivan:
# What does this PR do?

adds a configuration file for testing as per 
https://github.com/ansible-collections/community.general/blob/main/tests/config.yml#L7

# How should this be tested?
A further rework of our sanity test will use this file, this is a prereq


djdanielsson (approved):
LGTM

Issue #685, opened 2023-08-29
dispatch role failes to create organization if instance group doesn't exist yet


jon4hz:
# Summary
The dispatch role failes to create an organization if you assign an instance group to this organization that doesn't exist yet.
I experienced this issue when I tried to create an org and an instance group in the same run. 

# Issue Type

- Bug Report

# Ansible, Collection, Controller details

```console (paste below)
ansible-core: 2.15.3
controller_configuration: 2.5.0
Controller version: 4.4.0

```

- ansible installation method: pip

# OS / ENVIRONMENT

The playbook was executed in a gitlab pipeline. The runner image is based on alpine 3.18.

# Desired Behavior

I'd expect the dispatch role to create the instance group before creating the organization. Or create the organization without the instance group first and then assign it after creating the instance group (same as the role does for ee).

# Actual Behavior

The role fails because the instance group doesn't exist yet and therefore can't be assigned to the org.

```console (error)
TASK [infra.controller_configuration.organizations : Configure Controller Organizations | Wait for finish the organization creation] ***
failed: [my.host] (item={'failed': 0, 'started': 1, 'finished': 0, 'ansible_job_id': 'j600586684650.1889', 'results_file': '/app/.ansible_async/j600586684650.1889', 'changed': False, '__controller_organizations_item': {'name': 'myorg', 'description': 'myorg', 'instance_groups': 'myorg default'}, 'ansible_loop_var': '__controller_organizations_item'}) => {"__organizations_job_async_results_item": {"__controller_organizations_item": {"description": "myorg", "instance_groups": "myorg default", "name": "myorg"}, "ansible_job_id": "j600586684650.1889", "ansible_loop_var": "__controller_organizations_item", "changed": false, "failed": 0, "finished": 0, "results_file": "/app/.ansible_async/j600586684650.1889", "started": 1}, "ansible_job_id": "j600586684650.1889", "ansible_loop_var": "__organizations_job_async_results_item", "attempts": 1, "changed": false, "finished": 1, "msg": "Request to /api/v2/instance_groups/?name=myorg+default returned 0 items, expected 1", "query": {"name": "myorg default"}, "response": {"json": {"count": 0, "next": null, "previous": null, "results": []}, "status_code": 200}, "results_file": "/app/.ansible_async/j600586684650.1889", "started": 1, "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": [], "total_results": 0}
```

[...]

PR #686, opened 2023-08-30
[#685] Fix organization creation with instance group

adonisgarciac:
# What does this PR do?

If an organization has instance_groups defined, it has to be created beforehand. With this PR, instance and instance_group will be executed before organizations in the dispatch role.

# How should this be tested?

Create organizations with instance_groups

#685


Tompage1994 (approved):
Looks good

PR #691, opened 2023-09-13
rectified misspelled variable in job_templates

myselfgagandeep:
# What does this PR do?
This should fix the variable naming for the `ask_labels_on_launch` within job_templates that causes all the templates to be exported with erroneous naming.

Rectified misspelled variable that breaks the dispatch role while importing the yaml for job_templates
# How should this be tested?

resolves #[number]

### DIFF
--- a/changelogs/fragments/filetree_create_job_templates.yml
+++ b/changelogs/fragments/filetree_create_job_templates.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - filetree_create - Fixed the misspelled variable name that caused exported job_templates yaml files containing incorrect name.
+...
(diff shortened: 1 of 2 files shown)


sean-m-sullivan:
Can you add a file in the changelog fragments folder like this, 
https://github.com/redhat-cop/controller_configuration/blob/devel/changelogs/fragments/instancegroups_org.yml

Describing the bugfix and to which role.

otherwise looks like a great catch/fix.

myselfgagandeep:
Have updated the merge request with required fragment file

sean-m-sullivan (approved):
LGTM, thanks for the PR!

PR #692, opened 2023-09-14
Add failed_when to force the fail when there are undefined variables

ivarmu:
# What does this PR do?
Given the following files:
```
$ cat playbook.yaml 
---
- hosts: localhost
  connection: local
  gather_facts: false
  tasks:
    - ansible.builtin.include_vars:
        file: vars2.yaml
      register: variables
      failed_when: "'VARIABLE IS NOT DEFINED' in variables"
    - ansible.builtin.debug:
        var: variables
...
$ cat vars2.yaml 
---
mid_value: "{{ var_value }}"
...
$ cat vars1.yaml 
---
var_value: "hello world"
...
$ ansible-playbook playbook.yaml
[WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost does not match 'all'

TASK [ansible.builtin.include_vars] **********************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [ansible.builtin.debug] *****************************************************************************************************************************************************************************************************************
ok: [localhost] => {
    "variables": "VARIABLE IS NOT DEFINED!"
}

localhost                  : ok=2    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
```
The execution should fail because the variable `var_value` is not defined. `include_vars` is not failing naturally, so we must control that error explicitly.

# How should this be tested?
Manually tested.

N/A

N/A

### DIFF
--- a/roles/filetree_read/tasks/applications.yml
+++ b/roles/filetree_read/tasks/applications.yml
@@ -13,6 +13,7 @@
   loop_control:
     loop_var: __read_applications_definitions_item
   register: __contents_filetree_controller_applications
+  failed_when: "'VARIABLE IS NOT DEFINED' in __contents_filetree_controller_applications"
 
 - name: "Populate Applications list"
   ansible.builtin.set_fact:
--- a/roles/filetree_read/tasks/credential_input_sources.yml
+++ b/roles/filetree_read/tasks/credential_input_sources.yml
@@ -13,6 +13,7 @@
   loop_control:
     loop_var: __read_input_sources_definitions_item
   register: __contents_filetree_controller_credential_input_sources
+  failed_when: "'VARIABLE IS NOT DEFINED' in __contents_filetree_controller_credential_input_sources"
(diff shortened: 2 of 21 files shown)


silvinux (approved):
LGTM!

sean-m-sullivan:
If this is how it should look, great, don't love it as Variable is not defined though.

ivarmu:
> If this is how it should look, great, don't love it as Variable is not defined though.

Yeah! The problem is that the undefined variable name is not known, so it's the easiest way to detect the issue and stop the load process...

ivarmu:
> > If this is how it should look, great, don't love it as Variable is not defined though.
> 
> Yeah! The problem is that the undefined variable name is not known, so it's the easiest way to detect the issue and stop the load process... 

Maybe the more correct way should be modifying the `include_vars` module to fail when having this situation... I'm taking a look at this.

PR #694, opened 2023-09-17
Fix subelements conditional failure for filetree_create and object_diff

adonisgarciac:
# What does this PR do?

Loops with subelements were failing with error: "obj must be a list of dicts or a nested dict".

This was because subelement filter is executed before when and it was causing a failure when the list was not defined.

# How should this be tested?
- name: test
  hosts: localhost
  gather_facts: false
  vars:
    __controller_api_current_user_check_is_admin:
      is_superuser: false

  tasks:
        - name: "Test nested condition"
          ansible.builtin.set_fact:
            __full_controller_api_roles: []
          loop: "{{ current_users | default([]) | subelements('organizations', skip_missing=true) }}"
          when: __controller_api_current_user_check_is_admin.is_superuser


sean-m-sullivan (approved):

sean-m-sullivan:
needs to be rebased, and a line at end of file on the fragment.

Tompage1994 (approved):

PR #697, opened 2023-09-24
update links to forums

sean-m-sullivan:
# What does this PR do?
Updates our bug report, readme, and issue config to funnel people to use the forums.


djdanielsson (approved):
Was wondering if we should point to matrix as another option but idk

sean-m-sullivan:
@djdanielsson added matrix to the readme, its already in the issue config.

PR #704, opened 2023-10-09
Fix calling manifest when manifest_content is defined

DLV1810:
When users define manifest content like this

```
controller_license:
  manifest_content: "{{ lookup(...) }}"
```

It correctly calls the manifest.yml file and installs the license

# How should this be tested?

I've tested locally and manually, unsure if I need to add in automated tests.

No

N/A

### DIFF
--- a/roles/license/tasks/main.yml
+++ b/roles/license/tasks/main.yml
@@ -4,7 +4,7 @@
 - name: Use manifest file
   ansible.builtin.include_tasks: "manifest.yml"
   when:
-    - controller_license.manifest_file is defined or controller_license.manifest is defined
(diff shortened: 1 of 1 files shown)


ivarmu (dismissed):
This looks good to me. Thanks for the PR. For completeness, could you please add the following?

- Document the new parameter at https://github.com/redhat-cop/controller_configuration/blob/devel/roles/license/README.md
- Add a change log fragment

As there's no automated tests for licence by now... there's no need to add any test right now.

Thanks!

sean-m-sullivan (approved):

sean-m-sullivan:
LGTM, good catch on the logic

PR #708, opened 2023-10-12
Cleanup task name

dbk-rabel:
This jinja2 template is never evaluated. The loop is inside the task. There is only one task and it was called literally: "Include Tasks to get OBJECT DIFF {{ __task_diff.name }}"


sean-m-sullivan (approved):
Good change, thanks!

PR #709, opened 2023-10-13
Ensure dispatch is generalised

Tompage1994:
# What does this PR do?
Re-generalises the dispatch role. In theory you should be able to do _anything_ with it. This gets us back to that point.

# How should this be tested?
CI should still work

N/A

resolves #[number]

N/A


silvinux (commented):
LGTM!

sean-m-sullivan:
Going to play with this, I think there may be a better order where just after credential inputs, before projects we put in Ee's and 2nd org. So if they use just project it can get galaxy collections

Otherwise, LGTM good change

sean-m-sullivan (approved):
LGTM

PR #710, opened 2023-10-13
dispatch tweak

sean-m-sullivan:
# What does this PR do?
- dispatch - Fixed the order and behavior to run as a single task with options for organization behavior.

Also changed it from a json to yaml, which makes it easier to read and edit IMO


djdanielsson (approved):
LGTM

PR #716, opened 2023-10-20
Added role to host global vars to all other roles. Updated tasks tittles to solve issue707.

### LINKED ISSUE #707: Names of tasks and docs don't take into account the opportunity to delete objects
Hi.

Names of tasks and docs don't take into account the opportunity to delete objects.

For example, the task "Add organizations" can also delete organizations, not just add them.
https://github.com/redhat-cop/controller_configuration/blob/devel/roles/organizations/tasks/main.yml#L2

Same for READMEs:
https://github.com/redhat-cop/controller_configuration/blob/devel/roles/organizations/README.md?plain=1#L5

Yours
  David

ivarmu:
# What does this PR do?

Solves the Issue #707 and fixes some bogus taks titles in output.
Also fixes many lintering problems.

# How should this be tested?

Manual and Automatic tests has been executed.

resolves #707 

N/A


ivarmu:
I don't understand why I'm getting this lintering error:

`roles/global_vars/README.md:25 MD046/code-block-style Code block style [Expected: indented; Actual: fenced]`

The code block is written the same way as it is at other roles... and the configuration tells the code blocks to be `fenced`, not `indented`:

https://github.com/redhat-cop/controller_configuration/blob/9176050d4d910df3080ce31c3e413472c098321c/.markdownlint.yml#L16

sean-m-sullivan (approved):
sorry I've been absent, this looks good, Some of the spaces were style choices, but it makes sense to fix them, Do like the verb/action change and output, and its also taken a while to review.

PR #718, opened 2023-10-25
Cleanup task name

dbk-rabel:
This jinja2 template is never evaluated. The loop is inside the task. There is only one task and it was called literally: "Run infra.controller_configuration.{{ __role.role }}"

Very similar to #708

Although in this case I'm not sure if my proposed task name is the best.

Yours
  David

### DIFF
--- a/roles/dispatch/tasks/main.yml
+++ b/roles/dispatch/tasks/main.yml
@@ -1,5 +1,5 @@
 ---
-- name: "Run infra.controller_configuration.{{ __role.role }}"
+- name: "Run infra.controller_configuration roles"
   ansible.builtin.include_role:
     name: "{{ __role.role }}"
     apply:


dbk-rabel:
I removed my linter cleanup commit. I think there is very many linter warnings right now. Might be possible to fix a lot of them with ansible-lint --write, but I think this is too much for this specific pull request.

ivarmu:
Hi @dbk-rabel, I prefer to let the current names to work properly, as proposed in the PR #716 for giving a solution to the issue #707.

So, if you feel confortable with this, we can close this issue as duplicated.

ivarmu (changes requested):
Duplicate of issue #707 and PR #716

dbk-rabel:
> Hi @dbk-rabel, I prefer to let the current names to work properly, as proposed in the PR #716 for giving a solution to the issue #707.
> 
> So, if you feel confortable with this, we can close this issue as duplicated.

Yes, that's sounds good.  Thanks for putting in the effort!

PR #719, opened 2023-10-26
Change order of operations

Rickmarges:
# What does this PR do?

Add new/updated key to dict only after removing old ones.

When using plural versions of the `dupitems` (for example `job_templates`), only `team` and `role` remain in the `compare_list_reduced`.
This is due to an error in the order of operations. A singluar version is added to the new dict and then all occurences of `dupitems` are removed. However, the singular version is also in that list.

In this PR the order is changed to first removed unwanted keys and only then add the key we want

PS. also solved a unneeded noqa

# How should this be tested?

Manual tests:

1. In AWX/AAP manually add `execute` rights to `dummy_template` for the `dummy-team`.
2. Add to `controller_roles`: `{"role": "execute", "team": "dummy-team", "job_templates": ["dummy_template"]}`
3. Run a playbook to include diff and dispatch roles (with `present_state: true`) 

Before this PR the `execute` role would be removed (even though it's defined in config) from the team.
After this, the role is created/saved as expected


sean-m-sullivan (approved):
This looks good, makes sense, thanks for the contribution!

PR #721, opened 2023-11-06
Fix usage of compare_list_reduced variable

Rickmarges:
# What does this PR do?

The object diff plugin used the wrong variable in 1 place when using `with_present`.
When using that option, the `state: present` key is added to the `compare_list_reduced` dict, however, the `compare_list` was then extended and returned.

This also solves removing too much differences based on the if-statement on line 332 as less dict now have less than 3 keys

# How should this be tested?

Manual
1. Create list of roles to configure (for example: `{"controller_roles": [{"team": "dummy", "job_template": "dummy_template", "role": "read"}]`
2. Run the diff role with `include_present_state: true`
3. The diff output now actually contains the `state: present` key. Before it did not

### DIFF
--- a/plugins/lookup/controller_object_diff.py
+++ b/plugins/lookup/controller_object_diff.py
@@ -320,7 +320,7 @@ class LookupModule(LookupBase):
                 item.update({"state": "absent"})
         # Combine Lists
         if self.get_option("with_present"):
-            for item in compare_list_reduced:
+            for item in compare_list:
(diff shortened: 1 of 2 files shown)


ivarmu (approved):
LGTM

sean-m-sullivan (approved):
Sorry this took a bit for me to get to and review wanted to test and step through it, thanks!

PR #723, opened 2023-11-07
filetree_create generate unified files

ivarmu:
# What does this PR do?

- Add a new parameter for the `filetree_create` role that allows to export all the objects to single files, where each object type corresponds to a unique file that is containing all the objects of the given object type. This allows the new output files to be consumed both from the Ansible `group_vars` and the `filetree_read` role, increasing the flexibility for the end user.
- Adds a playbook to convert the current output files (structured) to a flatten ones
- Fixes some issues found in the output files

# How should this be tested?
Manual tests during the implementation.

N/A

N/A


djdanielsson (approved):
sounds like a very useful option for many users

sean-m-sullivan (approved):
Looks useful, wonder how we can extend this.

sean-m-sullivan:
This needs updated, and it wont allow me to rebase

sean-m-sullivan (approved):

PR #725, opened 2023-11-13
Fix calling manifest when manifest_url is defined

rdrazny:
# What does this PR do?

PR #704 added a fix for manifest_content, but manifest_url is still ignored. `manifest.yml` is still not included when manifest_url is defined.

# How should this be tested?

Manual test only.

No.

#704

### DIFF
--- a/roles/license/tasks/main.yml
+++ b/roles/license/tasks/main.yml
@@ -4,7 +4,7 @@
 - name: Use manifest file
   ansible.builtin.include_tasks: "manifest.yml"
   when:
-    - controller_license.manifest_file is defined or controller_license.manifest is defined or controller_license.manifest_content is defined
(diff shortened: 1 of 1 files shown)


sean-m-sullivan (approved):
Good catch, LGTM, however going to have to peek under the hood to figure out if the test failures are from this or not, I don't think they are but want to double check.

rdrazny:
> Good catch, LGTM, however going to have to peek under the hood to figure out if the test failures are from this or not, I don't think they are but want to double check.

Looks like content_objects "Survey Template with vars no org" and "workflow_test_template" have switched order in the actual output, and that's enough to throw the diff module off.

ivarmu:
I think replaying the failed jobs will work ok. It may be caused by a temporary connection issues when fetching the objects from the API. Also... only two of the test jobs are failing... it doesn't seem to be a real error in the code...

PR #737, opened 2023-12-07
Fixed roles diff when the roles are provided as a list

bogdanmuresan:
# What does this PR do?
The PR fixes roles diff, when the roles are provided as a list in the "roles" key, and not individually in "role".
All the elements in the compare list need to be expanded once more, for each role in the roles list.

At the moment, when using roles as a list, all the controller_roles are set to state absent.

# How should this be tested?

resolves #[number]

### DIFF
--- a/plugins/lookup/controller_object_diff.py
+++ b/plugins/lookup/controller_object_diff.py
@@ -300,6 +300,24 @@ class LookupModule(LookupBase):
             for item in list_to_remove:
                 compare_list_reduced.remove(item)
             compare_list_reduced.extend(list_to_extend)
+            # Expand all compare list elements when roles is provided as list
+            list_to_extend = []
+            list_to_remove = []
+            for item in compare_list_reduced:
+                expanded = False
+                dupitems = [
+                    "roles",
+                    "role"
+                ]
+                if "roles" in item:
+                    for role in item["roles"]:
+                        list_to_extend.append(self.map_item(item, "role", role, dupitems))
+                    expanded = True
+                if expanded:
+                    list_to_remove.append(item)
+            for item in list_to_remove:
+                compare_list_reduced.remove(item)
+            compare_list_reduced.extend(list_to_extend)
         elif (
             api_list[0]["type"] != "organization"
             and api_list[0]["type"] != "user"
--- a/plugins/lookup/controller_object_diff.py
+++ b/plugins/lookup/controller_object_diff.py
@@ -305,10 +305,7 @@ class LookupModule(LookupBase):
             list_to_remove.clear()
             for item in compare_list_reduced:
                 expanded = False
-                dupitems = [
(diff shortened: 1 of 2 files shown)


plugins/lookup/controller_object_diff.py
```diff
@@ -300,6 +300,24 @@ def run(self, terms, variables=None, **kwargs):
             for item in list_to_remove:
                 compare_list_reduced.remove(item)
             compare_list_reduced.extend(list_to_extend)
+            # Expand all compare list elements when roles is provided as list
+            list_to_extend = []
+            list_to_remove = []
```
ivarmu: LGTM! Only one thing... shouldn't it be better to use `list_to_extend.clear()` and `list_to_remove.clear()` instead of creating new lists and leaving the old ones orphaned?

ivarmu (approved):

ivarmu (changes requested):
Why don't adding the `role` and `roles` to the already existing `for` ? https://github.com/redhat-cop/controller_configuration/blob/64b3f62253f6a74bed6d15334edfc2b0f26a1c2b/plugins/lookup/controller_object_diff.py#L247-L262

plugins/lookup/controller_object_diff.py
bogdanmuresan: I am using the same variable name, so my assumption is the lists will be cleared and reused. Also, I see clear() is not available in all python versions.

bogdanmuresan:
> Why don't adding the `role` and `roles` to the already existing `for` ?
> 
> https://github.com/redhat-cop/controller_configuration/blob/64b3f62253f6a74bed6d15334edfc2b0f26a1c2b/plugins/lookup/controller_object_diff.py#L247-L262

Because roles is another dimension. So assuming we have:
```
---
controller_roles:
  - team: MYTEAM
    projects:
      - project1
      - project2  
    roles:
     - use
     - update
```
compare_list_reduced will need to have 4 elements at the end, a combination of the projects with each role:
project1 - use, project1 - update, project 2 - use, project2 - update

I don't think this can be done in the first loop, which expands only the projects. I then need to loop once more through the expanded list and expand  again for each role.

I guess I could move dupitems definition above the for, so it's scope also covers the second loop, but I thought it would be better to have them separate.

ivarmu (approved):

plugins/lookup/controller_object_diff.py
ivarmu: Not exactly, you are creating a new empty list, without freeing the old one. It's not so important because all the memory will be released when the execution is finished, but good to know.

bogdanmuresan: Thank you, changed it to clear(), if this is the best practice.

ivarmu (approved):
LGTM

PR #739, opened 2023-12-12
Fix organization wide permissions in roles object diff

bogdanmuresan:
# What does this PR do?

Attempts to fix role diff, when role is defined without specific resource_type - the role applies to all organization objects.
Sample controller_roles:

```---
controller_roles:
  - team: Ansible
    organization: Acme
    role: execute
```
When this is applied, it will add the "execute" permissions for all job templates, therefore, when object_diff runs, it should not remove those roles, returned by the api as individual roles, one per job template.

# How should this be tested?

resolves #[number]

### DIFF
--- a/changelogs/fragments/diff_roles_orga_wide.yml
+++ b/changelogs/fragments/diff_roles_orga_wide.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - Fixed roles diff when the role is set at the organization level for an user/team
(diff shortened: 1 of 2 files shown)


bogdanmuresan:
@ivarmu can you please have a look over this PR as well?

ivarmu (approved):
LGTM. I'm not really sure that these auto created roles should be necessary at all, but it's a good idea to leave them intact as the are created (and hopefully also managed) by the core processes.

I've relaunched some tests to let them pass...

ivarmu:
> LGTM. I'm not really sure that these auto created roles should be necessary at all, but it's a good idea to leave them intact as the are created (and hopefully also managed) by the core processes.
> 
> I've relaunched some tests to let them pass...

I've confirmed these auto created roles are needed for the roles to work as expected. Good catch @bogdanmuresan!

PR #747, opened 2024-01-03
avoid the hosts populated by a constructed inventory to be removed

### LINKED ISSUE #743: Controller host is removing/trying to remove constructed inventory hosts
# Summary

looks like object diff for hosts is not taking into account constructed inventories, and is setting state="absent" on hosts from constructed inventories
https://github.com/redhat-cop/controller_configuration/blob/devel/roles/object_diff/tasks/hosts.yml#L16

Error message:

```
  "msg": "Unable to delete host REDACTED: Delete constructed inventory hosts from input inventory.",
```

# Issue Type

- Bug Report

# Ansible, Collection, Controller details

```console (paste below)
ansible --version

Controller version

```

# OS / ENVIRONMENT

# Desired Behavior

# Actual Behavior

Please give some details of what is actually happening.
Include a [minimum complete verifiable example] with:

- playbook / task
- configuration file / list
- error

```console (error)

```

# STEPS TO REPRODUCE

```yaml (playbook/task)

```

```yaml (config/list/array/variables)

```

[minimum complete verifiable example]: http://stackoverflow.com/help/mcve

ivarmu:
# What does this PR do?
avoid the hosts populated by a constructed inventory to be removed, the same way it is controlled for the smart inventories and the inventories that are defining inventory sources.

# How should this be tested?
Automatic tests are not controlling this kind of scenarios.

resolves #743

N/A

### DIFF
--- a/changelogs/fragments/issue_743.yml
+++ b/changelogs/fragments/issue_743.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - Avoid the hosts populated by a constructed inventory to be removed
--- a/roles/object_diff/tasks/hosts.yml
+++ b/roles/object_diff/tasks/hosts.yml
@@ -13,7 +13,8 @@
                                             query_params={'organization': controller_organization_id.id,
                                               'has_inventory_sources': 'false',
                                               'not__total_hosts': '0',
-                                              'not__kind': 'smart'},
+                                              'not__kind': 'smart',
+                                              'not__kind': 'constructed'},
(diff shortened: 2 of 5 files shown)


Tompage1994 (changes requested):
Makes sense to me. We will need a changelog created for this though please.

Also if you have time, adding to the CI would be useful here to test for this

ivarmu:
> Makes sense to me. We will need a changelog created for this though please.
> 
> Also if you have time, adding to the CI would be useful here to test for this

I agree... I'm thinking on how to introduce these tests...

Tompage1994 (approved):

PR #763, opened 2024-02-07
Fix docs for `assign_galaxy_credentials_to_org` and `assign_default_e…

dbk-rabel:
…e_to_org`

1) `assign_galaxy_credentials_to_org` and `assign_default_ee_to_org` are not part of the organization data structure.

2) They are set to false if used with the dispatcher role.


Tompage1994 (approved):
Good spot. thanks for the contribution.

I've also spotted the same issue with the arg_spec and am raising a PR

Tompage1994:
Also made a minor change to that table as an extra column is required. I've pushed direct to your PR though and have set to auto-merge if CI passes

PR #784, opened 2024-03-18
Add playbooks to manage configuration files

ivarmu:
# What does this PR do?

Adds two playbooks to ease the management of the configuration files:

- `rename_objects.yaml`: If there's a need of renaming one credential (for example), this playbook can be used to rename the credential itself and update all the Job Templates / Workflows / Organizations... that are using the old credential name to the new one.
- `set_organization.yaml`: If you need to set the organization name for all the objects contained in a path, this playbook will do the job easy.

# How should this be tested?

This can be applied to any path populated with CasC content. Each playbook has an execution example in the last line of the file.

N/A

N/A

### DIFF
--- a/changelogs/fragments/add_playbooks.yml
+++ b/changelogs/fragments/add_playbooks.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - Add two playbooks to simplify management of Configuration as Code files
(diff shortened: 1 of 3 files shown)


silvinux (approved):
LGTM!

Tompage1994 (approved):

PR #791, opened 2024-03-20
Added option to not removing $encrypted$ from filetree_create exported credentials

### LINKED ISSUE #788: Should filetree_create export the credential with '$encrypted$' instead of ''?
### Discussed in https://github.com/redhat-cop/controller_configuration/discussions/786

<div type='discussions-op-text'>

<sup>Originally posted by **ecchong** March 18, 2024</sup>
Why are we replacing the '$encrypted$' string for password with '' empty string in filetree_create template 
 [current_credentials.j2](https://github.com/redhat-cop/controller_configuration/blob/1217522ee9078b6e9a65e6d6e271f2592f248c5b/roles/filetree_create/templates/current_credentials.j2#L15)? 

When importing the credential back into AAP, the '' causes the password to be overwritten with an empty string instead of being ignored when it sees '$encrypted$' string.
</div>

ecchong:
# What does this PR do?
Added option to not removing `$encrypted$` from filetree_create exported credentials.  It keeps the `$encrypted$` text, instead of `` empty string, in the credential files.  It prevents the existing credential value from being overwritten with empty string if the configure files are being used as-is to import back into AAP.

A condition is added to roles/filetree_create/templates/current_credentials.j2 to control the output.

# How should this be tested?
Include `show_encrypted: true` in task invoking `filetree_create`
```yaml
- name: Export AAP credentials for organization {{ org_name }}
  ansible.builtin.import_role:
    name: infra.controller_configuration.filetree_create
  vars:
    organization_filter: "{{ org_name }}"
    output_path: "{{ _objects_output_dir }}"
    flatten_output: false
    input_tag:
      - credentials
    show_encrypted: true
```
Observe password fileds in output files 
```yaml
---
controller_credentials:
  - name: "SSH root"
    description: "Test"
    credential_type: "Machine"
    organization: "Default"
    inputs:
      become_method: ''
      become_username: ''
      ssh_key_data: $encrypted$
      ssh_key_unlock: $encrypted$
      username: root
```

resolves #788

N/A


adonisgarciac:
Hi, I'd prefer keeping $encrypted$ text as default in order to prevent remove things unintentional. @ivarmu what do you think?

ivarmu:
My opinion is that, if having the `$encrypted$` as the sensitive data isn't causing any modifications to the controller objects when applied, it could give a false sensation of having the configuration completely ok, when it's not true at all... in case of a migration or complete disaster, the configuration as code wouldn't be as complete as expected... and the solution should be to populate the sensible data, as with the current approach. 

That said... I'd prefer giving the option to leave the `$encrypted$` if anyone is interested in it, but no for the default value.

Does it makes sense?

ivarmu (approved):
LGTM

PR #794, opened 2024-03-21
fixing all licenses to match and updating min version numbers in meta

djdanielsson:
fixing all licenses to match and updating min version numbers in meta


ivarmu (approved):
LGTM

PR #797, opened 2024-03-25
Forced bool output to be lowercase

### LINKED ISSUE #796: filetree_create should output yamllint compatible truthy values
# Summary

I use the `filetree_create` role to extract the configuration to assist in creating data structures. Unfortunately, the output of boolean values is `True` or `False`. When I run `ansible-lint` or `yamllint` it says these are incorrect truthy values since it prefers lowercase.

# Issue Type

- Bug Report

# Desired Behavior

The output should be valid YAML, at least related to truthiness.

ivarmu:
# What does this PR do?
Force the output of `bool` filter to be lowercase when templating the output files.

Manual and automatic tests.
# How should this be tested?

resolves #796 

N/A

### DIFF
--- a/changelogs/fragments/truthy_values_in_lowcase.yml
+++ b/changelogs/fragments/truthy_values_in_lowcase.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - Adjusted output of boolean filetree_create fields to provide ansible-lint compatible values.
+...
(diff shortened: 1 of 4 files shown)


Tompage1994 (changes requested):
Looks good, I'm happy with this.

Could you please just add a changelog fragment as this will affect the actual output and there's a slight chance someone might depend on the value being uppercase.

```
---
bugfixes:
  - Adjusted output of boolean filetree_create fields to provide ansible-lint compatible values.
...
```

Tompage1994 (approved):

PR #800, opened 2024-03-29
updated filetree_create documentation

bontreger:
Update the documentation for filetree_create with valid tags

Documentation updates to list out all valid tags for filetree_create (consistency issues with singular/plural word choices)

resolves #[number]


djdanielsson (approved):
thanks for the update

PR #802, opened 2024-04-02
forcing verbosity to be read as a string, to fix bug #801

djdanielsson:
# What does this PR do?
fixes bug where verbosity when set to 0 was being dropped 

# How should this be tested?
automated

# Is there a relevant Issue open for this?

resolves #801 

# Other Relevant info, PRs, etc

PR #804, opened 2024-04-04
Fix bug with #802

Tompage1994:
# What does this PR do?
There was an issue with #802 which meant that if verbosity was not set it would cause failure.

# How should this be tested?
CI should now pass

No

resolves problem caused by #802


djdanielsson (approved):
LGTM

PR #806, opened 2024-04-04
Resolve username to ID for filetree_create

Tompage1994:
# What does this PR do?
Fixes an issue where the username was used to lookup roles. This doesn't work if the username is made of entirely numbers. This resolves the username to an ID first.

# How should this be tested?
Create a user `123456` and run the filetree_create against it

resolves AAP-22273 (RH Internal Only)


djdanielsson (approved):
LGTM

PR #807, opened 2024-04-04
Workflows

djdanielsson:
# What does this PR do?
updates some workflows dealing with issues


Tompage1994 (changes requested):
Is this PR ready? 

It doesn't like the changes you've made to the PR template, could you fix please.

djdanielsson:
the PR template I just removed some white spaces so the comments would be right below the headings and make it easier to just write below it all, it isn't important just thought it might be a little better this way, if it isn't rendering correctly I can change it back

djdanielsson:
@Tompage1994  yes it is ready

Tompage1994 (approved):

PR #813, opened 2024-04-15
Update CI versions and set to latest of 22/23/24

Tompage1994:
# What does this PR do?
Update CI versions and set to latest of 22/23/24 versions of AWX

# How should this be tested?
Will need to be merged to be tested but should work as devel has usually worked. (If it fails that could be a legitimate fail anyway...)

No

No


djdanielsson (approved):
LGTM

PR #815, opened 2024-04-19
Fix changed_when statement in all roles to show correct state and dis…

keilr:
# What does this PR do?

In order to display diff output and correct changed state, `changed_when` statement needs to be set properly.

Currently `changed` is just inverted with the `not` operator, which doesn't make sense.

# How should this be tested?
We tested this in our environment. Please let me know if I can help you in adding automated tests.

No

In order to support diff mode I also raised the following issue and pull request:
- https://github.com/ansible/awx/pull/15118
- https://github.com/ansible/awx/issues/14903

These are related to to underlying `awx.awx` (= `ansible.controller`) modules, used in the `infra.controller_configuration` roles

### DIFF
--- a/changelogs/fragments/check_state.yml
+++ b/changelogs/fragments/check_state.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - Fix changed_when statement in all roles to show correct state
+...
--- a/roles/applications/tasks/main.yml
+++ b/roles/applications/tasks/main.yml
@@ -28,7 +28,7 @@
   async: "{{ ansible_check_mode | ternary(0, 1000) }}"
   poll: 0
   register: __applications_job_async
-  changed_when: not __applications_job_async.changed
+  changed_when: "(__applications_job_async.changed if ansible_check_mode else false)"
   vars:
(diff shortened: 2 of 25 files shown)


Tompage1994 (changes requested):
I agree this is a good change. I can see how we haven't spotted it before as well. I think we should add some check_mode testing to the tests to prove this out. If you fancy doing that, then great, otherwise I can raise an issue to fix that afterwards.

The only change I will request here though is that you add a changelog fragment (create a file under `changelogs/fragments` with a `bugfixes` item describing this change. You'll find similar examples in that directory.)

Thanks for your contribution

djdanielsson:
@keilr thank you for contributing back to the community, if you could just add the changeling as Tom requested we can get this merged

Tompage1994 (approved):
I've added the changelog myself and this should be all good now. I'll raise a separate issue to add testing

keilr:
@Tompage1994 @djdanielsson thanks a lot for your feedback and for adding the changelog. Sorry for the delay, I was on vacation 🏖️

PR #816, opened 2024-04-24
Abolish recommending to not validate certs

jpmens:
# What does this PR do?

Fixes a setting which should no longer be recommended in the year 2024

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ The awx.awx or ansible.controller collections MUST be installed in order for thi
   hosts: localhost
   connection: local
   vars:
-    controller_validate_certs: false
+    controller_validate_certs: true
   collections:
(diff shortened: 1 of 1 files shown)


Tompage1994 (approved):
I'm not necessarily sure we are advocating to set validate_certs to false, however I see no reason not to accept this.

Thanks for the contribution

jpmens:
> I'm not necessarily sure we are advocating to set validate_certs to false

I fully understand that, but have as you probably also,  seen people blindly copy/paste config examples.

sean-m-sullivan:
Going to chime in, we should not be setting it to false in examples, however in our tests we use it as the instances we stand up do not have valid certs, which is where it likely bleed over from.

PR #819, opened 2024-05-02
fix(ics): Change from every 4 weeks to first thursday every month

wombelix:
# What does this PR do?

The current .ics file has a recurrence of every four weeks. Based on that, the next meeting would be for example on Thursday 9th May. My understanding is that the meetings actually happen on the first Thursday every month. This PR changes the .ics accordingly. I also adjusted the link in the invite to point to open issues with the office hours label and the topic in the ansible forum.

# How should this be tested?

I did some manual tests and this .ics changed the series in Outlook on Mac to `Occurs the first Thusrday of every month effective 02.05.2024 from 15:00 to 16:00`. I'm not an .ics expert so a second pair of eyes would be helpful :)

no

Last time the file was changed was by https://github.com/redhat-cop/controller_configuration/pull/573

### DIFF
--- a/docs/aap_config_as_code_public_meeting.ics
+++ b/docs/aap_config_as_code_public_meeting.ics
@@ -3,17 +3,19 @@ VERSION:2.0
 PRODID:-//ical.marudot.com//iCal Event Maker
 CALSCALE:GREGORIAN
 BEGIN:VEVENT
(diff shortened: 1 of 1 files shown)


djdanielsson (approved):
LGTM

PR #826, opened 2024-05-13
Fix adhoc role permission name in object diff.

### LINKED ISSUE #799: Ad_hoc value as role name should be adhoc in object diff
# Summary

When using "adhoc" role as a permission, when it is retrieved from the API the value is "Ad Hoc", which gets converted to "ad_hoc"
in https://github.com/redhat-cop/controller_configuration/blob/devel/plugins/lookup/controller_object_diff.py#L233

so it no longer matches with config as code, and object diff attempts to remove it

awx.awx role module expects it as "adhoc", not "ad_hoc"
https://docs.ansible.com/ansible/latest/collections/awx/awx/role_module.html

# Issue Type

- Bug Report

bogdanmuresan:
# What does this PR do?

Fix for https://github.com/redhat-cop/controller_configuration/issues/799

# How should this be tested?

resolves #799 

### DIFF
--- a/changelogs/fragments/adhoc_role_name_object_diff.yml
+++ b/changelogs/fragments/adhoc_role_name_object_diff.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - Fix "adhoc" role permission name in object diff.
+...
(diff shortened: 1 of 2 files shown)


Tompage1994 (approved):
LGTM

PR #836, opened 2024-05-27
feat: support of export single project/workflow/job

przemkalit:
# What does this PR do?
The tasks should allow users to easily export all objects related to a specific project.

Currently, filetree_create retrieves every project and other defined objects, which is not desirable in the following use case scenario.

Use case scenario example for a project: 

1. The user creates a project in the QA AAP environment.
2. The user create many of job templates that use this project
3. The user wants to promote the project with job templates to the Production environment.
4. The user uses filetree_create to create a copy of the project with job templates and push the files to a remote repository.
5. The user uses filetree_read to import the object from the repository.

# How should this be tested?
Manual
1) Provide the variables to the usage of the role:
```
- name: Get project
  hosts: all 
  roles:
      vars:
        project_id: 4
  ```      
2) Execute the playbook and check if the only relevant data were exported.

N/A

Request on the forum, that what we want to achieve: https://forum.ansible.com/t/rfe-develop-new-functionality-for-exporting-configurations-filtered-by-project/5220

### CODE REVIEW

File: roles/filetree_create/README.md [hunk 1]
```diff
@@ -15,6 +15,9 @@ The following variables are required for that role to work properly:
 | `controller_api_plugin` | `ansible.controller` | yes | str | Full path for the controller_api_plugin to be used. <br/> Can have two possible values: <br/>&nbsp;&nbsp;- awx.awx.controller_api             # For the community Collection version <br/>&nbsp;&nbsp;- ansible.controller.controller_api  # For the Red Hat Certified Collection version|
 | `organization_filter` | N/A | no | str | Exports only the objects belonging to the specified organization (applies to all the objects that can be assigned to an organization). |
 | `organization_id` | N/A | no | int | Alternative to `organization_filter`, but specifiying the current organization's ID to filter by. Exports only the objects belonging to the specified organization (applies to all the objects that can be assigned to an organization). |
+| `project_id` | N/A | no | int | Specifiying the project id to filter by. Exports the project belonging to the specified organization. |
+| `job_id` | N/A | no | int | Specifiying the job template id to filter by. Exports the job template belonging to the specified organization. |
```
ivarmu: This can create confusion with `job_template_id ` and `job_id` (a job template run). Could this be renamed to `job_template_id`?

roles/filetree_create/tasks/job_templates.yml
```diff
@@ -1,8 +1,14 @@
...
 - name: "Get current Job Templates from the API"
   ansible.builtin.set_fact:
-    job_templates_lookvar: "{{ query(controller_api_plugin, 'api/v2/job_templates/',
-                                     query_params=(query_params | combine({'organization': organization_id})) if organization_id is defined else query_params,
+    job_templates_lookvar: "{{ query(controller_api_plugin, 'api/v2/job_templates/'        
+                                     query_params=(query_params | combine({'organization': organization_id})) if organization_id is defined and job_id is not defined
+                                     else
+                                     (query_params | combine({'organization': organization_id, 'project_id': project_id})) if organization_id is defined and project_id is defined
+                                     else
+                                     (query_params | combine({'organization': organization_id, 'id': job_id})) if organization_id is defined and job_id is defined
```
ivarmu: If a `job_template_id` is specified, I think it'll belong to only one organization... so it have no sense specifying both inputs... am I wrong? In a similar way, I'm not sure right now if a project can be used from Job templates in different organizations...

ivarmu (changes requested):
I'm adding comments for job templates that also apply to the other modified objects

ivarmu:
.

roles/filetree_create/README.md
przemkalit: Yes, you are right, I will rename it.

przemkalit: I will rename the workflow_job_id to workflow_job_ template_id, as well.

roles/filetree_create/tasks/job_templates.yml
przemkalit: Okay, I think you are right because the database doesn't allow to have a the same id, I didn't think about it, thanks.

roles/filetree_create/tasks/job_templates.yml
```diff
@@ -2,13 +2,13 @@
...
+                                     query_params=(query_params | combine({'organization': organization_id})) if organization_id is defined
                                      else
-                                     (query_params | combine({'organization': organization_id, 'project_id': project_id})) if organization_id is defined and project_id is defined
+                                     (query_params | combine({'project_id': project_id})) if project_id is defined
                                      else
-                                     (query_params | combine({'organization': organization_id, 'id': job_id})) if organization_id is defined and job_id is defined
+                                     (query_params | combine({'id': job_template_id})) if job_template_id is defined
                                      else 
-                                     query_params,                                  d
+                                     query_params,
```
ivarmu: I've been thinking about these lines and there's a problem with them... I tested to confirm my suspects... If `organization_id` is defined, neither `project_id` nor `job_template_id` will be used for the query... The same happens with `project_id`, that will avoid the usage of the `job_template_id`... Maybe the following approach should be better:

```yaml
---
- hosts: localhost
  connection: local
  gather_facts: false
  vars:
    query_params:
      param1: value1
    organization_id: 10
    project_id: 20
    job_template_id: 30
  tasks:
    - debug:
        msg:
          - "Current approach: {{ query_params | combine({'organization': organization_id}) if organization_id is defined
                else
                query_params | combine({'project_id': project_id}) if project_id is defined
                else
                query_params | combine({'id': job_template_id}) if job_template_id is defined
                else
                query_params
             }}"
          - "Proposed approach: {{ query_params | combine({'organization': organization_id} if organization_id is defined else {},
                                                          {'project_id': project_id} if project_id is defined else {},
                                                          {'job_template_id': job_template_id} if job_template_id is defined else {}
                                                         )
                                }}"
...
```

The output of the above is:

```console

TASK [debug] *********************************************************************************************************************************************************************************************************************************
ok: [localhost] => {
    "msg": [
        "Current approach: {'param1': 'value1', 'organization': 10}",
        "Proposed approach: {'param1': 'value1', 'organization': 10, 'project_id': 20, 'job_template_id': 30}"
    ]
}

localhost                  : ok=1    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
```

What do you think?

ivarmu (changes requested):

roles/filetree_create/tasks/job_templates.yml
przemkalit: I must say, when I changed to this, I was aware of it. 

I thought about what you wrote, because as you wrote the organization_id is not mandatory to filter out the projects or job templates. Your proposal is changing it to the way that organization_id can be used with project_id etc. If it would be better for the collection I could change to your suggestion.

ivarmu: Yeah, I agree that is changing my first review point of view... but I think it's more congruent like this, as if anyone provides the `job_template_id` alongside with `organization_id` and/or `project_id`, the expected, I think, is to have exported only the `job_template_id` that is associated with the already provided `project_id` from the `organization_id` organization... :-) sorry for being dizzy and thanks in advance for the changes :-)

przemkalit: No problem :-), I will add this modification, thanks for the feedback :-).

ivarmu (approved):
LGTM

ivarmu (changes requested):
Please, fix the errors listed by the linter

roles/filetree_create/tasks/job_templates.yml
```diff
@@ -1,8 +1,10 @@
 ---
 - name: "Get current Job Templates from the API"
   ansible.builtin.set_fact:
-    job_templates_lookvar: "{{ query(controller_api_plugin, 'api/v2/job_templates/',
-                                     query_params=(query_params | combine({'organization': organization_id})) if organization_id is defined else query_params,
+    job_templates_lookvar: "{{ query(controller_api_plugin, 'api/v2/job_templates/'
```
ivarmu: Missing a comma at the end of the line

roles/filetree_create/tasks/projects.yml
```diff
@@ -2,10 +2,12 @@
...
   ansible.builtin.set_fact:
     projects_lookvar: "{{ query(controller_api_plugin, 'api/v2/projects/',
-                                query_params=(query_params | combine({'organization': organization_id})) if organization_id is defined else query_params,
+                                query_params=(query_params | combine({'organization': organization_id} if organization_id is defined else {},
+                                {'id': project_id} if project_id is defined else {},
+                                {'name': project_name} if project_name is defined else {})),
                                 host=controller_hostname, oauth_token=controller_oauthtoken, verify_ssl=controller_validate_certs,
                                 return_all=true, max_objects=query_controller_api_max_objects)
-                       }}"
+                      }}"
```
ivarmu: Please, add the space that has been removed to maintain the alignment. Is not important, I know, sorry for my OCD :-)

ivarmu (changes requested):

roles/filetree_create/tasks/job_templates.yml
przemkalit: Oh sorry, I think I am blind for commas :|.

ivarmu (approved):
LGTM

roles/filetree_create/tasks/workflow_job_templates.yml
```diff
@@ -2,7 +2,9 @@
 - name: "Get current Workflow Job Templates from the API"
   ansible.builtin.set_fact:
     workflow_job_templates_lookvar: "{{ query(controller_api_plugin, 'api/v2/workflow_job_templates/',
-                                              query_params=(query_params | combine({'organization': organization_id})) if organization_id is defined else query_params,
+                                              query_params=(query_params | combine({'organization': organization_id} if organization_id is defined else {},
+                                              {'project_id': project_id} if project_id is defined else {},
```
ivarmu: The API is not accepting this parameter:

```console
TASK [infra.controller_configuration.filetree_create : Get current Workflow Job Templates from the API] **************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'awx.awx.controller_api'. Error was a <class 'ansible.errors.AnsibleError'>, original message: Failed to query the API: WorkflowJobTemplate has no field named 'project_id'. Failed to query the API: WorkflowJobTemplate has no field named 'project_id'"}
```

ivarmu (changes requested):

roles/filetree_create/tasks/workflow_job_templates.yml
przemkalit: Ok, I've removed it. I was pretty sure that it was related, but I was mistaken.

ivarmu (approved):
LGTM

silvinux (changes requested):
LGTM, but a changelog fragment file needs to be created: changelogs/fragments. Something similar to: 

---
Minor Changes:
  - filetree_create able to filter by project_id, workflow_job_template_id or job_template_id

https://github.com/redhat-cop/controller_configuration/blob/devel/changelogs/config.yaml

przemkalit:
Ok, I've added the missing changelog, anything more to add?

silvinux (approved):
LGTM

PR #848, opened 2024-06-27
allow max_hosts to be set to 0 (infinite) if controller_configuration_organizations_enforce_defaults is true

w4hf:
# What does this PR do?

If max_hosts > 0 on the Controller, the user cannot set it to zero using the configuration as code.
This tweak solve this issue if `controller_configuration_organizations_enforce_defaults` is set to `true`.

# How should this be tested?

Create an organization with max_hosts > 0 on the controller.
Try to set max_hosts to zero with the configuration as code. This will fail.
After applyin this PR set `controller_configuration_organizations_enforce_defaults` to `true` and re-run the configuration as code. The max_hosts will now be set to zero on the Controller.

This was tested on :
ansible [core 2.16.7]
python version = 3.11.8 (main, Feb 26 2024, 06:51:39) [GCC 8.5.0 20210514 (Red Hat 8.5.0-18)] (/opt/python/3.11/bin/python3.11)
  jinja version = 3.1.4

resolves #[number]


sean-m-sullivan (approved):
Thanks for the change, Looks great to me!

PR #859, opened 2024-07-09
instance group skip option when creating orgs

### LINKED ISSUE #858: No way to create and affect an Instance Group to an Organization in on go
Using latest published 2.7.1

As far as I can see, there's no way to bootstrap an AAP2 instance with an ad-hoc Container group affected to an Organization.

The reason is the dependency cycle: Organization -> Instance Group -> Credential -> Organization.

There's this existing mecanism in the Organization Controller, which deactivate the ee and cred affectation when using the vars `assign_galaxy_credentials_to_org` and `assign_default_ee_to_org`, which solves a similar issue.

Would it be possible to have similar var for Instance Groups?

djdanielsson:
# What does this PR do?

attempt to add the option to skip instance_group assignment to org until it can be created.

# How should this be tested?

manual at first maybe add automated later

resolves #858 


djdanielsson:
untested

adonisgarciac:
Hi @djdanielsson, I think you should add assign_instance_groups_to_org control var to dispatch role:

https://github.com/redhat-cop/controller_configuration/blob/devel/roles/dispatch/tasks/main.yml#L12-L15
https://github.com/redhat-cop/controller_configuration/blob/devel/roles/dispatch/defaults/main.yml#L14-L17
https://github.com/redhat-cop/controller_configuration/blob/devel/roles/dispatch/defaults/main.yml#L42-L47

sean-m-sullivan (approved):

djdanielsson:
the instance groups are created before the orgs in dispatch so I do not think it is needed @adonisgarciac

adonisgarciac:
you're right @djdanielsson. Then, I guess these changes are not solving #858. I think the problem in that issue is that they want to add a Instance Group/ container group (with a Credential for k8s/ocp)  and Credentials are created after instance group and after organizations:

"The reason is the dependency cycle: Organization -> Instance Group -> Credential -> Organization."

I mean, the problem is not when organizations with IG are created (because IG are created before). If I'm not mistaken, the actual issue is when a IG with credential is created but that credential is not present yet:

https://github.com/redhat-cop/controller_configuration/blob/devel/roles/instance_groups/tasks/main.yml#L6

djdanielsson:
there may be a need to change when instance groups are created, but that is not related to this. not everyone uses dispatch nor will it work for every configuration. The request was to add an option to skip instance groups when creating orgs which this does

nodje:
Great! 
thanks @djdanielsson 

BTW, any reason why 2.8.1 is still not available in galaxy.ansible.com 2 months release ?

PR #862, opened 2024-07-10
Idea of managing errors while creating the objects in controller

przemkalit:
# What does this PR do?
This PR proposes a method to manage errors while creating objects in the controller.

Background: I was testing the collection in our environment and encountered an annoying issue. While restoring objects from files created with filetree_create, the dispatch role stopped every time it encountered a problem, such as when an object was missing something (e.g., a playbook could not be found in the project). 

I came up with a block and rescue directive for the async task. The rescue block extracts errors from the result file and puts them into a list of dicts, which is then set as an artifact.

If you like this idea, I will introduce it to every object in the collection. However, if you have a different approach to this problem, I am open to suggestions.

# How should this be tested?
1) Create job_templates which playbook does not exist in the project in target controller.
2) Export with filetree_create.
2) Execute playbook:
```yaml
- name: Restore objects
   hosts: all 
   tasks:
     - name: Include vars from path directory
       ansible.builtin.include_vars:
         dir: '/path'
         extensions:
           - "yml"
           - "yaml"
      
     - ansible.builtin.include_role: infra.controller_configuration.filetree_read
      
     - ansible.builtin.include_role: infra.controller_configuration.dispatch
```

N/A

N/A


djdanielsson:
It's an interesting idea, we will have to do quite a bit of testing to ensure this doesn't break anything. When I have some time I will try it out and give more detailed feedback

roles/.DS_Store
djdanielsson, line 1: delete this file please

przemkalit: oh, sorry I try to watch over this file created by VSCode, removed.

roles/job_templates/tasks/async.yml
```diff
@@ -0,0 +1,29 @@
...
+      register: __job_templates_job_async_result
+      until: __job_templates_job_async_result.finished
+      retries: "{{ controller_configuration_job_templates_async_retries }}"
+      delay: "{{ controller_configuration_job_templates_async_delay }}"
+
+  rescue:
+    - name: "Load error details"
+      ansible.builtin.include_vars:
+        file: "{{ __job_templates_job_async_result.results_file }}"
+        name: error_data
```
Tompage1994: Can you rename the internal vars here to be prepended with `__` to fit in with existing coding practices.

e.g. `name: __error_data`
e.g. `__templates_error_list `

roles/job_templates/tasks/async.yml
Tompage1994, line 1: We are missing the `...` at the end of the file. Please ensure that the linting is passing before we can merge this

Tompage1994 (changes requested):
Overall I think this is good and may be useful. I don't think it will slow down execution and I believe it to be scalable.

I haven't tested this on a live system.

It would be good though if you could have a go at adding into our [CI testing](https://github.com/redhat-cop/controller_configuration/blob/devel/tests/configure_controller.yml). Get a JT which will fail and assert that the output is correct

roles/job_templates/tasks/async.yml
przemkalit: Ok I will work on the JT.

przemkalit:
> Overall I think this is good and may be useful. I don't think it will slow down execution and I believe it to be scalable.
> 
> I haven't tested this on a live system.
> 
> It would be good though if you could have a go at adding into our [CI testing](https://github.com/redhat-cop/controller_configuration/blob/devel/tests/configure_controller.yml). Get a JT which will fail and assert that the output is correct

Ok I will work on JT.

przemkalit:
Please let me know if the test provided by me is sufficient or should I make some changes.

Tompage1994:
> Please let me know if the test provided by me is sufficient or should I make some changes.

The testing you've done looks good but is unfortunately in the wrong place. Could you add it to the following section:

https://github.com/redhat-cop/controller_configuration/blob/devel/tests/configure_controller.yml

Tompage1994 (approved):
Thanks, that looked better.

I've just done a quick refactor of the test and pushed it to this PR to simplify it a little bit. If you're happy with that then I'm happy to merge this

przemkalit:
Thanks for the refactor, so as you approved this kind of error handling, could I add this change to the workflows, or should I added with a new PR?

Tompage1994:
@przemkalit Ok, I've made another change to the testing because we don't set the variables as we are setting stat rather than vars.

There is a bigger conversation to be had here about how we want to handle these errors, whether we want to fail the playbook by default (I think we may want to add in a variable here to set to fail on error. As it is this will continue on error)

In terms of creating the same on other roles, I would hold off for now. It may all go through this PR or could be separate but we need to decide how we want this to go for now. This will now sit here for a bit whilst we internally discuss and decide on direction.

Thanks for the contribution

przemkalit:
Hi, I've added small change which is related to existence of variables in eg. `extra_vars`.

roles/job_templates/tasks/main.yml
```diff
@@ -84,13 +84,8 @@
...
 
-- name: "Managing Controller Job Templates | Wait for finish the Job Templates management"
-  ansible.builtin.async_status:
-    jid: "{{ __job_templates_job_async_result_item.ansible_job_id }}"
-  register: __job_templates_job_async_result
-  until: __job_templates_job_async_result.finished
-  retries: "{{ controller_configuration_job_templates_async_retries }}"
-  delay: "{{ controller_configuration_job_templates_async_delay }}"
+- name: "Managing Controller Job Templates | include async file"
+  ansible.builtin.include_tasks: "{{ role_path }}/tasks/async.yml"
```
djdanielsson, line 88: do you really have to give it the path like this instead of `ansible.builtin.include_tasks: async.yml`

przemkalit: I've changed it because the CI testing was failing. The playbook tried to include the async file from the test playbook path instead of the role path.

djdanielsson:
update the branch and I will merge it

djdanielsson:
@Tompage1994 ready to merge it?

djdanielsson:
@Tompage1994 reminder, are you good with this now?

Tompage1994:
> @Tompage1994 reminder, are you good with this now?

Yes I'm good with it. We will want to look into replicating this across all the roles

PR #864, opened 2024-07-12
Omit id for objects during export with filtree_create

przemkalit:
# What does this PR do?
Propose a solution that allows exporting objects without IDs.

Background: In our environment, the inclusion of IDs in the filenames is causing a slight mess in our GitHub repositories. Imagine the following situation:

1. I exported a project with job templates from the DEV environment.
2. I imported it into the TEST environment.
3. I created a new job template in the TEST environment for this project.
4. I exported the project with job templates again, but this time from the TEST environment.
5. I discovered in the repository that there are files of job templates with both DEV and TEST IDs.

# How should this be tested?
```yaml
- name: Get objects
  hosts: all 
  roles:
      vars:
        omit_id: true
```

N/A
N/A

### DIFF
--- a/changelogs/fragments/filetree_create_omit_id.yml
+++ b/changelogs/fragments/filetree_create_omit_id.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - filetree_create able to create files without id values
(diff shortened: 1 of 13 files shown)


ivarmu (changes requested):
LGTM.

Could you add an extra space just after `else ''` to look consistent with other variables, resulting like `else '' }}`?

Another point is to add the new input variable in https://github.com/redhat-cop/controller_configuration/blob/devel/roles/filetree_create/README.md and in https://github.com/redhat-cop/controller_configuration/blob/devel/roles/filetree_create/meta/argument_specs.yml

przemkalit:
Ok, I've added the spaces, spotted missing id for teams and updated the documentation.

ivarmu (approved):
LGTM

Issue #831, opened 2024-05-21
CI Failing


Tompage1994:
The CI is failing to run the tests on all versions. Need to explore why this is. It waits for the stack to be available but it never is. There is no obvious error on the build

Tompage1994:
Underlying issue: https://github.com/docker/docker-py/issues/3256

Tompage1994:
Now just an issue with project updates and inventory source updates not working properly.

PR #866, opened 2024-07-15
Add manual release process

Tompage1994:
# What does this PR do?

Adds a manual release process for when the CI is broken to enable admins to push through releases which are of sufficient quality/

# How should this be tested?

It's the same pipeline, just with the CI removed and an input for the version to release.

No

#831 related

### DIFF
--- a/.github/workflows/release_manual.yml
+++ b/.github/workflows/release_manual.yml
@@ -1,0 +1,40 @@
+---
+# This workflow action will run pre-commit, which will execute ansible and yaml linting
+# See .pre-commit-config.yaml for what hooks are executed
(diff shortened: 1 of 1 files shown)


djdanielsson:
coming back to this, to be clear this can be forced if needed but has to be triggered differently then normal way we do a release so we can attempt a release if it fails and we want to force it we can then use this?

djdanielsson:
want to confirm before i merge @Tompage1994

Tompage1994:
> want to confirm before i merge @Tompage1994

Yes this is simply an additional workflow to run manually if needed when CI is broken

PR #867, opened 2024-07-15
Remove encrypted from surveys

przemkalit:
# What does this PR do?
Recently, I've discovered that `$encrypted$` is exported with the job template/workflow when a password is set in the survey field. During the import, this causes an issue, and the survey is not imported.

# How should this be tested?
1) Create the job template with survey, that has a password field, and set password in default choice field.
2) Export the job template with `filetree_create`.
3) There should not be any `$encrypted$` in the export.

N/A
N/A

### DIFF
--- a/changelogs/fragments/filetree_create_survey_password_removal.yaml
+++ b/changelogs/fragments/filetree_create_survey_password_removal.yaml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - filetree_create able to remove $encrypted$ while exporting job template and workflow
--- a/roles/filetree_create/templates/current_job_templates.j2
+++ b/roles/filetree_create/templates/current_job_templates.j2
@@ -95,7 +95,7 @@ controller_templates:
     survey_enabled: {{ current_job_templates_asset_value.survey_enabled | bool | lower }}
 {% set survey_spec_contents = query(controller_api_plugin, current_job_templates_asset_value.related.survey_spec,
                                     host=controller_hostname, oauth_token=controller_oauthtoken, verify_ssl=controller_validate_certs)[0] |
-                                    from_yaml | to_nice_yaml(indent=2,width=500) | regex_replace("\n\n[ ]*", "\\\\n") | indent(width=6, first=False) | replace("'{{", "!unsafe \'{{") | replace("^$", "")
+                                    from_yaml | to_nice_yaml(indent=2,width=500) | regex_replace("\n\n[ ]*", "\\\\n") | indent(width=6, first=False) | replace("'{{", "!unsafe \'{{") | replace("^$", "") | replace("$encrypted$", "\'\'")
 -%}
 {% if current_job_templates_asset_value.related.survey_spec is defined and survey_spec_contents | length > 3 %}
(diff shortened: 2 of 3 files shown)


Tompage1994 (commented):
If I'm reading this correctly this will set the value to `''`.

This doesn't feel like a good idea because this will mean we blank out the password. At least having `$encrypted$` means people know where they need to replace values.

An alternate strategy here would be to raise a PR against the lookup plugin to add a flag to not return values with `$encrypted$` and instead omit them from the results

przemkalit:
Okay the error looks like this:     
`"Failed to update survey: $encrypted$ is a reserved keyword, may not be used for new default in position 13."`
what if I just remove the dollars and leave only encrypted?

Tompage1994:
> Okay the error looks like this: `"Failed to update survey: $encrypted$ is a reserved keyword, may not be used for new default in position 13."` what if I just remove the dollars and leave only encrypted?

Ah I see. I hadn't read properly that this was within surveys so setting a default of being blank may be our best solution here as you've done. I'll approve for now but I'll request someone else takes a look

Tompage1994 (approved):

ivarmu (changes requested):
The proposed changes gives undesired results:

```
$ansible-playbook -i inventory.yaml example.yaml

PLAY [play all] *****************************************************************************************************************************************************************************************************************************

TASK [debug] ********************************************************************************************************************************************************************************************************************************
ok: [hosta] => {
    "msg": [
        "REDHAT_PASSWORD: \"$encrypted$\"\n",
        "REDHAT_PASSWORD: \"\\'\\'\"\n"
    ]
}

PLAY RECAP **********************************************************************************************************************************************************************************************************************************
hosta                      : ok=1    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
```

Instead of `replace("$encrypted$", "\'\'")`, it should be `replace("$encrypted$", "")`

```
ansible-playbook -i inventory.yaml example.yaml

PLAY [play all] *****************************************************************************************************************************************************************************************************************************

TASK [debug] ********************************************************************************************************************************************************************************************************************************
ok: [hosta] => {
    "msg": [
        "REDHAT_PASSWORD: \"$encrypted$\"\n",
        "REDHAT_PASSWORD: \"\"\n"
    ]
}

PLAY RECAP **********************************************************************************************************************************************************************************************************************************
hosta                      : ok=1    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```

przemkalit:
Okay, I left only the empty double quotes.

ivarmu (approved):
LGTM

adonisgarciac:
I don't like this change at all @Tompage1994 @ivarmu. If you export a WF or a JT with any encrypted data either in a survey or in the WF/JT itself, that value should be put in place vaulted or it has to be omitted. 

With this change, if someone exports a JT/WF, default password of survey will be replaced for nothing which could ends up in an unexpected situation. 

I'd prefer to have to review things manually like passwords after export objects with filetree_create rather than leave filetree_create exports objects with false values of some parameters

PR #871, opened 2024-07-16
Idea of adding objects that are missing in the target controller

przemkalit:
# What does this PR do?
Another idea was born while implementing CasC in our environment. The code is related to job templates but can be easily added to other object types.

What if someone wants to restore only non-existing objects on the target controller?

So, I found a variable called `controller_state` to create a condition to only add objects that are missing. However, I noticed that it took forever to finish because I was still looping over the object list. Additionally, a second issue appeared: if the same object was different in some way, the object was duplicated.

I came up with the idea to get all objects and filter out the ones that already exist.

What do you think about this change? I know that this variable `controller_state` is making problem with the roles, because role can only be present or be absent, but I believe it can be easily resolved.

# How should this be tested?
1) Create job templates in source controller.
2) Export job templates with filetree_create.
3) Import job templates into target controller.
```yaml
---
- name: Playbook to configure ansible controller post installation
  hosts: localhost
  connection: local
  vars:
    controller_hostname: ansible-controller-web-svc-test-project.example.com
    controller_username: admin
    controller_password: changeme
    controller_state: exists
  pre_tasks:
    - name: Include vars from controller_configs directory
      ansible.builtin.include_vars:
        dir: ./yaml
        ignore_files: [controller_config.yml.template]
        extensions: ["yml"]
  roles:
    - infra.controller_configuration.dispatch
```

N/A

N/A

### DIFF
--- a/roles/job_templates/tasks/main.yml
+++ b/roles/job_templates/tasks/main.yml
@@ -1,4 +1,32 @@
 ---
+- name: Verify Job Templates
+  when: controller_templates is defined and (controller_state is defined and controller_state == 'exists')
+  block:
+    - name: 'Retrieve job templates from controller'
+      ansible.builtin.set_fact:
+        __t_controller_job_templates: >-
+          {{
+            query(
+              controller_api_plugin,
+              'api/v2/job_templates/',
+              host=controller_hostname,
+              oauth_token=controller_oauthtoken,
+              verify_ssl=controller_validate_certs,
+              return_all=true,
+              max_objects=query_controller_api_max_objects
+            ) |
+            map(attribute='name') |
+            ansible.builtin.flatten
+          }}
+
+    - name: 'Extract job templates which are not defined in controller'
+      ansible.builtin.set_fact:
+        job_templates: >-
+          {{
+            controller_templates |
+            rejectattr('name', 'in', __t_controller_job_templates)
+          }}
+
 - name: "Managing Controller Job Templates"
   job_template:
     name:                                 "{{ __controller_template_item.name | mandatory }}"


adonisgarciac:
If I'm not mistaken, I can see a problem in this PR. If someone changes an object  manually in the controller, it will persist forever. Desired state of that object defined as code won't be applied and the manual change of it will keep

przemkalit:
Yes, it will stay the same, but I believe it could somehow be resolved. In the current state, the collection is not reliable when the environment is full of templates, workflows, and other objects. There must be some way to validate what is in the target controller, because if there were an emergency to restore what is missing on the target controller, it would take the same amount of time as restoring the full export. I am not insisting on my solution, but it would be good to think about how to do this the proper way.

Tompage1994 (commented):
I'd argue this is probably the function of the object_diff plugin? 

The issue I see here (if I'm reading correctly) is that we will never be able to update items at all here which doesn't feel quite right.

adonisgarciac:
Ok. Then I'd add some control like you proposed:

  when: controller_templates is defined and (controller_state is defined and controller_state == 'exists')

However controller_state == 'exists' means that it should check that the object exists but not modify or create it. So I'd change that for another control variable like only_new_objects. 

With that everything will work as for now, But if you add that variable as true, it will create only new objects without update the ones are already present, right? This has not to be the default behavior because as a I said, it won't update the objects as you've defined them as code.

przemkalit:
> I'd argue this is probably the function of the object_diff plugin?
> 
> The issue I see here (if I'm reading correctly) is that we will never be able to update items at all here which doesn't feel quite right.

I saw this plugin, but I didn't see it being used. However, I've just read the README, and this could be helpful. I will close this PR. Thanks for pointing it out! :)

PR #873, opened 2024-07-17
Add folder as list support while keeping backward compatibility

lennysh:
# What does this PR do?
This will allow users to specify more than one folder for each type.  This idea came from my wanting to have some credentials in all environments (without duplicating them in the folder structure), but also have credentials that are only in `DEV` or `PROD`.

# How should this be tested?
1. Update the existing filetree_read folder vars as a `list` instead of just a `string`.
     - There is no requirement to convert any or all of the vars.  You can easily mix the two formats as the new tasks consider this.
3. Execute playbook

# Example
```yaml
filetree_controller_credentials:
  - "{{ dir_orgs_vars }}/{{ orgs }}/env/common/controller_credentials.d/"
  - "{{ dir_orgs_vars }}/{{ orgs }}/env/{{ env }}/controller_credentials.d/"
```
Instead of:
```yaml
filetree_controller_credentials: "{{ dir_orgs_vars }}/{{ orgs }}/env/{{ env }}/controller_credentials.d/"
```
But both will still work!

N/A

N/A


roles/filetree_read/tasks/applications.yml
```diff
@@ -1,15 +1,26 @@
...
+  ansible.builtin.set_fact:
+    __filetree_controller_applications: "{{ [filetree_controller_applications] if filetree_controller_applications is string else filetree_controller_applications }}"
+
+- name: "Get list of files"
   ansible.builtin.find:
-    paths: "{{ filetree_controller_applications }}"
+    paths: "{{ __filetree_controller_applications_item }}"
     file_type: file
     recurse: true
+  loop: "{{ __filetree_controller_applications }}"
```
ivarmu, line 11: I believe that these changes are not necessary at all, as the `paths` parameter for the `ansible.builtin.find` module is already accepting a list of paths (https://docs.ansible.com/ansible/latest/collections/ansible/builtin/find_module.html#parameter-paths).

So, the only needed change is the variables' definitions at `roles/filetree_read/defaults/main.yml`, nothing else.

Would you mind @lennysh to remove the unneeded changes?

Thanks.

roles/filetree_read/tasks/credential_input_sources.yml
```diff
@@ -1,15 +1,26 @@
 ---
-- name: "Get list of files inside {{ filetree_controller_credential_input_sources }}"
+- name: Determine if filetree_controller_credential_input_sources is a string or list
+  ansible.builtin.set_fact:
+    __filetree_controller_credential_input_sources: "{{ [filetree_controller_credential_input_sources] if filetree_controller_credential_input_sources is string else filetree_controller_credential_input_sources }}"
+
+- name: "Get list of files"
   ansible.builtin.find:
-    paths: "{{ filetree_controller_credential_input_sources }}"
+    paths: "{{ __filetree_controller_credential_input_sources_item }}"
```
ivarmu, line 8: I believe that these changes are not necessary at all, as the `paths` parameter for the `ansible.builtin.find` module is already accepting a list of paths (https://docs.ansible.com/ansible/latest/collections/ansible/builtin/find_module.html#parameter-paths).

So, the only needed change is the variables' definitions at `roles/filetree_read/defaults/main.yml`, nothing else.

Would you mind @lennysh to remove the unneeded changes?

Thanks.

ivarmu (changes requested):
I've commented the first two files, but the same applies to all other roles' tasks.

roles/filetree_read/tasks/credential_input_sources.yml
lennysh: Well, that's embarrassing.  Pretty sure I had tested that, and it didn't work, but I just tried it with the changes I had already made to my vars and switched back to the previous version and it does work now.  I must of had a typo the first time I tested.  I don't even feel like this PR is needed at all now.

ivarmu: No problem @lennysh, all of us have been in the same situation sometime :-). Anyway, thanks for your contribution spirit ;-)

It is already documented at https://github.com/redhat-cop/controller_configuration/tree/devel/roles/filetree_read#organization-and-environment-variables...

Feel free to close the PR if you agree.

PR #874, opened 2024-07-19
updating ansible version to supported version 2.15.0

djdanielsson:


Tompage1994:
Do we really need to be restricting this to run only with supported versions? I get that we may not support anything broken with old ansible versions but I don't agree we should actually be stopping people from using older ansible?

branic:
I think it is a matter of what versions the CI is testing and how many versions the maintainers want to support. Setting the minimum version is a way to say "this is the earliest version that we test".

djdanielsson:
it doesn't stop people from using it, I think it will just give a warning that they are using a version we do not support which is what we want to be clear about

adonisgarciac (approved):
LGTM.

djdanielsson:
It looks like the pipeline is failing again, Im forcing merging this

PR #876, opened 2024-07-22
Issue with incorrectly exported approval role name

przemkalit:
# What does this PR do?
While restoring objects with dispatch, I've encountered an issue: the role `approval` is called `approve` in the controller, but the API requires `approval`. As a result, when the `filetree_create` exports the roles, it uses the wrong role name. I've added an if statement to check if the role name is `approve`, and if it is, it uses the `approval` role name.

# How should this be tested?
1) Create workflow job template.
2) Assign approve role to team.
3) Export objects with `filetree_create`.
4) Import objects with `filetree_read` and `dispatch`.

N/A

N/A


ivarmu (approved):
LGTM

PR #883, opened 2024-07-26
fix: missing escape special characters

### LINKED ISSUE #882: Missing escaping special characters in description.
## Summary

Recently, I tried to restore job templates in our environment, but an error is showing up due to some special characters in the description.

I've managed to fix it manually in the files in two ways:

original yaml fragment:
```yaml
...
    description: "XXX report \\10.10.10.128\something$\whatever\yyy\zzz"
...

```
The first solution:
```yaml
...
    description: "XXX report \\10.10.10.128\\something\\$\\whatever\\yyy\\zzz"
...
```

The second solution:
```yaml
...
    description: 'XXX report \\10.10.10.128\something$\whatever\yyy\zzz'
...

```

But maybe you would have better solution for this. 

## Issue Type

- Bug Report

Devel affected.

## OS / ENVIRONMENT

N/A

## Desired Behavior

Export objects with proper escaped special characters

## Actual Behavior

I've ran an import script and include_vars module failed with this error:

```console (error)
"We were unable to read either as JSON nor YAML, these are the errors we got from each:
JSON: Expecting value: line 1 column 1 (char 0)
Syntax Error while loading YAML.
 found unknown escape character
The error appears to be in '/repo_dir/controller_objects/control_vars/controller_job_templates/job_scheduled.yaml':
 line 4, column 70, but maybe elsewhere in the file depending on the exact syntax problem.
The offending line appears to be:
 name: \"job_scheduled\"
 description: \"XXX \\\\10.10.10.128\\something$\\whatever\\yyy\\zzz\"\n          
}
```

## STEPS TO REPRODUCE
1) Create job template with description: `XXX report \\10.10.10.128\something$\whatever\yyy\zzz`.
2) Export Job templates with filetree_create.
3) Try to Include_vars with exported files.

Edit: double quotes present in description can break things as well.

przemkalit:
# What does this PR do?
fix #882 

# How should this be tested?
explained in a bug

resolves #882

N/A

### DIFF
--- a/changelogs/fragments/filetree_create_escape_special_characters.yaml
+++ b/changelogs/fragments/filetree_create_escape_special_characters.yaml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - filetree_create export object description while escaping characters like `$`,`/`` and `"`
--- a/roles/filetree_create/templates/current_applications.j2
+++ b/roles/filetree_create/templates/current_applications.j2
@@ -3,7 +3,7 @@
 controller_applications:
 {% endif %}
   - name: "{{ current_applications_asset_value.name }}"
-    description: "{{ current_applications_asset_value.description | regex_replace('([\\\"])', '\\\\1') }}"
+    description: "{{ current_applications_asset_value.description | regex_replace('([\\\\\"$])', '\\\\\\1') }}"
     organization: "{{ current_applications_asset_value.summary_fields.organization.name | default(organization if organization != 'ORGANIZATIONLESS' else 'ToDo: The application \'' + current_applications_asset_value.name + '\' must have an organization') }}"
     authorization_grant_type: "{{ current_applications_asset_value.authorization_grant_type }}"
     redirect_uris: "{{ current_applications_asset_value.redirect_uris }}"
--- a/roles/filetree_create/templates/current_applications.j2
+++ b/roles/filetree_create/templates/current_applications.j2
@@ -3,7 +3,7 @@
 controller_applications:
 {% endif %}
   - name: "{{ current_applications_asset_value.name }}"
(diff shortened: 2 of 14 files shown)


adonisgarciac (changes requested):
I'm testing this change and it is transforming:
`
XXX report \\10.10.10.128\something$\whatever\yyy\zzz
`
to 

`
XXX report \\\\\\\\10.10.10.128\\\\something\\$\\\\whatever\\\\yyy\\\\zzz
`

which ends up applying the following description:

`
XXX report \\\\10.10.10.128\\something\$\\whatever\\yyy\\zzz
`

which is not the original one.

After some tests, I think we don't need to escape $ and we should apply something like:

`
regex_replace('([\\\"])', '\\\\1')
`

which transform the description to 

`
XXX report \\\\10.10.10.128\\something$\\whatever\\yyy\\zzz
`

and it will end up applying the origin description:

`
XXX report \\10.10.10.128\something$\whatever\yyy\zzz
`

which is the original one.

przemkalit:
Ok, I've changed the regex, thanks for the testing

adonisgarciac (changes requested):
Sorry, I've tested again and it seemed to work but only for \. For " symbol it didn't work.

I think the right expression would be: 

regex_replace('([\\\"])', '\\1')

Can you test it?

przemkalit:
> Sorry, I've tested again and it seemed to work but only for . For " symbol it didn't work.
> 
> I think the right expression would be:
> 
> regex_replace('([\"])', '\1')
> 
> Can you test it?

Sorry for late response, but this regex doesn't work and issue with loading yaml happens again.

przemkalit:
I've reverted the regex to the proposed by me, due to my previous comment.

przemkalit:
Hi, I've discovered that issue is related only when the files are included by the include_vars module instead of filetree_read, so I will close this PR.

PR #884, opened 2024-07-31
Roles import optimatization by modifying the export + 2 minor bugs

przemkalit:
# What does this PR do?
This PR changes the way roles are exported in the current version of the collection. Previously, every single permission role was added as a separate entry. I've modified it so that objects related to a role are matched with that role. As a result, the entry for a role is now presented as follows.

```yaml
controller_roles:
  - team: "Y"
    projects:
      - "A"
      - "B"
      - "C"
      - "D"
      - "E"
      - "F"
    role: "update"
  - team: "Y"
    projects:
      - "A"
      - "B"
      - "C"
      - "D"
      - "E"
      - "F"
    role: "read"

```
Thanks to that, the number of requests sent to the API is decreasing.

Let me know what you think. If you have any suggestions, I would be grateful.

Additionally, there are two changes: one related to the approval role, which I fixed for teams but not for users, and another issue with the `job template admin`, `project admin`, and similar roles. These roles are exported without underscores, which prevents them from being imported.

# How should this be tested?
1) Use filetree_create to create export.
2) Use filetree_read to load export.
3) Use dispatch to import.

N/A
#876


djdanielsson:
thoughts?
@silvinux
@ivarmu
@adonisgarciac

ivarmu (approved):
Good idea! LGTM

djdanielsson:
if you update the branch I will merge this

adonisgarciac (approved):
LGTM

PR #886, opened 2024-08-01
Feature role synchronization. 

### LINKED ISSUE #9: Roles synchronisation
N/A

I've discovered that while restoring the roles, they are not in sync.

For example, if there is no role set as Admin for project `X` in the `A` controller, but there is an Admin role for project `X` in the `B` controller, the Admin role should be removed from the `B` controller when an import is requested, as it is not included in the export created by `filetree_create`.

I've tested a solution to fetch the entire list of roles and compare it to the provided list, but it takes too much time to be reliable. Therefore, I am opening this request in case you have an idea on how to approach this. If there is something I didn't notice in the collection, please let me know.

**Additional context**
N/A

### LINKED ISSUE #872: Roles synchronisation
N/A

I've discovered that while restoring the roles, they are not in sync.

For example, if there is no role set as Admin for project `X` in the `A` controller, but there is an Admin role for project `X` in the `B` controller, the Admin role should be removed from the `B` controller when an import is requested, as it is not included in the export created by `filetree_create`.

I've tested a solution to fetch the entire list of roles and compare it to the provided list, but it takes too much time to be reliable. Therefore, I am opening this request in case you have an idea on how to approach this. If there is something I didn't notice in the collection, please let me know.

**Additional context**
N/A

przemkalit:
# What does this PR do?
In some ways implement redhat-cop/aap_configuration_extended#9.

# How should this be tested?
1) Create the export with filetree_create
2) Use filetree_read and dispatch with a new variable
```yaml
- name: Sync roles
  hosts: all 
  tasks:
    - name: Include vars from control_vars directory
      ansible.builtin.include_vars:
        dir: "{{ dir_orgs_vars }}"
        extensions:
          - "yml"
          - "yaml"

  roles:
    - role: infra.controller_configuration.filetree_read
    - role: infra.controller_configuration.dispatch
      vars:
        sync_roles: true
```

resolves redhat-cop/aap_configuration_extended#9

N/A


djdanielsson:
did you mix up this PR and one of your other ones? cuz this talks about filetree but no changes here are within the filetree role

przemkalit:
I don't think that I made mistake with PR, I thought a lot about this topic and this solution allows to synchronize roles between files exported using filetree_create and the target controller. It is not perfect because if someone delete all permissions from the user/team it won't be synchronized.

djdanielsson:
@Tompage1994 @sean-m-sullivan thoughts on this one? or do we need filetree team to review?

Tompage1994:
Sorry for taking a while on this. I've been back to this 2 or 3 times and I can't really get my head around what you're trying to achieve that the object_diff and filetree stuff doesn't do. 

My gut feel is that this isn't the right place for all this logic as it goes against what we have tried to do for all the other 'endpoint' roles where we are simply looping over the module (as well as adding in extra bits to speed up execution etc.) This does stray away from that and that does make me nervous. 

Let me know if you have compelling reasons why this needs to be part of this role specifically and not part of one of the filetree roles.

przemkalit:
Okay, I know the object_diff but there is an issue when there are lots of the roles (we have around 20k) and while I was using the object_diff it was killing the job due to memory leak. If you can suggest how to solve it it would be great.

Tompage1994:
> Okay, I know the object_diff but there is an issue when there are lots of the roles (we have around 20k) and while I was using the object_diff it was killing the job due to memory leak. If you can suggest how to solve it it would be great.

In honesty I've never tried doing object_diff at scale. If you find this works better then maybe adding this logic into the object_diff for roles as a replacement?

przemkalit:
Ok I will try to add this changes to object_diff.

przemkalit:
Hi, after analysing our approach with the object_diff we've decided not to push it to the collection, so I will close this PR.

PR #887, opened 2024-08-02
support for defining the organization

przemkalit:
# What does this PR do?
Sometimes, it happens that an object does not have an organization set. When creating the export, it is marked as 'organizationless' or just given some text. However, if there is only one organization, we should be able to set a variable for this, so that every object can be matched to the defined organization

# How should this be tested?
```yaml
- name: Get objects
  hosts: all 
  roles:
      vars:
        organization: 'organization_name'
```

N/A

N/A


ivarmu (approved):
LGTM

adonisgarciac:
LGTM

adonisgarciac (approved):

PR #894, opened 2024-08-13
add extra_data to simplified_workflow_nodes in filetree_create role

### LINKED ISSUE #893: Filetree_create does not include extra_data on workflow nodes
filetree_create should export extra_data of workflow nodes

Question posed at https://forum.ansible.com/t/filetree-create-does-not-include-extra-data-on-workflow-nodes/7722

We may need to refactor filetree_create to use awx.awx.export/ansible.controller.export to build exported resources instead of export them using ansible templates. Why haven't we used it so far? 

cc @ivarmu

adonisgarciac:
# What does this PR do?
Export extra_data from workflow_nodes using filetree_create frole

# How should this be tested?
Export, using filetree_create role, some WFJT which has to have a node with extra vars. Then, import that WFJT and WF nodes using dispatch role.

resolves #893


ivarmu (approved):
LGTM

PR #896, opened 2024-08-19
Bug: Skipping project export while exporting related objects

przemkalit:
# What does this PR do?
This PR fixes the issue I discovered today, where the export of the project is always skipped due to the existence of the `project_id` variable within the task in when statement.

# How should this be tested?

```yaml
- name: Get the job template and related project
  hosts: all 
  tasks:
    - name: Retrieve the job_template
      vars:
        job_template_id: 1
        export_related_objects: true
      ansible.builtin.include_role: infra.controller_configuration.filetree_create
```

N/A
#854


ivarmu (approved):
LGTM. Sorry for the late reply, I've been in Holidays :-)

PR #905, opened 2024-09-13
remove failing and flakey tests

jjwatt:
# What does this PR do?
These tests were failing or flakey when run against an awx devel installation. This PR removes them and makes
`ansible-playbook tests/configure_controller.yml`
work correctly and run with passes.

# How should this be tested?
Change tests/configure_controller.yml and tests/config/controller_auth.yml to match your AWX settings and run:
```shell
ansible-playbook tests/configure_controller.yml
```

No.

NA

### DIFF
--- a/tests/configure_controller.yml
+++ b/tests/configure_controller.yml
@@ -187,37 +187,4 @@
         controller_host:         "{{ controller_hostname }}"
         validate_certs:          "{{ controller_validate_certs }}"
       ignore_errors: true  # noqa ignore-errors
-
-    - name: "Get the organization ID"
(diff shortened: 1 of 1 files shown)


Tompage1994 (approved):
At this point, I'm happy enough for this. Perhaps we need a separate thing to test all the diff stuff as that is inherently flakey anyway,

jjwatt:
It seems GHA are failing. These tests were failing locally for me, too, until I put controller auth credentials in both tests/controller_configuration.yml and tests/configs/controller_auth.yml.

PR #909, opened 2024-09-18
Added an example credential with HashiCorp Vault source

ryangniadek:
# What does this PR do?
Currently, the documentation for the `credential_input_sources` role does not provide an example for HashiCorp Vault, which requires different `metadata` fields from the provided CyberArk example. 
- This PR adds an example of using HashiCorp Vault as a credential input source. 
- This PR also adds an example of creating the HashiCorp Vault source credential using the `credentials` role.

# How should this be tested?

This should be tested by comparing the provided example with working config as code implementations that use HashiCorp Vault as a credential input source

No

- We used roles within this collection with HashiCorp Vault credentials and had a hard time finding documentation. Ideally, all supported credential plugins would be documented and this gets us one step closer.
- It is debatable whether adding the source credential is in scope for this PR. Those changes are contained in b69ee64dcaf70ad0c50d31f34beae19f49baee7c if it's decided to remove them from the PR.

### DIFF
--- a/roles/credential_input_sources/README.md
+++ b/roles/credential_input_sources/README.md
@@ -78,12 +78,12 @@ This also speeds up the overall role.
 |:---:|:---:|:---:|:---:|:---:|
 |`target_credential`|""|yes|str|Name of credential to have the input source applied|
 |`input_field_name`|""|yes|str|Name of field which will be written by the input source|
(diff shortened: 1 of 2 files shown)


ivarmu (approved):
LGTM

PR #922, opened 2024-10-03
fix: issue when password is in a survey of WF node and schedule

przemkalit:
# What does this PR do?
Today I've notice encrypted value inside the extra_data, when there is a value inside the workflow node, then when the object is imported the dispatch is failing.

# How should this be tested?
1) Create a JT with survey that has required password variable.
2) Create a WF and add JT to the WF, fill the variable password as it is required value.
3) Export the objects.
4) Import the objects

N/A

N/A


przemkalit:
I did find the same issue in the schedules.

ivarmu (approved):
LGTM

PR #924, opened 2024-10-05
feature - support of exporting access lists

przemkalit:
# What does this PR do?
This PR introduce a feature that I was working recently, which is exporting permission of objects using access list endpoint of API. 

I don't know if you would like to allow this in the collection but I decided to share this feature, but let me know what do you think.

# How should this be tested?
```yaml
- name: Permission export
  hosts: localhost
  connection: local
  gather_facts: false

  tasks:
    - name: Export team permissions of an object
      ansible.builtin.include_role:
        name: filetree_create
        tasks_from: team_access_list
      vars:
        object_id: 1
        object_type: "job_templates"

    - name: Export user permissions of an object
      ansible.builtin.include_role:
        name: filetree_create
        tasks_from: user_access_list
      vars:
        object_id: 1
        object_type: "job_templates"
```

N/A

N/A


djdanielsson:
thank you for contributing so much recently, we are trying to re-organize everything and release an updated combined collection to make the 2.5 experience better, part of this change is moving filetree roles along with a few others into a new repo https://github.com/redhat-cop/aap_configuration_extended. please recreate your PR in it's new repo.

ivarmu (changes requested):
I think it's a very good idea and I appreciate your contribution a lot, I'm sure it'll be very useful. 

That said, one thing I would like to have here, is the possibility of using the object's name instead of the IDs, as the CasC intention is to avoid workiing with IDs and work only with the objects' names.

PR #928, opened 2024-10-13
Renaming to platform variables

sean-m-sullivan:
# What does this PR do?
creates a standard for the collection for authenticating to the platform and our changing our various standard variables over to a single standard.


roles/controller_ad_hoc_command/README.md
```diff
@@ -19,7 +19,7 @@ Currently:
 |`controller_state`|"present"|no|The state all objects will take unless overridden by object default|'absent'|
 |`controller_hostname`|""|yes|URL to the Ansible Controller Server.|127.0.0.1|
 |`controller_validate_certs`|`True`|no|Whether or not to validate the Ansible Controller Server's SSL certificate.||
-|`controller_username`|""|no|Admin User on the Ansible Controller Server. Either username / password or oauthtoken need to be specified.||
```
djdanielsson, line 22: Do we want to change this?

sean-m-sullivan: We want to change all of it, was working my way through instances as you can't replace module options

sean-m-sullivan:
All of our current tests are working plus some newly added ones, However we will need to add gateway and other tests for the newer roles
Gateway vars for tests can be found througout this folder
https://github.com/ansible/aap-gateway/tree/devel/gateway_configuration_collection/tests/integration/targets

PR #930, opened 2024-10-14
replace default GH token with fine grained access token

Tompage1994:
# What does this PR do?

The standard GH token isn't able to create PRs any more so I replaced it in the manual release and it works. This will fix it for the rest of the workflows.

# How should this be tested?

Try it

N/A


djdanielsson (approved):
LGTM

PR #934, opened 2024-10-15
Revert "Revert "Renaming to platform variables""

Tompage1994:
i.e. reopens Sean's #928


galaxy.yml
```diff
@@ -1,7 +1,7 @@
 ---
 namespace: infra
 name: aap_configuration
-version: 3.0.0
+version: 1.0.0
```
Tompage1994: This has always previously been <current_version>-devel (which IIRC is placed in by the CI) this is useful as it makes it easier to install from git and give a sensible version number (I guess it doesn't *really* matter)

roles/controller_ad_hoc_command_cancel/defaults/main.yml
```diff
@@ -1,4 +1,4 @@
 ---
 # These are the default variables specific to the ad_hoc_command_cancel role
-controller_configuration_ad_hoc_command_cancel_secure_logging: "{{ controller_configuration_secure_logging | default('false') }}"
+controller_configuration_ad_hoc_command_cancel_secure_logging: "{{ platform_configuration_secure_logging | default('false') }}"
```
Tompage1994: Should this be changed? Should it not be `platform_configuration_ad_hoc_command_cancel_secure_logging` (and the same for everything)

roles/controller_ad_hoc_command/meta/argument_specs.yml
```diff
@@ -76,45 +76,45 @@ argument_specs:
 
       # No_log variables
       controller_configuration_ad_hoc_command_secure_logging:
-        default: "{{ controller_configuration_secure_logging | default(false) }}"
+        default: "{{ platform_configuration_secure_logging | default(false) }}"
         required: false
         type: bool
         description: Whether or not to include the sensitive ad_hoc_command role tasks in the log. Set this value to `true` if you will be providing your sensitive values from elsewhere.
-      controller_configuration_secure_logging:
+      platform_configuration_secure_logging:
```
Tompage1994: I question whether the `platform_*` variables should actually be `aap_*`? In particular the `platform_configuration_secure_logging` which should surely match to the collection name `aap_configuration_secure_logging`?

That said, should the collection name be infra.platform_configuration? 🤔

Tompage1994 (commented):
Generally happy. @sean-m-sullivan I have a few comments but nothing major

galaxy.yml
sean-m-sullivan: I thought maybe with the new collection, we'd revert to 1.0.0 and start over? If we want to do that, Great, if not, I can revert this.

roles/controller_ad_hoc_command_cancel/defaults/main.yml
sean-m-sullivan: I was back and forth on renaming everything, vs renaming just the global vars, if we want to do another pass and just rename Everything all of it, we can do that. I believe there are some things like controller_settings, which are just a subset of what they were in the past. projects for example exist as separate entities on EDA and controller.

galaxy.yml
Tompage1994: My vote is that this is a continuation of the previous collection so makes sense to be `3.0.0`

@djdanielsson what do you think?

roles/controller_ad_hoc_command/meta/argument_specs.yml
sean-m-sullivan: I think Overall platform is TOO generic, but for variables that would work. Do we want to use aap instead of platform? not against it,

roles/controller_ad_hoc_command_cancel/defaults/main.yml
Tompage1994: Hmm... I wonder also if we should work to make the vars backwards compatible? (defaulting to current naming)

roles/controller_ad_hoc_command/meta/argument_specs.yml
Tompage1994: I think matching the collection name is the right thing to do tbh.

roles/controller_ad_hoc_command_cancel/defaults/main.yml
djdanielsson: I originally thought about that but I think at this point if we are breaking stuff we might as well do anything else we want to simplify or standardize now before it would cause another breaking change

galaxy.yml
djdanielsson: 3.0.0

.github/workflows/pre-commit.yml
```diff
@@ -23,8 +23,8 @@ jobs:
     uses: "redhat-cop/ansible_collections_tooling/.github/workflows/pre_commit_and_sanity.yml@main"
     with:
       collection_namespace: infra
-      collection_name: controller_configuration
-      collection_version: 2.10.0
+      collection_name: aap_configuration
+      collection_version: 3.0.0-devel
       collection_repo: https://github.com/redhat-cop/aap_configuration/
       collection_dependencies: awx.awx
```
djdanielsson: can this take a list?

CHANGELOG.rst
```diff
@@ -298,7 +298,7 @@ Minor Changes
 Breaking Changes / Porting Guide
 --------------------------------
 
-- infra.controller_configuration 2.2.3 is broken, it is aap_utilities release. We are bumping the version to minimize the issues.
+- infra.aap_configuration 2.2.3 is broken, it is aap_utilities release. We are bumping the version to minimize the issues.
```
djdanielsson: does this need to be changed back or do  we need to rewrite this section?

changelogs/changelog.yaml
```diff
@@ -384,7 +384,7 @@ releases:
   2.2.4:
     changes:
       breaking_changes:
-      - infra.controller_configuration 2.2.3 is broken, it is aap_utilities release.
+      - infra.aap_configuration 2.2.3 is broken, it is aap_utilities release.
```
djdanielsson: undo

.github/workflows/pre-commit.yml
Tompage1994: Not without changing it all I believe

Tompage1994:
This is now good to go from my perspective

Tompage1994:
We need to fix all the linting issues etc. before releasing though

.github/workflows/pre-commit.yml
djdanielsson: We will need to make that change to fix the workflows

Tompage1994: Ahh, I see what you're referring to now... (not clear in the GH ui). It does take a space separated list

```suggestion
      collection_dependencies: awx.awx ansible.eda ansible.hub
```

We would also need `ansible.platform` but that doesn't seem to be on galaxy so that may not be too helpful

Tompage1994: @sean-m-sullivan I think you're involved with the ansible.platform collection... is it going to be on galaxy? (or am I missing it?)

djdanielsson: I don't think so, we might need to change the workflow stuff but we can work on that later

sean-m-sullivan: It is not AFAIK, I was involved in the initial push of it, but have not been more recently, but it it is my understanding it will not be public.

djdanielsson (approved):
LGTM

PR #941, opened 2024-10-18
Another Round

sean-m-sullivan:
# What does this PR do?
Starting through the gateway collections to tweak them and the readmes have applications ->role user assignments done, Updating the vars in tasks, defaults, meta, and adding in the loop controls.


roles/gateway_organizations/README.md
```diff
@@ -43,7 +71,7 @@ Options for the `gateway_organizations` variable:
 
 ```json
 {
-  "organizations_list": [
+  "gateway_organizations": [
```
djdanielsson: I do wonder if we should just call this organizations and same with the controller role

roles/gateway_applications/README.md
```diff
@@ -46,7 +90,7 @@ Options for the `applications_list` variable:
 
 ```json
  {
-  "applications_list": [
+  "gateway_applications": [
```
djdanielsson: Similar to organizations I wonder if we should just call this applications

roles/gateway_organizations/README.md
branic: I agree with David. We shouldn’t use the term “gateway” at all.

roles/gateway_applications/README.md
branic: I agree with David. We shouldn’t use the term “gateway” at all.

djdanielsson (approved):
Looking good

PR #944, opened 2024-10-22
Update Readmes and fix a few things

sean-m-sullivan:
# What does this PR do?
Fixing a few items I missed, update readmes start a transition readme.


galaxy.yml
```diff
@@ -21,6 +21,9 @@ build_ignore:
     - tests/*
 license:
     - GPL-3.0-or-later
+dependencies:
```
djdanielsson, line 24: will this work for validated content?

djdanielsson (approved):
LGTM, if the dependencies are an issue we can fix it later

PR #957, opened 2024-10-28
Added job_template_check strategy plugin

Tompage1994:
# What does this PR do?

Adds a strategy plugin for checking that all dependent Job Templates for a playbook are met in Controller (i.e. we use job_launch to call it). It uses the same var as the job_launch role to check for these dependencies (`controller_launch_jobs`).

This comes from a customer use case where we want to create a pattern of launching jobs from inside playbooks, but we want to do a dependency check first to ensure that we don't fail half way through because of missing dependencies. This is similar to what happens for other dependencies (i.e. it checks to make sure roles, modules etc. exist before running)

# How should this be tested?

Add to your ansible.cfg
```
[defaults]
strategy = infra.aap_configuration.job_template_check
```

No

It would be nice to backport this to `infra.controller_configuration` as well

### DIFF
--- a/changelogs/fragments/job_template_check.yml
+++ b/changelogs/fragments/job_template_check.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - Added job_template_check strategy plugin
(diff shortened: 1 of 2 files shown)


djdanielsson:
I am fine with back porting it but I am thinking this should be in extended

sean-m-sullivan:
Agree, I think this collection should be wrapper roles only, its what we think can get into the platform collection once they start accepting roles, or the awx.awx collection, these are the roles we think CAN get supported.

PR #958, opened 2024-10-30
Update Guide

sean-m-sullivan:
# What does this PR do?
Updated the conversion guide, I think its all there, but needs some review in case I've missed some things.


djdanielsson (approved):
LGTM

PR #962, opened 2024-10-31
Revert job template to standard async

sean-m-sullivan:
# What does this PR do?
Found error that caused failures to pass through on job templates, reverting to standard job templates until this is sorted.


djdanielsson (approved):
LGTM

PR #994, opened 2024-12-06
Update office-hours-issue.yml

sean-m-sullivan:
Add office hours recording link, Instead of linking individual, and because we can no longer share google files outside of Red Hat I will be moving the office hour recordings to a personal drive for public consumption and easy sharing.


djdanielsson (changes requested):
linter failed

djdanielsson (approved):
LTGM

PR #1001, opened 2024-12-09
Fix misspelling of hub_ee_repository_sync

MallocArray:
# What does this PR do?

Fix name of role.  All other similar roles all stat with `hub` and this starts with `phub` which appears to be a misspelling

# How should this be tested?

resolves #[number]


djdanielsson (approved):
LGTM

PR #1004, opened 2024-12-11
Update required variables for controller_notification_templates

MallocArray:
Lines 7 and 8 of https://github.com/MallocArray/infra.aap_configuration/blob/devel/roles/controller_notification_templates/tasks/main.yml show this is a mandatory variable

# What does this PR do?

Update organization and notificaiton_type variables to be shown as required.  The current task sets them as mandatory and they are required in the AWX GUI as well

# How should this be tested?

resolves #[number]

### DIFF
--- a/roles/controller_notification_templates/README.md
+++ b/roles/controller_notification_templates/README.md
@@ -80,8 +80,8 @@ This also speeds up the overall role.
 |`new_name`|""|yes|str|Setting this option will change the existing name (looked up via the name field).|
 |`copy_from`|""|no|str|Name or id to copy the Notification template from. This will copy an existing notification and change any parameters supplied.|
 |`description`|""|no|str|The description of the notification.|
-|`organization`|""|no|str|The organization applicable to the notification.|
-|`notification_type`|""|no|str|The type of notification to be sent.|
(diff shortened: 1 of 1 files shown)


djdanielsson (changes requested):
This will need some further review, when looking at the module documentation, it does not show those as required and thus the code itself might need to be updated. Need to verify what is actually correct.

MallocArray:
I had noticed that as well, since
https://docs.ansible.com/ansible/latest/collections/awx/awx/notification_template_module.html
doesn't show it as Required, but in the AWX interface, it clearly is
image

Perhaps it isn't required if `state: absent` is set, so the documentation doesn't show it as required

djdanielsson:
I think that might be the case and thus idk if in code we want to make it required because some people might want to try and remove stuff. I am fine with updating the docs to say required but I think in the code we should remove that mandatory so remove can work.

MallocArray:
I agree, I think it would be better to remove the Mandatory filter on them and leave the documentation as not required, which will result in matching up with the official documentation for the module and make it the most flexible.

djdanielsson:
Feel free to make those changes if you want or I can get to it soon

MallocArray:
Replacing with #1006

PR #1006, opened 2024-12-12
Make organization and notification_type not mandatory to match module documentation in notification_template

MallocArray:
# What does this PR do?

Make organization and notification_type not mandatory to match module documentation in notification_template
These are not mandatory if setting state: absent for removal, so do not make them mandatory in this role
https://docs.ansible.com/ansible/latest/collections/awx/awx/notification_template_module.html

# How should this be tested?

Pipeline testing

No

Replacing #1004 for a cleaner commit history


djdanielsson (approved):
thanks for catching this and fixing it.

PR #1018, opened 2024-12-18
Error handling for controller objects

### LINKED ISSUE #984: Error handling vol. 2.0
Allow managing errors during the import process—why can this be helpful?

Imagine you have plenty of JT, WF, and PRJ files to import, but you’re unsure which ones are valid. Sure, they were exported with success, but what if the import process fails at the first error? What I was proposing is to allow users to generate a list of problematic objects, attempt to fix them, and then relaunch the import process only for those invalid objects.

There was solution introduced in #862, but was removed in #962. 

Today I wanted to add this change to WF and PRJ but I notice that they were removed, and decided to create feature request, and ask you guys, maybe you have a better solution for this.

przemkalit:
# What does this PR do?
This PR is related to #984.

I've prepared the changes for all controller objects. What is different that the default behaviour stay mostly the same, unless variable collect_logs is set to true.

Known issue:
- role name is not available in `['invocation']['module_args']`, I believe it is minor bug in the module itself.

# How should this be tested?
Manually

resolves #984 

N/A


djdanielsson (changes requested):
There are a couple errors in the markdown linter and also sanity checks are failing, can you try and fix those. I think if we were to accept this we would also want it to be more then just controller related objects, but I will wait to hear feedback from others to make sure they are happy with the current implementation before asking you to continue the work to the rest of the roles. @Tompage1994 @sean-m-sullivan

przemkalit:
> There are a couple errors in the markdown linter and also sanity checks are failing, can you try and fix those. I think if we were to accept this we would also want it to be more then just controller related objects, but I will wait to hear feedback from others to make sure they are happy with the current implementation before asking you to continue the work to the rest of the roles. @Tompage1994 @sean-m-sullivan

Sure if I could, but currently we don't have AAP 2.5 so I only worked on objects that I could test, and we still don't use EDA.

djdanielsson:
looks like there are some markdown lint errors

roles/controller_projects/README.md
```diff
@@ -24,6 +24,7 @@ Currently:
 |`controller_oauthtoken`|""|no|str|Controller Admin User's token on the Ansible Controller Server. This should be stored in an Ansible Vault at or elsewhere and called from a parent playbook. Either username / password or oauthtoken need to be specified.||
 |`controller_request_timeout`|`10`|no|int|Specify the timeout in seconds Ansible should use in requests to the Ansible Automation Platform host.||
 |`controller_projects`|`see below`|yes|str|Data structure describing your project or projects Described below. Alias: projects ||
+|`collect_logs`|`false`|no|Specify whether it collects logs for all failed tasks related to project creation, modification, or deletion, instead of failing on the first error.||
```
ivarmu: Type column is missing for this variable entry.

ivarmu (changes requested):

roles/controller_projects/README.md
przemkalit: Fixed.

przemkalit:
I've added support for other objects, but I didn't tested, because I don't have access to AAP 2.5.

djdanielsson:
I am kinda interested in the idea of finishing this up and merging it, @przemkalit have you gotten a 2.5 instance yet to test on? @sean-m-sullivan and @Tompage1994 outside of testing this what other thoughts do you both have on it?

branic:
It would be great for this to be merged. I was just looking at doing something similar for the `infra.controller_configuration` collection. So if this gets merged I use this a base for enabling the same functionality in that collection. (IHAC that is looking for this type of error handling, but is on AAP 2.4).

Tompage1994:
Generally I'm happy enough with this.

At a glance it seems all the new `async_*` files are very similar and could perhaps be condensed into one meta role which performs the actions instead (with the correct vars passed in). That way we reduce duplication and make it easier to adapt in the future

przemkalit:
> I am kinda interested in the idea of finishing this up and merging it, @przemkalit have you gotten a 2.5 instance yet to test on? @sean-m-sullivan and @Tompage1994 outside of testing this what other thoughts do you both have on it?

Hey, sorry I don't have it, and I don't have any way to test it properly, so it would be on you guys.

Tompage1994:
To explain more what I mean by my above comment, I have probably done all of what would be needed. It is untested though, just message app based coding...

So, I think each role now has all the same other than the variables they set essentially. So what I'd do is create a role and it would likely just have the single task file with something like this:

```
---
- name: async | await the results
  block:
    - name: async | Managing Controller Object | Wait for management to complete
      ansible.builtin.async_status:
        jid: "{{ __generic_job_async_results_item.ansible_job_id }}"
      register: __generic_job_async_result
      until: __generic_job_async_result.finished
      retries: "{{ __async_retries }}"
      delay: "{{ __async_delay }}"
  rescue:
    - name: async | Load error details
      ansible.builtin.include_vars:
        file: "{{ __generic_job_async_result.results_file }}"
        name: __error_data

    - name: async | Show error and stop execution
      when: not collect_logs
      ansible.builtin.fail:
        msg: "error: {{ __error_data['msg'] }}, response: {{ __error_data['response'] | default ('N/A') }}"

    - name: async | Building list of not processed objects
      ansible.builtin.set_fact:
        {{ error_list_var_name }}: "{{ lookup('ansible.builtin.vars', error_list_var_name) | default([]) + [
                                                                                {
                                                                                 'name': __error_data['invocation']['module_args']['name'] | default(''),
                                                                                 'error': __error_data['msg'] | default('')
                                                                                }
                                                                               ]
                                  }}"
...
```

Then when you would call that role you would do so with:
```
- name: Managing Controller Credentials | Wait for finish the credential management
  ansible.builtin.include_role: infra.aap_configuration.meta_error_handling
  loop: "{{ __credentials_job_async.results }}"
  loop_control:
    loop_var: __credentials_job_async_results_item
  vars:
    __generic_job_async_results_item: "{{ __credentials_job_async_results_item }}"
    error_list_var_name: __credentials_error_list
```

As I say, I've not tested but I'm pretty certain that would work the same and would significantly reduce the amount of code and files to maintain

djdanielsson (changes requested):
looks great, we just need a changelog fragment and I will merge it

djdanielsson (approved):
LGTM

sean-m-sullivan (approved):
Read over this, looks good to me.

PR #1022, opened 2025-01-10
fix tag for users role

lennysh:
# What does this PR do?
Fixes the tag for the users role

# How should this be tested?
Execute role with the `users` tag instead of the `teams` tag

N/A

N/A


djdanielsson (approved):
good catch, thanks

PR #1041, opened 2025-03-07
Allow to set the delay and the number of retries to attempt when creating workflow nodes

sbilque:
# What does this PR do?

Allow to use the asynchronous retry variables when creating the workflow nodes in the `controller_workflow_job_templates` role.

# How should this be tested?

Manually

N/A

Before this change, the task was only attempting 10 retries, while all other synchronization tasks attempt the specified number with the variable `controller_configuration_workflow_async_delay`. 

Similarly, there was no delay, while the other tasks were waiting for the specified delay with the variable `aap_configuration_async_delay` or `controller_configuration_workflow_async_delay`


Tompage1994 (approved):
LGTM

PR #1043, opened 2025-03-09
Updated role(s) README.md files

matthew-willis-redhat:
# What does this PR do?

Two roles had incorrect code listed in the 'Standard Role Usage' section. Appears to be from copy/paste from other roles/README.md or just incorrect when conditional. Updated to reflect the correct role and when conditions. 

# How should this be tested?

N/A - updated documentation of role README.md

resolves #[number] - No issue created for PR.

N/A


sean-m-sullivan (approved):
LGTM, great catch

PR #1049, opened 2025-03-19
docs(README.md): corrected the column value for new_name parameter

BharadwajAyinapurapu:
# What does this PR do?
### This PR corrects the column value, which was initially set reverse for the column names - **Default Value** & **Required**

# How should this be tested?

### This can be verified by viewing the official documentation, and confirming if its displaying the right value against the right column - [Ansible documentation](https://galaxy.ansible.com/ui/repo/published/infra/controller_configuration/content/role/job_templates/)


djdanielsson (approved):
good catch, thanks

PR #1054, opened 2025-03-28
Restucture the gateway_organizations role 

### LINKED ISSUE #1052: Improve organization management
It is confusing from a user perspective to manage organizations. Organization management is split across two different roles (`gateway_organizations` and `controller_organizations`). The two roles use the same variable (`aap_organizations`), but the documentation of the supported variables is split across the two roles. In addition it is not clear when one or both roles should be used.

Introduce an `organizations` role that combines the documentation and variables into a single role. This role would then use the existing `gateway_organizations` and `controller_organizations` roles as appropriate and in the correct order. The `dispatch` role could still use each of the individual roles so they could be run at the appropriate times.

Another approach would be to combine the two roles into one role and update dispatch to use the new role, but this may introduce additional complexity to the role and may impact the sequencing with the `dispatch` role.

**Additional context**
None

Tompage1994:
# What does this PR do?
Allows so that only one role needs to be called to create and configure the organization.

The solution I went for here means that there are no breaking changes at all. One issue comes around needing to retain the usability by awx.awx users. By bringing the two roles together this would break things if you didn't want to use any of the platform stuff. This solution means all users who use AAP2.5 should transition to use `gateway_organizations` instead (and I've added to the readme to this extent).

# How should this be tested?
I believe the CI would pick up this change. But the best way of testing is just as follows:

```
---
- name: Test org building
  hosts: localhost
  vars:
    aap_organizations:
      - name: orgtest001
        galaxy_credentials:
          - Ansible Galaxy
  roles:
    - name: infra.aap_configuration.gateway_organizations
```

resolves #1052 

I'm unsure whether this is a minor or major change, but went with major as it would involve adjusting how people use the collection.

I've also adjusted the dispatch so that it ONLY uses the gateway role and applies the variables as previously.

### DIFF
--- a/.markdownlint.yml
+++ b/.markdownlint.yml
@@ -18,4 +18,6 @@ MD0046:
 MD033:
   allowed_elements:
     - br
+    - a
+    - img
 ...
--- a/changelogs/fragments/organizations.yml
+++ b/changelogs/fragments/organizations.yml
@@ -1,0 +1,5 @@
+---
+major_changes:
+  - Restucture the gateway_organizations role so that only one role needs to be called to create and configure the organization. Adds the logic which existed from the controller_organizations role previously.
+  - Dispatch no longer calls the controller_organizations role by default, as the gateway_organizations role should be sufficient.
+...
--- a/roles/controller_organizations/README.md
+++ b/roles/controller_organizations/README.md
@@ -2,7 +2,7 @@
 
 ## Description
 
-An Ansible Role to create/update/remove Organizations on Ansible Controller.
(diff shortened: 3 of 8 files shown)


djdanielsson (changes requested):
some lint issues, looks like you might have some dup vars in defaults

djdanielsson:
I didn't love the idea at first but after thinking about it, I think it is a good solution

branic:
I'm in favor of this solution. I haven't had a chance to test it, but with a quick reading through it looks good.

djdanielsson:
I guess I can approve it, I think we need to take another look at the CI, I think we should be able to pull the collections from console if we tried. Also you need to update the base branch and then I will approve and merge

djdanielsson (approved):
LGTM

PR #1063, opened 2025-04-11
explicit fqdn on ansible.controller modules

sean-m-sullivan:
this codifies the changes we have all seen where awx.awx does not work with the gateway changes, it also makes the galaxy importer run clean so we comply with the new requirements for uploading to console.redhat.com repo. 

# How should this be tested?
This updates automated tests.

resolves #[number]


djdanielsson (changes requested):
Shouldn't we change the ansible-lint file to require fqcn now? I think this was the last thing holding that back.

Tompage1994 (approved):
I've just updated the changelog very slightly to be more explicit, but this is good and I agree we should do this.

Tompage1994:
> Shouldn't we change the ansible-lint file to require fqcn now? I think this was the last thing holding that back.

I think you're right on that, I just took the liberty of updating the PR

djdanielsson (approved):
LGTM

PR #1090, opened 2025-05-16
Add enabled and disabled actions in global_vars

adriengentil:
This PR adds `enabled` and `disabled` actions in global vars to align
with the states supported by `eda.rulebook_activation`:
https://github.com/ansible/event-driven-ansible/blob/main/plugins/modules/rulebook_activation.py#L151-L161


djdanielsson (approved):
LGTM

PR #1092, opened 2025-05-16
Add dependencies

sean-m-sullivan:
# What does this PR do?
Add dependencies

Add back gateway organizations readme.


djdanielsson (approved):
LGTM

Tompage1994 (changes requested):
Arguably this is a major change. This will cause failures for installation if the user doesn't have access to the underlying modules. (at the moment they could just about get by with awx.awx and some of the other upstream collections, albeit with several issues).

This is taking things to the next step so should be a major change IMO, so in need of a changelog

ivarmu:
> Arguably this is a major change. This will cause failures for installation if the user doesn't have access to the underlying modules. (at the moment they could just about get by with awx.awx and some of the other upstream collections, albeit with several issues).
> 
> This is taking things to the next step so should be a major change IMO, so in need of a changelog

I agree with @Tompage1994 but I also think that this collection is against AAP 2.5, with all the implications it have... and one of them is the non-supporting the `awx.awx` collection anymore, am I wrong?

djdanielsson:
No, awx.awx doesn't work anymore we changed it to ansible.controller. So this shouldn't be a breaking change

sean-m-sullivan:
This was a breaking change for [v3.3.0](https://github.com/redhat-cop/infra.aap_configuration/blob/devel/CHANGELOG.rst#id79)
and included in the changelog, this just codifies it in the dependencies now that we know the console.redhat.com isn't borked when doing that. 

Also all. the roles have the [FQDN explicitly set](https://github.com/redhat-cop/infra.aap_configuration/blob/aca2512d175675b2218dce3066758b3a23736edf/roles/controller_credential_types/tasks/main.yml#L3)

So not a breaking change, just actually putting it in the dependencies. It MIGHT break some people ansible-builder for EE's that were doing it without ansible.controller, but those EE's would have never worked to begin with and were broken from the get go.

sean-m-sullivan:
Added back gateway dependencies readme, now dependent on https://github.com/redhat-cop/ansible_collections_tooling/pull/64

Tompage1994 (approved):
I'm happy with this now. I'm happy that because we already have the FQDN set it is only a minor change (i.e. some people may have to slightly alter how they bring in this collection (for example in how we do CI) but otherwise, I see it should be ok

PR #1093, opened 2025-05-19
renamed ah_* variables to hub_*

### LINKED ISSUE #1056: Refactor ah_* variables to hub_*
In relation to the hub_* roles, the vars are all named ah_*. These should be replaced with hub_*

They should also be defaulted so that they are backwards compatible. e.g.

```
hub_namespaces: "{{ ah_namespaces }}"
```

cc @redhat-cop/automation-cop-tower-mgrs

Tompage1994:
# What does this PR do?
- Renamed ['ah_state', 'ah_path_prefix', 'ah_token'] variables to ['hub_state', 'hub_path_prefix', 'hub_token'] in hub_ roles.
- Defaulted the new variables ('hub_state', 'hub_path_prefix', 'hub_token') to the old ones for backward compatibility.

# How should this be tested?
Use the new variables. Also use the old ones and check they still function

resolves #1056 

None


roles/hub_collection/meta/argument_specs.yml
```diff
@@ -50,7 +50,7 @@ argument_specs:
         required: false
         description: URL to the Ansible Automation Platform Server.
         type: str
-      ah_path_prefix:
+      hub_path_prefix:
```
ivarmu: Maybe this change could be done in a future PR? If the default value for `hub_path_prefix` is `ah_path_prefix`, both variables should be allowed by now... In a future release may be deprecated at all...

djdanielsson (approved):
LGTM

roles/ansible_config/templates/ansible.cfg.j2
```diff
@@ -13,7 +13,7 @@ auth_url={{ server.auth_url }}
 {% endif %}
 {% if server.token is defined %}
 token={{ server.token }}
-{% elif ah_token is defined %}
-token={{ ah_token }}
+{% elif aap_token is defined %}
+token={{ aap_token }}
```
djdanielsson: Thinking about it, I think this in theory is a breaking change, though I think we should just keep it breaking idk

Tompage1994: You're right. This was the first I did. It should probably be as below?

```suggestion
token={{ aap_token | default(ah_token) }}
```

djdanielsson:
Now just need to fix the CI

djdanielsson:
looks like it still doesn't like your changelog message

PR #1094, opened 2025-05-19
The 'auto_migrate_users_to' field can't be 'false' as default. Omit instead.

ivarmu:
# What does this PR do?

The field `auto_migrate_users_to` expects a name of an `authenticator`, is not a boolean. So, the current default value `false` causes the module to look for the `false` authenticator, which doesn't exist.

# How should this be tested?

Manually

N/A

N/A


djdanielsson (approved):
LGTM

djdanielsson:
if you rebase it then I can merge it

PR #1101, opened 2025-05-22
Remove temporary redirection

Tompage1994:
Remove the temporarily created file.

Do not merge before 12:00 EST


djdanielsson (approved):
LGTM

Issue #1100, opened 2025-05-22
Implement automated releasing to improve our release cycle.


Tompage1994:
**Is your feature request related to a problem? Please describe.**
Implement automated releasing to improve our release cycle. 

Initial plan would be to release every two weeks.

A/C:
- Release on a schedule (2 weeks)
- Don't release if there is a major or breaking change
- Manual release can still be issued
- Automated versioning will need to be implemented to calculate the new version number of the release depending on the changelog fragments which have been created (check to see if the antsibull-changelog framework has that ability)

Tompage1994:
@djdanielsson I realise I missed one of the things we discussed in my PR.

I didn't implement not releasing if there is a major change. Do we want to do this, or should we just leave it to do so for now? We don't often have major changes so I guess it isn't too much of an issue

djdanielsson:
might want to add that logic before we forget and then do a major change months down the road and it does a release

PR #1115, opened 2025-05-30
Prevent an empty automatic release

### LINKED ISSUE #1114: Scheduled release will release even if no changes happened
## Summary
At present a scheduled release will release even if no changes happened. We need to make sure it does not because we don't want a bumped version with no changes.

If the release is launched manually it should continue to bump the version and do a release, just not when scheduled.

## Issue Type

- Bug Report

Tompage1994:
# What does this PR do?
I realised that at present if no changes had been made we would still create a new patch release. This shouldn't be the case. I've made it so that it is still possible to perform a release that has no changelogs (we may need to do this in some circumstances), this is possible when we perform a manual release.

# How should this be tested?

See it stopping the release here when no changes: https://github.com/Tompage1994/controller_configuration/actions/runs/15344682751 

and continuing here when there are: https://github.com/Tompage1994/controller_configuration/actions/runs/15344892665 (I cancelled it doing the actual deployment to galaxy/AH, but it still proves it out)

resolves #1114 

Related to #1100

### DIFF
--- a/.github/files/calculate_change.sh
+++ b/.github/files/calculate_change.sh
@@ -1,3 +1,9 @@
+if [ -z "$(ls changelogs/fragments/*.yml changelogs/fragments/*.yaml)" ]; then
+  echo "change_present=false" >> "$GITHUB_OUTPUT"
+else
+  echo "change_present=true" >> "$GITHUB_OUTPUT"
+fi
+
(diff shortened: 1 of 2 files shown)


djdanielsson (approved):
LGTM

PR #1121, opened 2025-06-05
Remove unnecessary line

Tompage1994:
This was a hangover from when the jobs were all in this file. The logic is now in the tooling repo


djdanielsson (approved):
LGTM

PR #1123, opened 2025-06-05
Reverts the aap_token being applied to hub roles. 

### LINKED ISSUE #1122: aap_token does not work against hub modules
## Summary

The aap_token does not work against hub modules 

## Issue Type

- Bug Report

```console (paste below)
ansible --version 
latest

all latest

Controller version 
AAP 2.5
```

We recently changed to using `aap_token` for all the hub roles, but this doesn't work. We should change this to `hub_token` as a bugfix.

Tompage1994:
# What does this PR do?
Reverts the aap_token being applied to hub roles. 
Instead hub_token is now used.
# How should this be tested?
The API requires a token created by hub. This works with that. The best way to get the token is to use the ah_token module.

resolves #1122 

N/A

### DIFF
--- a/changelogs/fragments/hub_token.yml
+++ b/changelogs/fragments/hub_token.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - Reverts the aap_token being applied to hub roles. Instead hub_token is now used.
+...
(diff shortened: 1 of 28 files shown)


djdanielsson (approved):
LGTM

PR #1130, opened 2025-06-24
Remove unnecessary lint config

Tompage1994:
# What does this PR do?
Remove unnecessary lint config. Also fixes one subsequent lint failure

# How should this be tested?
CI will work

N/A

Removed similar in aap_configuration_extended. The key change that allows us to remove the mock_modules is that we now have the actual modules loaded when running lint in the CI


djdanielsson (approved):
LGTM

PR #1135, opened 2025-07-10
add object_ids parameter to role_user_assignment

adonisgarciac:
# What does this PR do?
adding object_ids to gateway_role_user_assignments

### DIFF
--- a/changelogs/fragments/objects_ids.yml
+++ b/changelogs/fragments/objects_ids.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - Add object_ids parameter to gateway_role_user_assignments role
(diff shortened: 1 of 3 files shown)


ivarmu (changes requested):
As `object_id` is deprecated, shouldn't it be also be reflected here?
```
   object_id  Deprecated
              This option is deprecated and will be removed in a release after 2026-01-31.
              For associating a user to team(s)/organization(s), please use the object_ids param.
               -------------
              Primary key/Name of the object this assignment applies to.
              This option is mutually exclusive with object_ids and object_ansible_id.
        default: null
        type: int

   object_ids  List of object IDs(Primary Key ) or names this assignment applies to.
               This option is mutually exclusive with object_id and object_ansible_id.
        default: null
        elements: str
        type: list
```

Also, the documentation may be updated and a changelog fragment is needed.

adonisgarciac:
I've added it to README and changelog. I wouldn't remove object_id even though it is deprecated until deprecation is effective. I added deprecation warning to README

ivarmu (approved):
LGTM

PR #1140, opened 2025-07-31
add object_ids to arguments specs

tosmi:
# What does this PR do?

- ansible.platform as of 20250702 supports the objects_ids parameter which contains a list of ids or names assignments

- set object_id to optional as this is now deprecated and is mutual exclusive with object_ids

- #1135 added the parameter object_ids but the argument spec was missing

# How should this be tested?

use object_ids in gateway_role_user_assignments

no

PR #1135


djdanielsson (changes requested):
can you make a changelog fragment to note this change? also in that change please be specific which version of the ansible.platform collection is required

djdanielsson (approved):
LGTM

PR #1149, opened 2025-08-26
quick attempt to create some roles for testing

djdanielsson:


roles/gateway_role_user_assignments/README.md
```diff
@@ -1,112 +1,217 @@
-# Ansible Role infra.aap_configuration.gateway_role_user_assignments
```
sean-m-sullivan, line 1: I hate the readme changes, can we revert them? otherwise this looks fine

sean-m-sullivan (approved):

roles/gateway_role_user_assignments/README.md
djdanielsson: I generated them just to have something that was there for testing because I didn't have the time to use our normal template

PR #1156, opened 2025-09-03
Fix accidentally removed usage of variables

### LINKED ISSUE #1154: *_async_retries and *_async_delay variables no longer work
## Summary
Since #1018 we have accidentally removed the `*_async_retries` variables (such as `hub_configuration_collection_repository_sync_async_retries`).

I don't think we meant to do this, and we should build the logic back in to allow this to work again. The documentation still also refers to these vars.

We also have the same issue with `*_async_delay`

We should implement the same solution as we have done for `cas_secure_logging` being set to the role `*_secure_logging` var.

## Issue Type

- Bug Report

Collection version `3.5.0`

## Desired Behavior

Applying `hub_configuration_collection_repository_sync_async_retries` makes a change

## Actual Behavior

`hub_configuration_collection_repository_sync_async_retries` is ignored and only the global value is used

Tompage1994:
Fix accidentally removed usage of '*_async_delay' and '*_async_retries' variables

# What does this PR do?
Adds in the logic to make use of the variables for each role which set the async delay and retries.

# How should this be tested?
variables such as `hub_configuration_collection_repository_sync_async_delay` will work again

resolves #1154 

None

### DIFF
--- a/changelogs/fragments/1154_retries_delay.yml
+++ b/changelogs/fragments/1154_retries_delay.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - Fix accidentally removed usage of '*_async_delay' and '*_async_retries' variables
+...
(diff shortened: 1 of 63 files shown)


branic (approved):
LGTM

PR #1160, opened 2025-09-04
Remove pull_request as workflow trigger which always fails

Tompage1994:
# What does this PR do?

Remove pull_request as workflow trigger which always fails

# How should this be tested?

Merge and test.

No

https://github.com/redhat-cop/ansible_collections_tooling/pull/85


djdanielsson (approved):
LGTM

Issue #1165, opened 2025-09-08
infra.aap_configuration:3.5.1 fails while waiting for objects creation


carlosfparada:
## Summary

Just upgraded to infra.aap_configuration collection version 3.5.1 and it started to fail while waiting for async answers (any role). This did not happen with previous versions. Downgrading to 3.5.0 and it works fine.

```console (paste below)

TASK [infra.aap_configuration.collect_async_status : Managing AAP Object | Wait for management to complete] **********************
fatal: [localhost]: FAILED! => {"censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result"}

TASK [infra.aap_configuration.collect_async_status : Handle Async Job Failure] ***************************************************
fatal: [localhost]: FAILED! => {"censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result"}

```

Adding the following vars, the 3.5.1 version also works properly (tested by @djdanielsson suggestion):
    controller_configuration_credentials: false
    controller_configuration_execution_environments: false
    controller_configuration_projects: false
    controller_configuration_project_update: false
    controller_configuration_inventories: false
    controller_configuration_hosts: false
    controller_configuration_inventory_sources: false
    controller_configuration_inventory_source_update: false
    controller_configuration_job_templates: false

## Issue Type

- Bug Report

## Ansible, Collection, Controller details

```console (paste below)
# ansible-galaxy collection list | grep infra   
infra.aap_configuration                  3.5.1       
infra.ah_configuration                   2.1.0       
infra.controller_configuration           3.1.3       
infra.eda_configuration                  1.1.0

cparada@cparada-mac AAP % ansible-galaxy collection list | grep control
ansible.controller                       4.6.19

ansible --version
ansible [core 2.18.8]
```

- ansible installation method: one of source, pip, OS package, EE

## STEPS TO REPRODUCE

Use infra.aap_configuration 3.5.1 to add objects to AAP, e.g. 

    - {role: infra.aap_configuration.controller_credentials, when: controller_credentials is defined}
    - {role: infra.aap_configuration.controller_execution_environments, when: controller_execution_environments is defined}
    - {role: infra.aap_configuration.controller_projects, when: controller_projects is defined}
    - {role: infra.aap_configuration.controller_project_update, when: controller_projects is defined}
    - {role: infra.aap_configuration.controller_inventories, when: controller_inventories is defined}
[...]

PR #1166, opened 2025-09-08
fixing no_log issue

### LINKED ISSUE #1164: Cannot set cas_secure_logging
## Summary

Cannot overwrite cas_secure_logging, because it is owerwritten by calling role.

## Issue Type

- Bug Report

## Desired Behavior

I want to overwrite cas_secure_logging for troubleshooting.
I was expecting to be able to be able to do this, but I can't.
I was also expecting that this is can be done by using aap_configuration_secure_logging.

## Actual Behavior

As seen for example in https://github.com/redhat-cop/infra.aap_configuration/blob/devel/roles/controller_instances/tasks/main.yml#L53C33-L53C68
it is not set to controller_configuration_instances_secure_logging, as might be expected, but to controller_configuration_instances which somehwat confuses me.

Yours
  David

djdanielsson:
# What does this PR do?
should fix some issues that appear to have accidentally added in #1156

# How should this be tested?
manually (currently untested)

resolves #1164

#1156 and #1165

### DIFF
--- a/changelogs/fragments/no_log_fix.yml
+++ b/changelogs/fragments/no_log_fix.yml
@@ -1,0 +1,4 @@
+---
+bugfix:
+  - fixing bug that got introduced in 3.5.1 causing no_log issues
+...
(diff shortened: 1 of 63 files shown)


djdanielsson:
@Tompage1994 @branic this is related to some recent PR's you all did. does this look correct?

branic (changes requested):
This looks good to me. The only thing I see missing is a changelog fragment.

branic (approved):
LGTM

Tompage1994 (approved):
Oh that looks more like what I should have done. 😬

PR #1174, opened 2025-09-16
Fix async logging default value

branic:
# What does this PR do?

Removes setting the default for the `aap_configuration_collect_logs` in every role and sets the default in the only location the variable is used.

# How should this be tested?

Manually

No

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ ansible.cfg
 tests/localtest.yml
 .ansible/
 .venv
+.devcontainer
+ansible_collections/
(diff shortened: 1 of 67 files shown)


branic:
There are a couple of other minor inconsistencies that I also fixed.

djdanielsson:
Why move away from the defaults file? Just not sure I understand the value of this change

branic:
To simplify. The same value was being set as the default in every role which is lots of redundant code. This particular variable doesn't have role specific variants so there isn't a need to set it in each role. By specifying the default in the only place it is used ensures that there is always a default (even when new roles are added) and eliminates all the redundant code.

Had I realized I was creating so much duplication when I was working on the PR that introduced the async logging I would have done this then, but there were so many variable changes I didn't see the forest through the trees until I started implementing the async logging change in the infra.controller_configuration collection.

djdanielsson (approved):
LGTM

PR #1179, opened 2025-09-17
More async related fixes

branic:
# What does this PR do?

This PR adjusts the roles readme's to be consistent and fixes the wording related to the `aap_configuration_collect_logs` variable.

In addition I found that applying the aysnc error collecting was missed for the `controller_settings` role so this has also been addressed.

The last is a minor fix to the checking if the `aap_configuration_collect_logs` variable is set. As it is used in a `when` clause it should always be cast to a boolean.

# How should this be tested?

Manually

No


changelogs/fragments/async_fixes
```diff
@@ -0,0 +1,5 @@
+---
+minor_changes:
```
djdanielsson: I would call it a bug fix

djdanielsson (changes requested):

changelogs/fragments/async_fixes
branic: Makes sense to me. I've made that change.

Tompage1994 (approved):

djdanielsson (approved):

PR #1181, opened 2025-09-22
updating hub_publish role

djdanielsson:
# What does this PR do?

trying to update all vars to use hub_ vs ah_ but keep backwards compatibility as much as possible. I do wonder if defining hub_collection_list will be an issue.

# How should this be tested?

for now it needs to be manually checked

none

none


sean-m-sullivan (approved):
This should look good, are you able to test this with hub collection list not defined?

djdanielsson:
having a hard time testing this one, either it skips everything or errors it feels like

sean-m-sullivan (approved):
LTGM, not finding flaws

PR #1183, opened 2025-09-25
fix each role `*_secure_logging` default value on doc for consistency with `defaults/main.yml`

akira6592:
# What does this PR do?

fix each role `*_secure_logging` default value on doc for consistency with `defaults/main.yml`.

# How should this be tested?

Document only.

N/A

N/A


roles/eda_credential_types/README.md
```diff
@@ -27,7 +27,7 @@ eda_configuration_credential_types_secure_logging defaults to the value of aap_c
 
 |Variable Name|Default Value|Required|Description|
 |:---:|:---:|:---:|:---:|
-|`eda_configuration_credential_types_secure_logging`|`false`|no|Whether or not to include the sensitive Registry role tasks in the log.  Set this value to `true` if you will be providing your sensitive values from elsewhere.|
+|`eda_configuration_credential_types_secure_logging`|`true`|no|Whether or not to include the sensitive Registry role tasks in the log.  Set this value to `true` if you will be providing your sensitive values from elsewhere.|
```
djdanielsson, line 30: cred types shouldn't need secure logging by default, it's the creds themselves that do

roles/controller_users/README.md
```diff
@@ -44,11 +44,11 @@ Enabling this will enforce configuration without specifying every option in the
...
 The role defaults to false as normally the add user task does not include sensitive information.
-`controller_configuration_user_secure_logging` defaults to the value of `aap_configuration_secure_logging` if it is not explicitly called. This allows for secure logging to be toggled for the entire suite of configuration roles with a single variable, or for the user to selectively use it.
+`controller_configuration_users_secure_logging` defaults to the value of `aap_configuration_secure_logging` if it is not explicitly called. This allows for secure logging to be toggled for the entire suite of configuration roles with a single variable, or for the user to selectively use it.
 
 |Variable Name|Default Value|Required|Description|
 |:---:|:---:|:---:|:---:|
-|`controller_configuration_user_secure_logging`|`false`|no|Whether or not to include the sensitive user role tasks in the log. Set this value to `true` if you will be providing your sensitive values from elsewhere.|
+|`controller_configuration_users_secure_logging`|`true`|no|Whether or not to include the sensitive user role tasks in the log. Set this value to `true` if you will be providing your sensitive values from elsewhere.|
```
djdanielsson, line 51: I am kinda concerned about changing this name, while I agree that it matches the naming convention vs what we have currently, if users have been using the old name us changing it now will break that

djdanielsson: @sean-m-sullivan @Tompage1994 thoughts?

djdanielsson (changes requested):

roles/controller_users/README.md
sean-m-sullivan: agreed, maybe we default to the old?

roles/eda_credential_types/README.md
akira6592: @djdanielsson Thank you for the comment.

I fixed only document according to [`defaults/main.yml`](https://github.com/akira6592/infra.aap_configuration/blob/79b0213aec744397fe47b62c8238513d8179b9f4/roles/eda_credential_types/defaults/main.yml#L3) (`true`).

Should I fix default to `false` in `defaults/main.yml` ?

roles/controller_users/README.md
akira6592: @djdanielsson @sean-m-sullivan

Thank you for the comment.

The variable names in `defaults/main.yml` and `tasks/main.yml` are `controller_configuration_users_secure_logging`.

I fixed only document according to implementation.

Should I fix implementation side?

djdanielsson: ah, ok then it is good

djdanielsson (approved):
LGTM

akira6592:
Thank you for merging!

PR #1185, opened 2025-09-30
include vars for dispatch option

djdanielsson:
# What does this PR do?

looking to support an interesting way to include vars from different sources and organize it all with dispatch

# How should this be tested?

manually

no

none


Tompage1994:
I'm not sure I entirely understand what's going on here and what the use case is

sean-m-sullivan (approved):

sean-m-sullivan:
it allows you to include vars for say aap_gateway_all or aap_gateway_dev, or whatever they are named, as long as the prefix matches, useful for those not using inventories as much or with multiple environments.

roles/dispatch/defaults/main.yml
```diff
@@ -1,4 +1,5 @@
 ---
+dispatch_sort_included_vars: false
```
Tompage1994: I don't like the name of the variable as I don't think it really covers what it does.

We may want to collect ideas for this but I was thinking something more along the lines of `dispatch_include_wildcard_vars` or similar.

In particular we aren't really 'sorting' the vars in any way and really they don't need to be explicitly 'included' as they could be applied using inventories or any other mechanism really.

roles/dispatch/README.md
```diff
@@ -8,6 +8,27 @@ An Ansible Role to run all roles in the infra.aap_configuration collection.
 
 This is a meta role, its purpose is to run the other roles in the collection, it does not run all of them, and can be used to call roles in a custom order, The control variable is the `aap_configuration_dispatcher_roles` which will pull roles from the different services and run them, If you wish to just run a subset of services from our default lists remove entries from this var.
 
+There is also an option `dispatch_sort_included_vars` which is default set to `false` but if true it expects you to have used include_vars dir option and the vars themselves to match the naming but add on to them for example `controller_templates_dev`.
```
Tompage1994: Can we improve this readme a little more to make it clearer exactly what it does please? Perhaps the example could be more of a walkthrough of what actually happens. (i.e. you include/apply some vars -> they get concatenated to the main var -> they're applied to the role)

roles/dispatch/tasks/main.yml
```diff
@@ -1,4 +1,257 @@
 ---
+- name: Sort out vars that are included
+  when: dispatch_sort_included_vars
+  block:
```
Tompage1994: Can these all be separated out into a separate task file which is only included when the variable is set to true?

Tompage1994: e.g. `tasks/include_wildcard_vars.yml` (or matching whatever the var name ends up being)

Tompage1994 (changes requested):
A deeper look into it and Sean's explanation has helped. I get what's going on here and I think this is a really useful addition.

BUT I think we need better documentation and examples in that readme AND I think the name of the variable isn't great as it doesn't really capture what it's doing

djdanielsson:
having weird issues with actions, it seems to be a commit behind. anyways finally got all the errors cleaned up

roles/dispatch/README.md
djdanielsson: I tried explaining it a lot more, let me know what you think now

Tompage1994 (approved):
LGTM

sean-m-sullivan (approved):

PR #1186, opened 2025-10-01
Async status updates

djdanielsson:
# What does this PR do?

Add cas_object_label to all roles using collect_async_status

Add cas_object_label variable to all Ansible roles that include the
infra.aap_configuration.collect_async_status role. This provides
better logging and debugging information by labeling async operations
with descriptive names based on the specific resource being processed.

For roles with existing loop_control labels, the cas_object_label copies
the exact label value. For roles without labels, descriptive labels
are created based on the operation type and resource name.

This was done with assistance from grok-code-fast-1
# How should this be tested?

manually

no

none

### DIFF
--- a/changelogs/fragments/async_improved_logging.yml
+++ b/changelogs/fragments/async_improved_logging.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - Added logic to better label the collect_async_status role/tasks to more easily understand where automation is at any given point
+...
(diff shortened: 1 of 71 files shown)


djdanielsson:
I have done some basic testing and it appears to be working well

sean-m-sullivan (approved):

djdanielsson:
I found a few vars that aren't correct while it doesn't error it doesn't display how I think it should so double checking as many as I can easily test with some configs I have already written and then it should be ready to merge

djdanielsson:
I am sure there are a few that need fixing but those can be fixed in the future I would like to get this in sooner along with a couple other things before our next release

PR #1192, opened 2025-10-03
updates for 2.6

djdanielsson:
# What does this PR do?

adding gateway_role_team_assignments and gateway_role_definitions and updating gateway_role_user_assignments

# How should this be tested?

manually

no

none

### DIFF
--- a/changelogs/fragments/aap_26_additions.yml
+++ b/changelogs/fragments/aap_26_additions.yml
@@ -1,6 +1,6 @@
 ---
 minor_changes:
-  - Updated `gateway_role_user_assignments` role
-  - Added `gateway_role_team_assignments` role
-  - Added `gateway_role_definitions` role
+  - Updated `gateway_role_user_assignments` role these changes will only work in AAP 2.6+
+  - Added `gateway_role_team_assignments` role for AAP 2.6+
+  - Added `gateway_role_definitions` role for AAP 2.6+
 ...
--- a/changelogs/fragments/aap_26_additions.yml
+++ b/changelogs/fragments/aap_26_additions.yml
@@ -1,0 +1,6 @@
+---
+minor_changes:
+  - Updated `gateway_role_user_assignments` role
+  - Added `gateway_role_team_assignments` role
+  - Added `gateway_role_definitions` role
+...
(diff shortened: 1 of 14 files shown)


Tompage1994 (changes requested):
Looks good. Needs a changelog though

djdanielsson:
yea I am working on that and trying to make sure I note that the 2 roles will only work in 2.6

roles/gateway_role_definitions/meta/argument_specs.yml
```diff
@@ -0,0 +1,100 @@
+---
+argument_specs:
+  main:
+    short_description: An Ansible Role to create role_definitions on Ansible gateway for AAP 2.6+.
+    options:
+      aap_role_definitions:
```
djdanielsson: do these arg_specs look good?

Tompage1994: Doesn't seem to match the variable you're actually using (`gateway_role_definitions`)

roles/gateway_role_user_assignments/README.md
```diff
@@ -39,74 +54,65 @@ This also speeds up the overall role.
...
 
 ## Data Structure
 
-### Role User Assignments Arguments
+### Role User Assignment Variables
+
+**WARNING: Some options only will work in AAP 2.6+**
 
-Options for the `gateway_role_user_assignments` variable:
+Options for the `aap_role_user_assignments` variable:
```
Tompage1994: Have you actually updated the variable?

changelogs/fragments/aap_26_additions.yml
```diff
@@ -0,0 +1,6 @@
+---
+minor_changes:
+  - Updated `gateway_role_user_assignments` role these changes will only work in AAP 2.6+
```
Tompage1994, line 3: Have you actually changed the tasks? It doesn't seem to be the case, in which case it should still work in 2.5.

Ideally any change to this role will be backwards compatible to 2.5

Tompage1994 (changes requested):

changelogs/fragments/aap_26_additions.yml
djdanielsson: so some options work with 2.5 but not all of them I believe

djdanielsson:
@Tompage1994 does it look good now?

sean-m-sullivan (approved):
LGTM, good changes

Tompage1994 (approved):

PR #1196, opened 2025-10-07
adding new roles to dispatcher

djdanielsson:
# What does this PR do?

adds the new roles to dispatch

# How should this be tested?

manually

no

none

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ collections:
 
 |                                      Collection Name                                |            Purpose            |
 |:-----------------------------------------------------------------------------------:|:-----------------------------:|
-| ansible.platform repo (no public repo for this collection)                          | gateway/platform modules      |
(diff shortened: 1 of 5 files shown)


Tompage1994 (changes requested):
Just wondering whether team assignments should come after user assignments?

Also, I think this value is documented in the readme as well, so probably needs updating

djdanielsson:
I wasn't positive on what order to put them in, I changed it a couple times and just decided to commit something and let others give feedback. The primary thing I wanted to do is get this out there now.

akira6592:
>Also, I think this value is documented in the readme as well, so probably needs updating

I think the same is true for `meta/argument_specs.yml`.

sean-m-sullivan (approved):

sean-m-sullivan:
I'd put them as Team assignments first, than users, Teams should be prioritized over user assignments,

Tompage1994 (approved):

PR #1197, opened 2025-10-07
Fixing more inconsistencies

djdanielsson:
# What does this PR do?

AI found some more inconsistencies

# How should this be tested?

manually, which I have not so needs thorough testing to ensure arg specs aren't messed up. I think it just uncommented them all but they were created a while ago so needs testing.

no

none


akira6592:
@djdanielsson Thank you for fixing!

If we change to the default values of `*_async_retries` `30` to `50`, I think we should apply to the `README.md`.

sean-m-sullivan (approved):
LGTM If it passes an integration test would be the only thing I'd check on as the argument spec changes.

Tompage1994:
We decided NOT to include the sub-options of each variable in the arg spec as it could potentially break some people's automation if they choose to hold additional fields under a variable (which there are some use cases for). In this case we would break the role for them. I propose keeping them commented out

djdanielsson:
I am conflicted on this... I would really like to have it there because it can really help catch issues but I understand the concern... I wonder how many people are passing variables that are not direct arguments.

djdanielsson:
commented it back out for now so we could make the other changes

PR #1205, opened 2025-10-08
Fix gather_facts to false in configure_aap.yml

akira6592:
# What does this PR do?

Unnecessary `gather_facts` set to false in `configure_aap.yml` for better performance.

# How should this be tested?

manual locally tested

N/A

N/A

### DIFF
--- a/changelogs/fragments/gather_face_false.yml
+++ b/changelogs/fragments/gather_face_false.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - Unnecessary gather_facts set to false in configure_aap.yml
+...
--- a/changelogs/fragments/gather_face_false.yml
+++ b/changelogs/fragments/gather_face_false.yml
@@ -1,4 +1,4 @@
 ---
-minor_changes:
+bugfixes:
   - Unnecessary gather_facts set to false in configure_aap.yml
 ...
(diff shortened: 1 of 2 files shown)


sean-m-sullivan:
Gather facts by default runs and is true, The play does not need any host facts, is there a reason we need to start gathering facts and add additional time to a run? We turned off gather facts to make it run faster.

changelogs/fragments/gather_face_false.yml
```diff
@@ -0,0 +1,4 @@
+---
+minor_changes:
```
djdanielsson: lets call this a `bugfixes` because of the way versioning works I don't think it should be considered a minor_change.

djdanielsson (changes requested):

akira6592:
I think there is no reason to start gathering facts. So I set to `gather_facts: false` by this PR.

I apologize for the misleading title. I have corrected it.

changelogs/fragments/gather_face_false.yml
akira6592: @djdanielsson Thank you. I fixed it.

djdanielsson (approved):
LGTM

Tompage1994 (approved):

akira6592:
Thank you for merging!

sean-m-sullivan:
It was too early and misunderstood, I thought we were changing behavior, this is how it should run, I approve the facts to false.

PR #1212, opened 2025-10-14
Fix hub_publish role

### LINKED ISSUE #1211: hub_publish role broken
The hub_publish role doesn't do anything since 3.7.0 when using hub_collections.

This is due to several conditionals in tasks being incorrect. PR incoming.

myllynen:
Resolves: #1211

### DIFF
--- a/roles/hub_publish/tasks/main.yml
+++ b/roles/hub_publish/tasks/main.yml
@@ -7,7 +7,7 @@
     - name: Git checkout
       ansible.builtin.git:
         repo: "{{ __hub_collection_item.git_url }}"
-        dest: "{{ aap_configuration_working_dir }}/{{ __hub_collection_item.collection_name }}/"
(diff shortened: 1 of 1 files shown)


roles/hub_publish/tasks/main.yml
```diff
@@ -135,17 +132,28 @@
...
       when:
         - not hub_auto_approve
+        - __hub_collection_list is defined
       register: approval
       retries: 3
       delay: 2
       until: not approval is failed
 
   always:
+    - name: Cleanup working dir
```
djdanielsson, line 146: I don't think you want this to always run

myllynen: Fair enough, I moved it so that only if everything succeeds then the dirs created will be removed.

djdanielsson (approved):
LGTM

PR #1226, opened 2025-11-07
Fix issue #1225

### LINKED ISSUE #1225: Controller instance groups assigned to org before rollout
## Summary

Instance Groups needs to be created before assignment to the organizations.

## Issue Type

While adding instance groups to an organization the role out is failing instead of missing instance groups assigned to the organization. That is caused while the organization is created before the instance groups will be created to the automation controller.

## Desired Behavior

While creating the organization first set the `assign_instance_groups_to_org` variable to false. 

## Actual Behavior

Manuel creating the Instance Group in the AAP so it's present while creating the Organizations.

## STEPS TO REPRODUCE

Add an instance group to your configuration and also assign these to an organization.

kreativmonkey:
# What does this PR do?

Use the provided switch `assigne_instance_groups_to_org` from the `controller_organization` role while creating the organization before the instance groups are deployed. 

https://github.com/redhat-cop/infra.aap_configuration/blob/aff0707ba30ed3ee2720795e7757c8402decaef4/roles/controller_organizations/defaults/main.yml#L12

# How should this be tested?

I didn't test if the behevior is working like expected. But while it's working for assignments of the other ressources and the implementation is the same, i expect that this change is working:

https://github.com/redhat-cop/infra.aap_configuration/blob/aff0707ba30ed3ee2720795e7757c8402decaef4/roles/controller_organizations/defaults/main.yml#L9-L11

This will close #1225

### DIFF
--- a/roles/dispatch/defaults/main.yml
+++ b/roles/dispatch/defaults/main.yml
@@ -16,6 +16,7 @@ gateway_configuration_dispatcher_roles:
     assign_galaxy_credentials_to_org: false
     assign_default_ee_to_org: false
     assign_notification_templates_to_org: false
+    assign_instance_groups_to_org: false
   - role: gateway_applications
     var: aap_applications
     tags: applications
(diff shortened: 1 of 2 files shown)


kreativmonkey:
How to solve the pre-commit failing step? I think there is nothing i can do to solve these.

djdanielsson:
It appears to be an issue with the CI itself. We are looking into seeing what we can do about that

djdanielsson (approved):
LGTM

PR #1234, opened 2025-11-20
adding project sync option to fix #1233

### LINKED ISSUE #1233: redeployment of eda_project will not sync
## Summary
A redeployment of an existing EDA-Project will not sync the project with the Git Repository.
In the `ansible.eda.project` module exists a parameter `sync` which is set to `false` by default.
this can't be overwritten using `infra.aap_configuration.eda_projects` due to missing parameter.

## Issue Type

- Bug Report

`ansible.eda.projects` (Version 2.10.0) - parameter `sync`: Triggers the synchronization of the project.
`infra.aap_configuration.eda_projects` (Version 3.8.4): parameter is missing.

djdanielsson:
# What does this PR do?

adding project sync option

# How should this be tested?

manually

resolves #1233

none

### DIFF
--- a/changelogs/fragments/1233.yml
+++ b/changelogs/fragments/1233.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - added the option of project sync to eda_projects
+...
--- a/changelogs/fragments/1233.yml
+++ b/changelogs/fragments/1233.yml
@@ -1,5 +1,5 @@
 ---
-bugfixes:
+minor_changes:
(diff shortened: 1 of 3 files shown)


Tompage1994 (changes requested):
Approving, BUT there is also the `proxy` option on the ansible.eda.project module which isn't implemented and may be good to implement here too? 

Also, technically this should be a minor_change rather than a bugfix because of changing APIs

Tompage1994 (approved):

PR #1248, opened 2025-12-11
Make it possible to exclude roles from the default list with dispatch role

### LINKED ISSUE #1247: Using dispatch role no easy way to define inventory source without updating them
The dispatch role includes both `controller_inventory_sources` and `controller_inventory_source_update` in its `controller_configuration_dispatcher_roles` default list and both of these roles use `controller_inventory_sources` for looping.

This means there is no easy to to prevent inventory sources being updated when they are defined.

The loop in `controller_inventory_source_update` prefers `inventory_sources` but only runs if `controller_configuration_dispatcher_roles` is defined, this looks more like a bug not a way to control updating.

One could redefine `controller_configuration_dispatcher_roles` but that would then add maintenance burden to keep the local definition in sync with the role definition otherwise on each collection update.

A role-specific variable or an extra optional variable for each inventory source would be one way to tackle this.

The most generic option could be to make it possible to define a list of roles to be excluded by the dispatch role. That way the dispatch role's definitions do not need to be duplicated and no new role/list-specific variables would be needed. This scheme would also make it possible to skip the indexing roles with hub configuration, which are often unhelpful given how few images are typically synced from a container registry.

So in the case, by default nothing would change but if `aap_configuration_dispatcher_exclude_roles` (or such) would be a non-empty list then any role listed there would be excluded from the final `aap_configuration_dispatcher_roles`.

Thanks.

myllynen:
Resolves #1247

### DIFF
--- a/changelogs/fragments/dispatch_exclude_roles.yml
+++ b/changelogs/fragments/dispatch_exclude_roles.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - Make it possible to exclude roles from the default list in the dispatch role
--- a/roles/dispatch/README.md
+++ b/roles/dispatch/README.md
@@ -6,7 +6,11 @@ An Ansible Role to run all roles in the infra.aap_configuration collection.
 
 ## Variables
 
-This is a meta role, its purpose is to run the other roles in the collection, it does not run all of them, and can be used to call roles in a custom order, The control variable is the `aap_configuration_dispatcher_roles` which will pull roles from the different services and run them, If you wish to just run a subset of services from our default lists remove entries from this var.
+This is a meta role, its purpose is to run the other roles in the collection, it does not run all of them, and can be used to call roles in a custom order.
+
+The control variable is `aap_configuration_dispatcher_roles` which will set the roles to run for different services.
+
+If you wish to run just a subset of roles either use `aap_configuration_dispatcher_exclude_roles` to exclude entries from the default list or define `aap_configuration_dispatcher_roles` with wanted roles locally.
 
 ### Option: `dispatch_include_wildcard_vars`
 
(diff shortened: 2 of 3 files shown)


djdanielsson:
If someone wants to skip something they can override the values, I am not sure this is worth doing. @Tompage1994 @sean-m-sullivan thoughts?

Tompage1994:
Agreed, this functionality is possibly by just defining your own dispatch vars

sean-m-sullivan:
I'm on the fence on this, it adds more versatility, in a simple way, and I'm not sure its "needed", but its a nice to have for some people working off a slightly deviated default list. I'd vote it in, but will wait on Tom to weigh in.

myllynen:
I've used my own definition so far but there are a few reasons I proposed this:

1) If I merely want to skip a role, say indexing registries, with this approach the intention is much clearer than listing multiple roles and their variables and tags but not the one I want to skip - so this is more elegant and more expressive approach

2) With this approach my own configuration does not require me to periodically check whether all the roles available in the collection are included in case the collection adds new ones, or perhaps changes the ordering, variables, or tags of them for whatever reason

3) The amount of new code added is minimal and should be near zero overhead and maintenance burden for the collection

Thanks.

Tompage1994:
Having thought a little more about it, I think maybe this does make sense. In the same way you have `tags` and `skip_tags` it gives you a way to define either way which you want to do. 

So I'm in favour. 

I'll try get a chance to look properly at this later beyond the generic idea but I'd want to be sure the logic we use is solid here and edge cases are explored.

Tompage1994 (changes requested):
I am happy with this. The only thing I'd ask to add is an example in the readme of excluding a role. (e.g. the controller_inventory_source_update role)

myllynen:
Thanks, I added such an example now.

Tompage1994 (approved):

Issue #1241, opened 2025-12-03
Tweak wildcard variable handling


myllynen:
Currently wildcard vars are handled like

`lookup('ansible.builtin.varnames', 'gateway_settings.*'))`

which works but causes duplicated items e.g. with the following defined:

gateway_settings
gateway_settings_test

Tweaking the parameter to be

`lookup('ansible.builtin.varnames', 'gateway_settings_.*'))`

would solve this issue and allow both of the above work together or independently.

The only potential downside would be that variables like

gateway_settingstest

wouldn't be accepted anymore but that seem more like a good thing.

What do you think, would a PR for this be acceptable? Thanks.

myllynen:
It might be best to use this format:

`lookup('ansible.builtin.varnames', '^gateway_settings_.*'))`

as otherwise something like todo_gateway_settings_test would not be ignored.

djdanielsson:
For settings it should be dict combine. I am working on also adding functionality to ensure it's unique but I haven't finished testing it yet. I should be opening a PR soon

djdanielsson:
Good point on the regex lookup though I will look to add that to all of them

myllynen:
Thanks for looking into this.

Perhaps you could also consider these ideas while at it, I'll leave it up to you to determine whether to actually use them or not:

1) Using query() instead of lookup() would allow for more compact code for loop
2) Given that it is possible to use variables also on the left-hand side with set_fact it might be possible to use just one task with product() and/or inline jinja but the resulting code might not be somewhat hard to read

myllynen:
FWIW, I created a standalone example to demonstrate that the second item is doable, this would in essence allow you to just list the known AAP configuration variable names in role variables and then use similar task to put it all together:

https://github.com/myllynen/misc/blob/master/ansible_config_reader.yml

Tested with ansible-core 2.16 - 2.20 but I guess it's indeed a bit hard to read at least initially.

myllynen:
Few more additional considerations here might be whether to use `append` or `append_rp` when combining dicts and if the list of known variables (see the PR) should be a role-variable so that it could be reused elsewhere.

myllynen:
This has been fixed as part of the above commits, closing.

PR #1249, opened 2025-12-11
Rewrite dispatch/wildcard handling to use just one task

myllynen:
Avoid a long series of tasks when we can just one dynamic on instead.

The task construct has been tested with ansible-core 2.16 - 2.20.

Relates to #1241

### DIFF
--- a/roles/dispatch/tasks/include_wildcard_vars.yml
+++ b/roles/dispatch/tasks/include_wildcard_vars.yml
@@ -1,262 +1,70 @@
 ---
-- name: include_wildcard_vars | Set fact for gateway_authenticators
+- name: include_wildcard_vars | Combine wildcard vars
+  vars:
+    all_vars:
+      - gateway_authenticators
+      - gateway_authenticator_maps
+      - gateway_settings
+      - aap_organizations
+      - aap_applications
+      - gateway_http_ports
+      - gateway_service_clusters
+      - gateway_service_keys
+      - gateway_service_nodes
+      - gateway_services
+      - aap_teams
+      - aap_user_accounts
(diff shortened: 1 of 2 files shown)


djdanielsson:
Can you add unique filter to this so if there are multiple objects the same name and type it would only have 1 copy

Tompage1994 (approved):

djdanielsson:
once you get unique working here I will approve and merge

roles/dispatch/tasks/include_wildcard_vars.yml
```diff
@@ -1,262 +1,70 @@
...
+      - eda_decision_environments
+      - eda_rulebook_activations
+    var_name: "{{ item | regex_replace('^(' + all_vars | join('|') + ')_.*', '\\1') }}"
+    curr_val: "{{ lookup('ansible.builtin.vars', item) }}"
+    is_dict: "{{ true if curr_val is mapping else false }}"
+    prev_val: "{{ lookup('ansible.builtin.vars', var_name, default={} if is_dict else []) }}"
+  ansible.builtin.set_fact:
+    "{{ var_name }}": "{{ prev_val | combine(curr_val, list_merge='append', recursive=true)
+                          if is_dict else
+                          prev_val + curr_val }}"
```
djdanielsson:
```suggestion
                          (prev_val + curr_val) | unique }}"
```

djdanielsson (changes requested):
I think this is all that needs to be done

myllynen:
Thanks, I updated the PR with your proposal included.

Tompage1994 (approved):

PR #1256, opened 2026-01-07
Adjust hub role listing for the dispatch role

### LINKED ISSUE #1246: Incorrect dispatch hub role listing
In roles/dispatch/defaults/main.yml the definition for hub_configuration_dispatcher_roles contains:

```
hub_configuration_dispatcher_roles:
  - role: hub_ee_registry
    var: hub_ee_registries
    tags: registries
...
  - role: hub_ee_registry
    var: hub_ee_registries
    tags: registry
```

So two almost identical elements which is probably unintentional.

myllynen:
Resolves #1246

### DIFF
--- a/roles/dispatch/defaults/main.yml
+++ b/roles/dispatch/defaults/main.yml
@@ -1,5 +1,6 @@
 ---
 dispatch_include_wildcard_vars: false
+
 gateway_configuration_dispatcher_roles:
   - role: gateway_authenticators
     var: gateway_authenticators
@@ -73,9 +74,6 @@ hub_configuration_dispatcher_roles:
   - role: hub_ee_image
     var: hub_ee_images
     tags: images
-  - role: hub_ee_registry
-    var: hub_ee_registries
-    tags: registry
   - role: hub_ee_registry_index
     var: hub_ee_registries
     tags: ee_indices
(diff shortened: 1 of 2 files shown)


roles/dispatch/defaults/main.yml
```diff
@@ -61,6 +61,12 @@ hub_configuration_dispatcher_roles:
   - role: hub_collection
     var: hub_collections
     tags: collections
+  - role: hub_publish
```
Tompage1994: I'm not so keen on this one. 

The publish role is kind of intended as something separate and was deliberately omitted from the dispatch because it doesn't directly have a correlation 1:1 with an object being created (i.e. it does a little more).

If people did want to include it, they would be easily enough able to add this to the dispatch vars themselves. This flow is really meant to be our recommended flow.

Tompage1994 (changes requested):

roles/dispatch/defaults/main.yml
myllynen: But if the variables are not defined then it's a no-op so shouldn't change anything unless someone intentionally defines them? Now that the variables have been separated it should further reduce the element of surprise.

Tompage1994: I think we will be killed by this line though unfortunately:

https://github.com/redhat-cop/infra.aap_configuration/blob/6303df387b6ddbd17a8012284346c722590d410c/roles/hub_publish/defaults/main.yml#L15

Where `hub_collections` is also used in the `infra.aap_configuration.hub_collection` role

myllynen: Would it be possible to remove that fallback now when the next collection release looks to be a new major release 4.0 or do you see that needed for the foreseeable future?

myllynen: Ok, I misread the version status, 4.1.0 was just released.

I'll update the PR with an alternative approach that might be more suitable.

myllynen:
I separated the publish roles from the other hub roles now and added a variable to include them or not, what do you think about this approach?

Tompage1994 (approved):

PR #1258, opened 2026-01-07
Allow conditionally preventing object roles to be added to objects

Tompage1994:
# What does this PR do?
Allow conditionally preventing object roles to be added to objects directly.

By setting `aap_configuration_apply_object_roles: false`

Consider the scenario that I have a tight pipeline which is created with dispatch controlling which items can be created by the objects which are provided by the end users. With the addition of object roles being added directly, people may be able to subvert the controls by providing themselves admin access. This ensures that if we don't want to give someone the ability to directly add the objects we can.

# How should this be tested?


djdanielsson (approved):
LGTM

PR #1263, opened 2026-01-09
Replace pool_id to subscription_id in license management

antoinep-rh:
Make `controller_license` role compatible with `license` module.
See https://github.com/redhat-cop/infra.aap_configuration/issues/1068#issuecomment-3727633565


djdanielsson (approved):
LGTM

Tompage1994:
I don't think this has done all we expect it to...

Surely we didn't want to change just the variable, but we actually wanted to change the attribute on the module? It seems we did it in one instance but not the other: 

```
    pool_id: "{{ controller_license.subscription_id | default(omit) }}"
```

I can raise a new PR with this though.

PR #1265, opened 2026-01-12
Use labels where appropriate in hub_publish

myllynen:
This avoids littering output with details of find/stat/etc details.


djdanielsson (approved):
LGTM

PR #1274, opened 2026-02-02
fix: controller_license: re-allow parameters in argument_specs that s…

dbk-rabel:
…hould be allowed

# What does this PR do?
It re-allows parameters in argument_spec that should be allowed / are expected by the ansible code.


dbk-rabel:
This should fix #1276 1276

djdanielsson (changes requested):
please add a changelog fragment with a line about the change that you are making

djdanielsson (approved):
LGTM

PR #1275, opened 2026-02-06
Add webhook url parameter

snecklifter:
# What does this PR do?
Adds webhook_url to the job template docs and code

# How should this be tested?
It doesn't look like CI is set up to test (no external git service) this however it should be documented as a parameter.

Nope.


Tompage1994 (changes requested):
Hey @snecklifter unless I'm missing something, the ansible.controller module doesn't have a `webhook_url` parameter.

From memory of how this works, you can't actually specify the URL, it is auto-generated when you specify webhook_service and webhook_credential.

snecklifter:
You're not missing anything and are absolutely correct.

Issue #1295, opened 2026-03-11
infra.aap_configuration.controller_schedules fails when aap_configuration_enforce_defaults: true


wbrandt-az:
## Summary

I try to deploy my Application to AAP using `infra.aap_configuration`. By default I try to use 
`aap_configuration_enforce_defaults: true`
When I want to create a scheduler for a Jobtemplate an error occurs: "Unable to update scheduler"
Setting `aap_configuration_enforce_defaults: false` it works.

## Issue Type

- Bug Report

## Ansible, Collection, Controller details

```console (paste below)
ansible --version

ansible [core 2.20.1]
  config file = None
  configured module search path = ['/home/wbrandt/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/wbrandt/.local/lib/python3.14/site-packages/ansible
  ansible collection location = /home/wbrandt/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/wbrandt/.local/bin/ansible
  python version = 3.14.3 (main, Feb  4 2026, 00:00:00) [GCC 15.2.1 20260123 (Red Hat 15.2.1-7)] (/usr/sbin/python)
  jinja version = 3.1.6
  pyyaml version = 6.0.3 (with libyaml v0.2.5)

ansible-galaxy collection list

Collection                               Version
---------------------------------------- ------------
ansible.controller                       4.7.8
ansible.eda                              2.10.0
ansible.hub                              1.0.3
ansible.platform                         2.6.20251106
community.general                        12.3.0
community.mysql                          4.0.1
infra.aap_configuration                  4.2.1

Controller version
AAP Controller version 4.7.8

```

- ansible installation method: one of source, pip, OS package, EE
pip install

## Actual Behavior

Please give some details of what is actually happening.
Include a [minimum complete verifiable example] with:
```yaml
- name: Update Scheduler
  hosts: scheduler
  vars:
    aap_configuration_enforce_defaults: true

    controller_templates:
      - name: "server_report"
        organization: "{{ aap_org }}"
        credentials:
          - "mycredential"
        execution_environment: "Hub Default execution environment"
        extra_vars: {}
        inventory: "common_inventory"
        playbook: playbooks/report/main.yml
        project: "myproject"

    controller_schedules:
      - name: "server_report"
        rrule: "DTSTART;TZID=Europe/Berlin:20250823T224500 RRULE:INTERVAL=1;FREQ=DAILY"
[...]

PR #1298, opened 2026-03-16
removing enforce defaults for schedules on execution environments

djdanielsson:
removing enforce defaults for schedules on execution environments because the it causes the api to return an error


djdanielsson:
#1295

branic (approved):
LGTM

Issue #1280, opened 2026-02-10
Make Error Handling Optional


Tompage1994:
**Is your feature request related to a problem? Please describe.**
The error handling part of the collection causes a lot of noise in the logs.

**Describe the solution you'd like**
We need to change async status into three files which are imported, likely by a list:
- collect_async_status
- register_value
- error_handle

This way we can then only include the pieces we need and won't be adding too much output when unwanted.

**Describe alternatives you've considered**
We considered if we could just include conditionals, but this will still clutter output

**Additional context**
None

PR #1299, opened 2026-03-16
#1280 error handling improvements

djdanielsson:
this was Claude-4.6-opus solution for #1280 figured I would throw it in there while also making some tweaks to fix the error handling issue that #1295 showed it didn't account for.


djdanielsson:
I have not tested this yet, wanted to see what people thought of it because it was generated. I will try and test it in the next few days.

branic:
In a quick look there are some questions I have about new variables that have been introduced, but I haven't gone through the changes in detail. I'll let you get some testing done before I do an in-depth review.

djdanielsson:
ok this one is tested and works but still not sure i love it, would like to see what others think of it

Tompage1994:
@djdanielsson As you know I'm lazy. I trust it works. But can you provide a sample output of what we will see with this change?

PR #1306, opened 2026-03-17
Fix dispatch ordering: credentials before instance_groups

### LINKED ISSUE #1014: controller_instance_groups managed before controller_credentials, which may be required
## Summary

If setting up a new controller, `controller_instance_groups` is higher in the dispatch defaults main.yml than controller_credentials.  This results in Instance groups being unable to be managed if they require a credential, such kubernetes authentication credentials.

Moving controller_instance_groups to lower in this list, after controller_credentials would address this

## Issue Type

- Bug Report

```console (paste below)
ansible --version

infra.aap_configuration   3.1.0  
Controller version
AWX 24.6.1
```

## Desired Behavior

A new controller from scratch should build items in an order that does not depend on future items

## Actual Behavior

Please give some details of what is actually happening.
Include a [minimum complete verifiable example] with:

- playbook / task
- configuration file / list
- error

Instance group failed to be created because the credential needed by it had not been created yet

```console (error)

```

## STEPS TO REPRODUCE

```yaml (playbook/task)

```

```yaml (config/list/array/variables)

```

djdanielsson:
Instance groups may reference credentials (e.g. container group credentials), but controller_instance_groups was dispatched before controller_credentials and controller_credential_types. This caused failures when setting up a new controller from scratch.

Reorder so credential_types, credentials, and credential_input_sources are processed before instance_groups.

Fixes #1014

Made-with: Cursor

### DIFF
--- a/changelogs/fragments/1014-fix-dispatch-ordering.yml
+++ b/changelogs/fragments/1014-fix-dispatch-ordering.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - Fix dispatch ordering so credential_types, credentials, and credential_input_sources are processed before instance_groups.
+...
--- a/roles/dispatch/defaults/main.yml
+++ b/roles/dispatch/defaults/main.yml
@@ -107,9 +107,6 @@ controller_configuration_dispatcher_roles:
   - role: controller_instances
     var: controller_instances
     tags: instances
-  - role: controller_instance_groups
-    var: controller_instance_groups
-    tags: instance_groups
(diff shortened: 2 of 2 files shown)


djdanielsson:
I think we need to double check this one

sean-m-sullivan (approved):

sean-m-sullivan:
This looks good, makes sense, the ordering on instance groups just needs to be before job templates.

sean-m-sullivan (approved):

PR #1310, opened 2026-03-17
1270 - Fixed labels and notification_templates

djdanielsson:
Fixed labels and notification_templates not being applied when defined directly
in user data because the Jinja2 expressions checked ``related.*`` (API-exported
format) before the direct user value. Reordered the fallback chain to check the
direct value first, matching the existing correct pattern used by ``credentials``.
Affected roles - controller_job_templates, controller_schedules,
controller_workflow_job_templates, controller_projects,
controller_inventory_sources, controller_organizations, gateway_organizations


Tompage1994 (approved):

djdanielsson:
idk why it says there are conflicts I tried doing it by hand and it seems to be fine, going to close and open a new PR from a fresh branch

PR #1314, opened 2026-03-18
1313 - fix no default password for gateway user

### LINKED ISSUE #1313: In aap_user_accounts, users_default_password is always sent regardless resulting in error
## Summary

When updating users using aap_user_accounts which are using external authenticators, we get:

 '{''password'': [''Password can not be set for this user'']}'

despite not setting a password parameter

## Issue Type

- Bug Report

```console (paste below)
Automation Controller Version
4.7.9
Event-Driven Ansible Version
1.2.6
Automation Hub Version
4.11.6
infra.aap_configuration
4.2.0

```

## OS / ENVIRONMENT

Custom EE with 4.2.0 from Red Hat Automation Hub

## Desired Behavior

User should become managed without error.

## Actual Behavior

```
started: 1
finished: 1
stdout: ''
stderr: ''
stdout_lines: []
stderr_lines: []
ansible_job_id: j352622371139.340
results_file: /runner/.ansible_async/j352622371139.340
msg: '{''password'': [''Password can not be set for this user'']}'
invocation:
  module_args:
    username: Redacted
    password: VALUE_SPECIFIED_IN_NO_LOG_PARAMETER
    first_name: Redacted
    last_name: Redacted
    is_superuser: true
    gateway_hostname: ansible.apps.redacted
    gateway_username: automation-service-account
    gateway_password: VALUE_SPECIFIED_IN_NO_LOG_PARAMETER
    gateway_validate_certs: true
    state: present
    gateway_token: null
    gateway_request_timeout: null
    email: null
    is_platform_auditor: null
    organizations: null
    authenticators: null
    authenticator_uid: null
    associated_authenticators: null
_ansible_no_log: false
attempts: 1
changed: false
```

## STEPS TO REPRODUCE

```yaml (playbook/task)

aap_user_accounts:
  - username: Redacted
    is_superuser: true
    first_name: Redacted
    last_name: Redacted
    associated_authenticators:
      - "keycloak"

```

snecklifter:
# What does this PR do?
Updates gateway_user to send empty password parameter instead of change_me

# How should this be tested?
Omit the password parameter for an externally authenticated user

resolves #1313


djdanielsson:
I am wondering more if we should even define anything at all or force the user to set it. They could choose nothing or some random string. thoughts @sean-m-sullivan @Tompage1994 ?

branic:
I don't think we should define it at all so the user is forced to set it. I would rather have an undefined variable error thrown than setting some string we have chosen.

roles/gateway_users/defaults/main.yml
```diff
@@ -24,7 +24,7 @@ aap_user_accounts: []
 #  - authenticator_uid
 #  - state
 
-users_default_password: change_me
+users_default_password: ""
```
djdanielsson:
```suggestion
```

djdanielsson (changes requested):

roles/gateway_users/defaults/main.yml
```diff
@@ -24,7 +24,6 @@ aap_user_accounts: []
 #  - authenticator_uid
 #  - state
```
djdanielsson, line 26:
```suggestion
```

djdanielsson (approved):
made some changes and now it looks good, thanks for the contribution

PR #1318, opened 2026-03-19
Add 'tags: always' to include_wildcard_vars and main task definitions

DLV111:
# What does this PR do?

Always include the wildcard vars, regardless of what tags are passed when the feature is enabled

When running the dispatch with a tag of --tags=project, the wildcard_vars component fails to run, it only works as expected when no tags are run

# How should this be tested?

Run with include wildcards enabled and include a tag of --tag=project or any other valid tag. It should run the process without this the expected outcome is different. Note this issue is not seen if you're using the base var eg controller_projects: over controller_projects_myteam:

N/A

N/A

### DIFF
--- a/changelogs/fragments/dispatch_tags.yml
+++ b/changelogs/fragments/dispatch_tags.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - Added `tags: always` to tasks in the dispatch role for include_wildcard_vars to ensure they run regardless of applied tags.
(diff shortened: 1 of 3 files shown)


changelogs/fragments/dispatch_tags.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
```
djdanielsson:
```suggestion
bugfixes:
```

djdanielsson (changes requested):
I think we should consider this a bug, but thanks for catching it and contributing

djdanielsson (approved):
LGTM

PR #1341, opened 2026-04-17
Also handle {% .. %} like {{ .. }} in controller_credential_types.injectors

myllynen:


lennysh:
@myllynen Can you provide an example `before` and `after` showing what this change fixes?

myllynen:
Before only "{  {" is accepted in CaC as a way to define variables for credential type injectors meaning that they will become "{{ ... }}". After the patch also "{  %" will get translated to "{% ... %}" which unlocks jinja conditionals.

sean-m-sullivan:
@myllynen we've been actually looking at depreciating this, as its a hack, The unsafe method is the currently accepted method to fix this
https://github.com/redhat-cop/infra.aap_configuration/tree/devel/roles/controller_credential_types#formatting-injectors

If others were to agree to merge, this would also need a documentation update.

myllynen:
The background for this is that I wanted to move credential type definitions in my CaC example from an organization section to AAP-wide since credential types are not organization-specific. A related issue/question is at https://github.com/redhat-cop/infra.aap_configuration/issues/1342.

But I noticed that with the earlier used raw/endraw approach these worked for injectors, regardless of where they were defined:

controller_credential_types_test_org: ...
controller_credential_types_common: ...

However, this didn't work, and that's how I found out about "{  {" and thought to use "{  %" instead:

controller_credential_types: ...

Thanks.

djdanielsson:
I think !unsafe is probably best, I think this was left to try and support what we had working before the move to !unsafe, if anything I would probably rather remove everything and require !unsafe as the only solution... idk

myllynen:
Unfortunately !unsafe doesn't work with include_vars so that was the reason I moved away from it. In practicaly terms, creating helper playbook that don't require users to provide several -e options !unsafe can't be used.

Tompage1994:
I think I'm in agreement that we do need to keep a way of doing this without using unsafe. I've also found for multi line strings (which I guess this change would be used for) `!unsafe` does not work. 

I was actually thinking about expanding the usage of the `{  {` notation to other areas such as `extra_vars` on a job template (I'm sure there are others too) .

myllynen:
How shall we go about this? Merging this would allow for using all kinds of injectors and it certainly looks like `!unsafe` is not an alternative that would work in all scenarios.

Tompage1994:
I have tested this and confirmed that it works. @djdanielsson @sean-m-sullivan I think this should be approved on the condition that some documentation is added [here](https://github.com/myllynen/infra.aap_configuration/blob/credentials-types-escaping/roles/controller_credential_types/README.md#formatting-injectors) and that an example is added below

djdanielsson:
ok

Tompage1994 (approved):

Issue #986, opened 2024-11-26
Validating arguments against arg spec 'main' fails for some roles


cmasopust:
## Summary
When dispatching all the roles, some are failing with the at "Validating arguments against arg spec 'main'" with the following error message (e.g. for inventory_sources):

```
TASK [infra.aap_configuration.controller_inventory_sources : Validating arguments against arg spec 'main' - An Ansible Role to create inventory sources on Ansible Controller.] ***
fatal: [localhost]: FAILED! => {"msg": "the field 'tags' should be a list of ((<class 'str'>,), <class 'int'>), but the item '['inventories', 'inventory_sources']' is a <class 'list'>"}
```

Pretty sure I'm doing something wrong, but.... what?

## Issue Type

- Bug Report

## Ansible, Collection, Controller details
AAP version is 2.5
infra.aap_configuration version is 3.1.0

```console (paste below)
$ ansible --version
ansible [core 2.14.17]
  config file = /home/cmaso/.ansible.cfg
  configured module search path = ['/home/cmaso/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.9/site-packages/ansible
  ansible collection location = /home/cmaso/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.9.19 (main, Sep 11 2024, 00:00:00) [GCC 11.5.0 20240719 (Red Hat 11.5.0-2)] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```
## OS / ENVIRONMENT

RHEL 9

Thanks for your help!

djdanielsson:
can you provide an example of the inventory source you are attempting to pass?

cmasopust:
it happens for any role that is configured in **roles/dispatch/defaults/main.yml** with a list of tags instead of a single tag.

e.g. for controller_projects:

```yaml
  - role: controller_projects
    var: controller_projects
    tags:
      - inventories
      - projects
```
it results in the following error when dispatching:
```
TASK [infra.aap_configuration.meta_dependency_check : Print dependency check status] **********************************************************************
[...]

PR #1343, opened 2026-04-17
fix: ensure apply:tags is always a string for Ansible 2.13 compat

amayagil:
Ansible 2.13.7 wraps any list returned by a Jinja2 template in apply:tags into another list, causing tag validation to fail with "the field 'tags' should be a list of str/int, but the item is a list".

Using a conditional ensures string tags pass through unchanged while list tags (e.g. ['inventories', 'projects']) fall back to their last (most specific) element as a single string.

Made-with: Cursor

# What does this PR do?

resolves #[number]

### DIFF
--- a/roles/dispatch/tasks/main.yml
+++ b/roles/dispatch/tasks/main.yml
@@ -8,7 +8,7 @@
   ansible.builtin.include_role:
     name: "{{ __role.role }}"
     apply:
-      tags: "{{ __role.tags }}"
+      tags: "{{ __role.tags if __role.tags is string else __role.tags | last }}"
   when: vars[__role.var] is defined
   tags: always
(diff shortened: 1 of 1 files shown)


djdanielsson:
Thank you for the PR, I have done some tests this is related to #986. We have documented in our meta/runtime.yml that this collection requires ansible-core 2.16.0+ this issue was fixed in an early version of ansible-core 2.15. We do not support running this collection in an unsupported old version of ansible-core, plus this change would alter intended tag behavior. Some dispatcher entries intentionally use multiple tags such as `['inventories', 'projects']` so they can be selected by both a broad tag and a specific tag. Converting list tags to only the last element would drop the broader tag and change existing selection behavior. Because of all this I am closing this PR

PR #1376, opened 2026-06-16
Feature/support new_name and roles parameters in controller roles integration

nick-a-meyer:
When new_name is utilized in conjunction with controller_roles, an error will be thrown indicating that the target for controller_roles does not exist. This is because the target object is renamed. Need to make sure that if there was already an object with the new_name and the preceding creation/update task fails, that controller_roles are not applied to the preexisting object leading to a potential security issue

# What does this PR do?

Adds support for using `new_name` attribute in conjunction with `roles` attribute across all 6 resource types that support both features.

**Primary Fix:**
When a resource is renamed using `new_name`, the roles integration now correctly references the new name instead of the old name, preventing "target does not exist" errors.

**Secondary Fix (Security):**
Added filtering to prevent roles from being assigned to wrong resources when creation fails but error collection mode (`aap_configuration_collect_logs: true`) is enabled.

**Changes:**
- Updated roles integration in 6 controller roles to use `{{ _item.new_name | default(_item.name) }}`
- Added `_failed_items` filtering to skip failed resources when assigning roles

## How should this be tested?

1. Create an object with roles and then attempt to rename it with new_name. This should no longer throw an error.
2. Create an object with new_name and then try to rename a second object with roles to match it. This should throw an error and the object with the target name should not have roles applied to it.

### DIFF
--- a/changelogs/fragments/1376-support-new-name-in-controller-roles-integration.yaml
+++ b/changelogs/fragments/1376-support-new-name-in-controller-roles-integration.yaml
@@ -1,0 +1,5 @@
+---
+bugfixes:
+  - "Fix inline role assignments when ``new_name`` is specified on controller resources. Now uses the new name for role assignment."
+  - "Prevent inline role assignments from being applied to pre-existing objects when the resource creation/update fails during a ``new_name`` operation."
+...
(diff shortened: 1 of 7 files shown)


djdanielsson (changes requested):
can you make a changelog fragment for this please

nick-a-meyer:
@djdanielsson added fragment and rebased. Please let me know if anything else is needed.

djdanielsson (approved):
LGTM

PR #1379, opened 2026-06-18
add missing rename of controller_teams to aap_teams

Zathrus:
# What does this PR do?

There is a missing conversion mapping from controller_teams to aap_teams. The controller_teams was part of controller for 2.4, but the corresponding 2.5+ object is gateway's aap_teams

## How should this be tested?

No testing needed? Validate that this is the proper mapping.

No


djdanielsson (changes requested):
I think this should be in the controller section and if you could add a changelog fragment saying like a "trivial" doc change

Tompage1994 (approved):

PR #1381, opened 2026-06-22
Updating controller_hosts_groups documentaiton and arguments to match code

ForcedHero:
# What does this PR do?

Updates the Documentation on the controller_hosts_groups role to match whats in the code and updates the arguments_specs.yml in the same to match

## How should this be tested?

If applicable, using this module add hosts to groups amd make sure that the secure logging values are respected.  (shouldnt be any changes though as it was documentaiton changes. 

Not that I see. I jsut notices this and it looked like something that I could fix.

resolves #[number]


djdanielsson (approved):
LGTM

PR #1383, opened 2026-06-23
Fix - Setting state absent on workflow with nodes defined fails

nick-a-meyer:
# What does this PR do?

I should be able to define a full workflow object including nodes and then be able to simply flip the state flag to absent to remove it. Currently, if I set the state flag to absent, the relevant workflow role will still try to update the nodes causing failures since the parent workflow simply does not exist.

This fix adds a conditional to the node update, making sure that state: present appears on the parent workflow before attempting to do a node update operation.

## How should this be tested?

On a new cluster the following playbook can be used to test the fix and replicate the error. 

```
- name: Testing state absent bug for workflows with nodes defined
  hosts: localhost
  connection: local
  become: false
  gather_facts: false
  vars:
    aap_hostname: <OMITTED>
    aap_username: <OMITTED>
    aap_password: <OMITTED>
    controller_workflows:
      - name: Testing
        state: absent
        organization: Default
        inventory: Demo Inventory
        simplified_workflow_nodes:
          - identifier: Test Node
            unified_job_template: Demo Job Template
            lookup_organization: Default
  roles:
    - infra.aap_configuration.dispatch
```

N/A

N/A

### DIFF
--- a/changelogs/fragments/support-state-absent-on-fully-defined-workflows.yml
+++ b/changelogs/fragments/support-state-absent-on-fully-defined-workflows.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - "Setting state: absent on a fully defined workflow with nodes defined will no longer throw an error."
+...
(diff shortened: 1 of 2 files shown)


djdanielsson (approved):
LGTM

nick-a-meyer:
@djdanielsson thank you. I do not have write access so please merge this at your leisure.

PR #1387, opened 2026-06-24
Remove unused gateway_teams/tasks/async_gateway_team.yml file

myllynen:
The file is not included from anywhere so unused.


Tompage1994:
Can you please add a bugfix changelog for this. 
*_technically_* someone _could_ be calling this file in this role (Quite why, I have no idea) so we should at least note it in the changelog (plus the CI will fail without)

Tompage1994 (approved):

PR #1388, opened 2026-06-24
Fix inline object roles failing with duplicate names across organizations

### LINKED ISSUE #1382: Directly added roles fail with non-unique names
In `roles/controller_job_templates/tasks/main.yml` we have:

```
- name: Add roles to job templates
  ansible.builtin.include_role:
    name: infra.aap_configuration.controller_roles
  when: aap_configuration_apply_object_roles
  vars:
    controller_roles: "{{ _controller_roles | from_yaml | default([], true) | list }}"
    _item_list: "{{ job_templates if job_templates is defined else controller_templates }}"
    _controller_roles: |
      {% for _item in _item_list %}
      {% if _item.roles is defined %}
      {% for _role in _item.roles.keys() %}
      - role: "{{ _role }}"
        job_templates:
          - "{{ _item.name }}"
        teams: {{ _item.roles[_role].teams | default(omit) | to_json }}
        users: {{ _item.roles[_role].users | default(omit) | to_json }}
      {% endfor %}
      {% endif %}
      {% endfor %}
```

This will fail if job template names are not unique. This could be fixed by adding this after `job_templates`:

```
        lookup_organization: {{ _item.organization | default(omit) | to_json }}
```

I see the following using directly added roles:

```
roles/controller_credentials/tasks/main.yml
roles/controller_instance_groups/tasks/main.yml
roles/controller_inventories/tasks/main.yml
roles/controller_job_templates/tasks/main.yml
roles/controller_projects/tasks/main.yml
roles/controller_workflow_job_templates/tasks/main.yml
```

We could fix `controller_job_templates` as above and organization is not applicable to `controller_instance_groups`.

The question then is what to do with the others? The corresponding modules do have `organization` (not `lookup_organization`) parameter but it is at the moment unclear would using that help.

Tompage1994:
## Summary

- Fixes [#1382](https://github.com/redhat-cop/infra.aap_configuration/issues/1382) by passing `lookup_organization` when generating `controller_roles` entries from inline `roles` definitions on organization-scoped controller resources.
- Applies the fix to `controller_job_templates`, `controller_credentials`, `controller_inventories`, `controller_projects`, and `controller_workflow_job_templates`.
- Leaves `controller_instance_groups` unchanged because instance groups are not organization-scoped.

When the same object name exists in multiple organizations, the `ansible.controller.role` module needs `lookup_organization` to resolve the correct target. The inline role templates previously only passed the object name, causing failures in those cases.

## Test plan

- [ ] Configure two job templates (or credentials, inventories, projects, or workflows) with the same name in different organizations, each with inline `roles` assignments
- [ ] Run the corresponding role with `aap_configuration_apply_object_roles: true`
- [ ] Verify roles are applied to the correct object in each organization
- [ ] Verify credentials without an `organization` field still work (no `lookup_organization` emitted)


djdanielsson (approved):
LGTM

PR #1389, opened 2026-06-24
Fix YAML parsing failure for inline object roles with lookup_organization

Tompage1994:
## Summary

- Follow-up fix for [#1382](https://github.com/redhat-cop/infra.aap_configuration/issues/1382) / [#1388](https://github.com/redhat-cop/infra.aap_configuration/pull/1388).
- Replaces Jinja `{% if %}` block tags inside the inline `_controller_roles` YAML template with a single-line `default(omit) | to_json` expression for `lookup_organization`.
- The block tags mis-indented generated YAML keys, causing `from_yaml` to fail when parsing inline role definitions (e.g. on inventories with `organization` and `roles` set).

## Test plan

- [ ] Apply inline roles on an inventory with `organization` and `roles` defined (same config that failed after #1388)
- [ ] Verify `Add roles to inventories` completes without `from_yaml` parse errors
- [ ] Verify credentials without an `organization` field still work (no `lookup_organization` emitted)


djdanielsson (approved):
LGTM

PR #1391, opened 2026-06-25
Do not set changed_when in gateway roles to report changes correctly

myllynen:
Gateway roles use ansible.platform modules which in the most recent collection version 2.7.20260615 all use action plugins and support reporting change status correctly, also in check mode.


Tompage1994:
To me, the problem here is not that the change isn't reported correctly (I would assume it always would have?) but that we set `async: 0` meaning that there will be no meaningful changed value. Instead we need to determine the change in the `collect_async_status` role.

Happy to be proved wrong, I have not yet tried the newly updated ansible.platform collection

myllynen:
I'm happy to prove you wrong :) Even with `async: 0` the related roles indeed report the change status correctly.

I've used my simple CaC example at https://github.com/myllynen/aap-automation/tree/master/examples/cac/simple/aap_cac/test_org to test this. With this patch on the initial run everything reports changed as expected, from then on only if I tweak the CaC somehow.

But without this patch nothing gateway related reports changes, not even in cas.

cdenney-silex:
These tasks should correctly report their changes in the collect_async_status role. I think the changed_when: false (or the check_mode logic) is intended to deduplicate the number of tasks that report changes. Wouldn't this PR would cause Gateway changes to be reported twice?

myllynen:
I have now tested with the following versions:

AAP 2.7-2
ansible.platform 2.7.20260630
infra.aap_configuration 4.7.0

And applying this CaC on a freshly installed AAP instance with zero post-install changes in place:

```
aap_organizations_test_org:
  - name: Test Org
    description: Test organization
```

Without this patch zero changes are reported on the first and subsequent runs and also zero changes if updating the description.

With the patch applied one change is reported on the first run and no changes on subsequent runs until the description is updated after which a change is reported when applying the CaC after that for the first time.

Note that the situation is different for non-gateway_* roles. There collect_async_status reports the changes e.g. for controller_* configurations.

Tompage1994 (approved):
Ok, I've now had time to test and this does work. It seems to work differently than the controller modules which is slightly odd. But happy to accept this now. Sorry it took a while

ursetto:
This still won't work in check mode because most of the `gateway_*` roles use `async: 1000` instead of `async: "{{ ansible_check_mode | ternary(0, 1000) }}"` like other roles, so you get the error `check mode and async cannot be used on same task`.

myllynen:
That sounds clearly something that ought to be fixed but it's a separate issue, though?

My proposal would be merge this one "soon" unless there are issues specific to this PR and then figure out the solution to this finding in a new ticket/PR with a clean(er) starting point.

PR #1398, opened 2026-07-09
Define wildcard var list in vars/ and allow for external use

myllynen:
This is zero change for the collection/role but playbooks outside of the collection can now reuse the variable list.


roles/dispatch/vars/main.yml
```diff
@@ -0,0 +1,59 @@
+---
+__aap_cac_vars:
```
Tompage1994: The general convention is to use `__` when the variable is entirely internal. If we are doing this to allow it to be exposed, then I would say it should be something like the following

```suggestion
aap_configuration_all_vars:
```

Tompage1994 (changes requested):
Minor change required. Could you then also add something to the readme to show people how to make use of this change

Tompage1994 (approved):

PR #1404, opened 2026-08-05
Minor wildcard handling improvement and optimization

myllynen:
Define the wildcard pattern only once. Avoid when, use filter instead. Tweak setting is_dict.


djdanielsson (approved):
LGTM

PR #1416, opened 2026-08-21
Skip existing collection versions in hub_publish

### LINKED ISSUE #1412: Improve custom collection uploading
Currently, when uploading a certain version of a custom collection, say, foo.bar 1.2.3, it works as expected and does the needed git clone, collection building, and the actual upload.

However, the next time the same automation runs no check for this version of the collection is done but instead cloning, building, uploading are done again.

It would seem appropriate that before cloning/building/uploading a version of a collection, a check would be performed whether it exists already and skip these step if so. If's there a reason to indeed build and upload the same version again in some cases, then perhaps a separate parameter to force this could be provided. Thanks.

Tompage1994:
## Summary

- Add `hub_skip_existing_collections` (default `true`) so `hub_publish` queries Automation Hub before clone/build/upload and skips items whose version already exists
- Support optional per-item `namespace`, `name`, and `collection_version` fields for pre-clone skip checks without reading `galaxy.yml`
- Apply skip logic to `hub_custom_collections`, `hub_collection_list` tarball paths, and local-source dict entries; honor per-item `overwrite_existing`/`force` at upload time

Fixes #1412

## Test plan

- [x] `ansible-lint roles/hub_publish/` passes (production profile)
- [x] Manual integration test against AAP: first run publishes collection, second run skips clone/build/upload with assertions on `__hub_publish_pending_items`, `__hub_collection_list`, and `__hub_publish_should_skip`
- [ ] CI pre-commit workflow

### DIFF
--- a/changelogs/fragments/1412-skip-existing-hub-collections.yml
+++ b/changelogs/fragments/1412-skip-existing-hub-collections.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - hub_publish - Skip clone, build, and upload when a collection version already exists in Automation Hub. Use hub_overwrite_existing or per-item force to re-publish. Optional namespace, name, and collection_version fields enable pre-clone skip checks. Set hub_skip_existing_collections to false to restore the previous always-build behavior.
+...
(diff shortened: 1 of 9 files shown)


djdanielsson (approved):
LGTM

PR #1434, opened 2026-09-04
Default aap_request_timeout to 10 for blank AAP credential injection

djdanielsson:
## Summary

- Set `aap_request_timeout: 10` in `global_vars` defaults
- Update all controller, gateway, hub, and EDA roles to pass `request_timeout` / `gateway_request_timeout` using `default(10, true)`

When a GitOps job uses the **Red Hat Ansible Automation Platform** credential without the optional `request_timeout` input, AAP injects `aap_request_timeout` as an empty string via `extra_vars`. That value overrides role defaults and is not treated as undefined, so patterns like `default(omit)` or `default(omit, true)` still fail module argspec validation (`cannot be converted to float`).

Using `default(10, true)` coerces empty, null, and other falsy injected values to `10` while still allowing explicit overrides such as `aap_request_timeout: 30`.

## Test plan

- [ ] Run a CaC playbook with `aap_request_timeout: ""` set as an extra var and confirm roles no longer fail with `cannot be converted to float`
- [ ] Confirm an explicit `aap_request_timeout: 30` override is still honoured
- [ ] Verify hub, controller, gateway, and EDA dispatch paths all receive a numeric timeout


lennysh (approved):
LGTM

PR #1441, opened 2026-09-10
Fix RA example : fix decision environment name and add organization key

w4hf:
Fix the Rulebook Activation yaml definition example :

- organization is mandatory when state=present
- DE is now called 'Hub Default Decision Environment' instead of 'Automation Hub...'

# What does this PR do?

Fix rulebook activation yaml definition example 

## How should this be tested?

Using the fixed yaml now works out of the box (providing dependencies are present)

n/a


djdanielsson (changes requested):
please create a changelog fragment for this PR

djdanielsson (approved):
LGTM

PR #1447, opened 2026-09-14
fix: reorder gateway_role_definitions before   gateway_authenticator_maps in dispatch

imedaouidene:
# What does this PR do?
Reorders `gateway_role_definitions` to run before
  `gateway_authenticator_maps` in the
  `gateway_configuration_dispatcher_roles` list.

  When an authenticator map references a custom role
  definition, the dispatch fails because the role doesn't
  exist yet at the time the authenticator map is created.
  This fix ensures role definitions are created first.

  Updated in all places where the ordering is defined:
  - `roles/dispatch/defaults/main.yml`
  - `roles/dispatch/meta/argument_specs.yml`
  - `roles/dispatch/vars/main.yml`
  - `roles/dispatch/README.md`
  - `docs/CONVERSION_GUIDE.md`
  - Changelog fragment added

  ## How should this be tested?

  1. Define a custom `gateway_role_definition`
  2. Create a `gateway_authenticator_map` that references
  that custom role
  3. Run the dispatch role — it should succeed without
  errors about missing role definitions

N/A

  No logic changes — just moved `gateway_role_definitions`
  from after `gateway_users` to right after
  `gateway_authenticators` (before
  `gateway_authenticator_maps`) in all files that define
  the gateway dispatcher ordering.


djdanielsson (approved):
LGTM

PR #1449, opened 2026-09-15
fix: allow playbook-level override of assign_* flags in dispatch role

imedaouidene:
## What does this PR do?

  Fixes the `gateway_organizations` role to allow users to override `assign_galaxy_credentials_to_org`
  (and `assign_default_ee_to_org`, `assign_notification_templates_to_org`,
  `assign_instance_groups_to_org`) at the playbook level when using the dispatch role.

  The `dispatch` role sets these variables via task-level `vars:` on `include_role`, which has higher
  precedence than playbook vars. This blocks the documented ability to set `assign_*: false` at the
  playbook level — the dispatch `vars:` always wins regardless of what the user sets.

  ### Changes

  - **`roles/dispatch/tasks/main.yml`**: Renamed internal variables to `_dispatch_assign_*` so they no
  longer shadow the user-facing `assign_*` variables.
  - **`roles/gateway_organizations/tasks/main.yml`**: Updated conditions to check user-defined
  `assign_*` first, then fall back to `_dispatch_assign_*`, then to `true` for backward compatibility.
  - **`roles/gateway_organizations/defaults/main.yml`**: Removed `assign_*` defaults (the fallback
  `default(true)` in the task conditions provides the same behavior).
  - **`roles/gateway_organizations/README.md`**: Updated variable descriptions to document the
  playbook-level override.

  ### Variable resolution order

  assign_galaxy_credentials_to_org              # user playbook var (wins if set)
    → _dispatch_assign_galaxy_credentials_to_org  # dispatch internal (false=pass1, true=pass2)
      → true                                       # fallback for backward compat

  ## How should this be tested?

  1. Set `assign_galaxy_credentials_to_org: false` at the playbook level
  2. Run dispatch — both passes should skip galaxy_credentials assignment
  3. Remove the override — verify default two-pass behavior still works (first pass omits, second pass sets)
  4. Call `gateway_organizations` role directly (without dispatch) — verify default behavior unchanged
  (galaxy_credentials sent)

  resolves #

  Fully backward compatible: standalone usage of `gateway_organizations` (without dispatch) behaves
  identically. The `default(true)` fallback in the conditions replaces the removed role defaults.


djdanielsson (changes requested):
Thanks for the PR — the precedence fix makes sense to me.

I'm not sure I follow the ordering / fresh platform failure scenario though. The two-pass assign_* flow is meant to handle that: pass 1 skips assignment, controller_credentials runs before the second gateway_organizations pass in default dispatch order, then pass 2 assigns. I run fresh builds through dispatch all the time without hitting this. What does your setup look like where pass 2 fails because credentials don't exist yet?

That said, I do see value in the change: a playbook-level assign_* override is documented on the role but blocked today because dispatch sets the same names via task vars:. The _dispatch_assign_* approach seems reasonable for that. Users can also override via controller_configuration_dispatcher_roles, but a single playbook var is nicer.

Before merge, could you reframe the changelog/PR around the override/precedence fix rather than the ordering issue, rename the changelog fragment, and fix the markdown lint issue (CI failing).

